Issue #43, opened 2011-10-18
Manage transfers


almet:
Transfers of money from one person to another should be handled in a simple way.

almet:
Dupe with #137

PR #96, opened 2012-11-03
New feature : Settle the bill

### LINKED ISSUE #18: Add the ability to reduce the number of transfers to solve the balance
@Feth worked and published his code here: https://github.com/feth/peinard/

There is still a need to package it and for us to use decimals but that's basically it :)

aavenel:
Update patch from @Lastpixl with current code and add unit tests.

It seems to work well ! However, we need a way to manage transfers (#43) but I think we can already merge this because it's a feature long overdue ...
Close #18.

### DIFF
--- a/budget/models.py
+++ b/budget/models.py
@@ -51,13 +51,15 @@ class Project(db.Model):
 
     def get_transactions_to_settle_bill(self):
         """Return a list of transactions that could be made to settle the bill"""
+        #cache value for better performance
+        balance = self.balance
         credits, debts, transactions = [],[],[]
         # Create lists of credits and debts
         for person in self.members:
-            if self.balance[person.id] > 0:
-                credits.append({"person": person, "balance": self.balance[person.id]})
-            elif self.balance[person.id] < 0:
-                debts.append({"person": person, "balance": -self.balance[person.id]})
+            if balance[person.id] > 0:
+                credits.append({"person": person, "balance": balance[person.id]})
+            elif balance[person.id] < 0:
+                debts.append({"person": person, "balance": -balance[person.id]})
         # Try and find exact matches
         for credit in credits:
             match = self.exactmatch(credit["balance"], debts)


sschnug:
A few questions regarding the settling-algorithm (peinard???):

* Optimization Goal: What exactly is optimized?
 * Number of transfers? (A->B + A->C = 2)
 * Sum of transfer-values? (A->B:1 + A->C2 = 3)
 * Some combination? Something else?

* What guarantee does the algorithm give?
 * Optimal solution is guaranteed
 * Some kind of approximation-value is achieved

Let's assume, that the target-function is Min(|transfers|): the following example is clearly not optimal
demo-screenshot

aavenel:
It is a bug, settling-algorithm should minimize number of transfers, thanks for the report :+1:

sschnug:
Well, i get the feeling this will be tough (depending on the goals)! 

If minimization of transfers is enough -> there is a O(n * log n) algorithm doing this (sort and shift from each node to the next) => problem: transaction values accumulate "in the middle" => huge money amounts!

If minimization of transfers is desired & some extra rule like "transfer only negative to positive", this problem seems to be np-complete. In this case: which approach to pursue? -> Approximation or Exact Solution?

aavenel:
Hum I tried to reproduce the bug but I'm getting the optimal result. Could you post a "bills" capture too ? 

image

sschnug:
Sure... Used was a git-clone (from today) and the "demo-mode" on local machine!

http://i.imgur.com/OxwCf2I.png
Bug

aavenel:
Thanks ! Looks like a rounding bug. Unfortunatly, we store balances as a float instead of using a decimal type (https://docs.python.org/2/library/decimal.html#module-decimal). This leads to several subtle bugs like this one :/

As for the complexity, keep in mind that in practice, the "n" is very small. I didn't understand your comment about "transaction values accumulate in the middle", could you explain a bit more ?

sschnug:
Ah... nevermind the comment about the sorting-based algorithm. No need to discuss it further (including the accumulation issue).

Is there a good test-mechanism for this settle-algorithm? As mentioned, i don't like the fact at all, that it is so connected to the db-model. I would have used an abstract solver, which operatoes just on the balance-column on the left. This would be perfekt for testing (random-instances).

And i'm still doubtful about the algorithm guarantees, but without some "easy testing", i'm not sure how to pursue this (even if i have a ready solver).

aavenel:
Ok, I have a quick fix for this, but I don't like it. I will open a pull request to discuss this.

@sschnug You're right, it shouldn't be connected to the db-model. In fact, all it needs are the balances. The problem is I needed the exact balances (i.e. not rounded) to be able to reproduce the bug.

Feel free to write unit-tests for this feature, help is always appreciated :)
You can have a look here (https://github.com/spiral-project/ihatemoney/blob/master/budget/models.py#L52) to check the function we use.

sschnug:
All that db-model noise in this code :-/

Is this code based on [peinard](https://github.com/feth/peinard/blob/master/peinard.py)? If so, it would be easier for me to reason about the original code.

At first sight (@peinard): this is a heuristic and doesn't give any guarantees (well maybe it's a factor-2 approximation).

aavenel:
In fact the only function that really solve the problem is this one.

```python
    def exactmatch(self, credit, debts):
        """Recursively try and find subsets of 'debts' whose sum is equal to credit"""
        if not debts:
            return None
        if debts[0]["balance"] > credit:
            return self.exactmatch(credit, debts[1:])
        elif debts[0]["balance"] == credit:
            return [debts[0]]
        else:
            match = self.exactmatch(credit-debts[0]["balance"], debts[1:])
            if match:
                match.append(debts[0])
            else:
                match = self.exactmatch(credit, debts[1:])
            return match
```

Looks like a brute force algorithm so I think it only return optimal solution.

Issue #120, opened 2015-04-29
AssertionError on project creation


JocelynDelalande:
I tested it in two different machines, one with Debian Wheezy, the other with jessie, using virtualenvs.

Steps to reproduce :

- Install ihatemoney last master, as recomended in the README
- put a *budget/settings.py*  with the following contents :

```
DEBUG = True
SQLACHEMY_ECHO = DEBUG
TESTING = True
```
- Try to create a projet
- bang ! `AssertionError: The message does not specify a sender and a default sender has not been configured`

A stacktrace appears, it seems that the `DEFAULT_MAIL_SENDER = ("Budget manager", "budget@notmyidea.org")` mentioned in default_settings.py is ignored.

Any ideas ?

JocelynDelalande:
My bug report was inacurate, this has nothing to do with **TESTING** setting , but with #121.

I close this one, the hot spot being on the pull request #121.

PR #121, opened 2015-04-30
nicely deprecates DEFAULT_MAIL_SENDER 

JocelynDelalande:
… as it changed upstream for flask-mail>=0.8

Note that it breaks default installation (as reported in #120).

For the record, this change has been made very quiet at flask-mail, see https://github.com/mattupstate/flask-mail/issues/51


ametaireau:
Thanks!

PR #122, opened 2015-05-01
Made an URL prefix configurable in settings

JocelynDelalande:
For information, I made that change to be able to package a ihatemonney app for  [yunohost project](https://yunohost.org/).

Flask handles the cookie scoping to the url prefix via the [APPLICATION_ROOT](flask.pocoo.org/docs/0.10/config/) setting.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -48,7 +48,12 @@ To deploy it, I'm using gunicorn and supervisord::
 Don't forget to set the right permission for your files !
 
 Also, create a `settings.py` file with the appropriate values if you need to
-use a different database for instance.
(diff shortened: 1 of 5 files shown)


ametaireau:
Thanks for the changes, sorry I missed them I'm not sure how. It seems to miss tests in order to be something we can merge. Can you write that?

JocelynDelalande:
@ametaireau reworked the PR totally so that it's testable (it was impossible to test as it was written). Does it sounds good to you ?

JocelynDelalande:
@almet  bump ?

almet:
Yup, that's great! Thanks.

PR #123, opened 2015-05-01
specified sizes for string columns for MySQL

JocelynDelalande:
Otherwise, the mysql sqlalchemy connector complains (cause VARCHAR2 of MySQL should have a max size defined).

For information, I made this change to package ihatemoney for [yunohost](https://yunohost.org)


ametaireau:
LGTM.

PR #124, opened 2015-05-20
No amount field autocomplete

QuentinRoy:
Avoid autocompleting when entering new bill amount (useless and dull: one inadvertently cancels the bill when canceling the autocomplete proposition using the 'escape' key).

### DIFF
--- a/budget/templates/forms.html
+++ b/budget/templates/forms.html
@@ -81,7 +81,7 @@
     {{ input(form.date, class="datepicker") }} 
     {{ input(form.what) }} 
     {{ input(form.payer) }} 
-    {{ input(form.amount) }} 
+    {{ input(form.amount, autocomplete="off") }} 
     {{ input(form.payed_for) }} 
     </fieldset>
     <div class="actions">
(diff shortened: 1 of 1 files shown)


Glandos:
If cancelling the autocomplete cancels the form, then this behavior should be fixed. But disabling autocompletion will annoy me.

ametaireau:
To me, It seems that autocompleting amounts isn't something we really need.

@Glandos can you maybe tell us what your actual use case is to see how removing autocompletion will impact it?

QuentinRoy:
The only use case I see where it may be useful is if you repeatedly enters the same bill.
However, it can only save time for 100+ bills: the autocomplete is only triggered after the first digit is entered. Once it is entered, entering the second will take less or as time as selecting the *first* autocomplete proposition.

Glandos:
Using only the mouse, you can simply click in the field to trigger the autocompletion menu (at least in Firefox), so you can enter a previous bill (e.g. a recurrent one) without leaving the mouse.

QuentinRoy:
Good point. So there is a tradeoff, I don't know what it the best solution.
I think I still prefer no autocomplete as I don't use this functionality (it actually doesn't work like that on my browser apparently) and even if 'escape' does not cancel the bill anymore, having to press it each time to cancel the autocomplete is boring.

Glandos:
I also understand your use case. But I can't think of anything more useful to comply with both use cases.

ametaireau:
> except when it actually cancels the bill too

This actually seems to be another bug that we should adress :)

Issue #94, opened 2012-10-23
Add user coefficients.


Korri:
We currently use an App on Android, and is has a realy cool feature, you can assign a default percentage to each user (default and per transaction), so for example we have a couple here so they have a default percentage of 200% as they share the same money (And like that we don't need to split every purchase between the two of them).

Having this feature could be really cool.

ametaireau:
Where are we with this? @aavenel do you want to work on it?

ametaireau:
Another possibility is to add this coefficient information to the list of people (e.g. that would work for couples).

This seems to minimize the complexity (no change when adding the new bills) and could only when editing the users (not sure exaclty how to do that, though).

JocelynDelalande:
+1 for the proposition of coefficients on users, I think it's a rather light feature in terms of UI, for same reasons that @berteh mentioned.

Another use case, which is quite complicated to handle without software and fits nicely with user coefficients :

- John, Mag and Bob go on holidays together
- John stays 3 days, Mag 2 days and Bob 2 days, they log their bills on IHM.

Assigning weights based on the days people stay is a nice way to handle that (doing it with spreadsheet is not straightforward).

ametaireau:
I've updated the title of the issue. I think the other proposal (in-bill coefs) is a no-go based on the feedback we currently have.

JocelynDelalande:
I started working on it.
As it implies schema modification, one question that remains IMHO before merging is how to handle migrations, see https://github.com/spiral-project/ihatemoney/issues/83

JocelynDelalande:
UI proposal for users list

![screenshot-localhost 5000 2015-08-20 11-41-28](https://cloud.githubusercontent.com/assets/429633/9380467/aa7d09ac-4730-11e5-9a7f-88913156720e.png)

I suggest hiding the coefficient if everyone is x1

aavenel:
Looks good for me, but how do you define the coefficient for each user ?

JocelynDelalande:
> Looks good for me, but how do you define the coefficient for each user ?

Coming later, work in progress :)

JocelynDelalande:
UI proposal for the users list to access member form
![screenshot-localhost 5000 2015-08-22 10-40-53](https://cloud.githubusercontent.com/assets/429633/9423202/9c477fb2-48ba-11e5-9bfe-e57a4cb9f770.png)

zorun:
I guess this issue can be closed, it has been implemented by #131 :)

[...]

Issue #83, opened 2012-03-08
Handle db migration?


feth:
Hello,

I'd like to try and add a feature to this app, that requires an sql schema migration.
Do you have an opinion about the best way to properly handle schema migrations?

Basically, all tools work the same and rely on a version field (integer) stored in a special table.
If not found, table and field are created (and version assumed to be 0).
If found, application compares the field with its own schema version, and if necessary, applies required migrations.

SqlAlchemy has its own tool: http://readthedocs.org/docs/sqlalchemy-migrate
They describe the generic steps involved: http://caneypuggies.alwaysreformed.com/wiki/DevelopingWithMigrations

Googling for "flask migration", I also found https://github.com/adamrt/flask-evolution/tree/master/flaskext -however I'm not sure what it is good for.

JocelynDelalande:
I'm not yet very familiar with flask, but I see two libs to do that :

- https://github.com/stackforge/sqlalchemy-migrate
- https://pypi.python.org/pypi/alembic

Both seems active, if someone more experimented on flask has an opinion on it, it would be handy :), as well as pointing out some software that nicely integrate those libs.

Coming from Django world, I find the migrations library for SQLA a bit disappointing :

- complicated setup
- complicated migration (eg: you seem to have to configure the target db by hand on each deployment)
- no automatic migration creation from models change

… But let's do with what exists :)

I thought for a while that we could do some DIY migrations but it would tend to be a nightmare, for example, IIRC, sqlite has no way to add a column without recreating the table.

JocelynDelalande:
any input on that ?

almet:
I won't invest time on this project to develop new features like this, unfortunately. Feel free to take it if you need it! I'll hapilly review and provide feedback if needed.

PR #131, opened 2015-08-27
Added Members weights handling

JocelynDelalande:
Implementing the feature discussed in #94 
UI/forms is quite basic (no ajax), but that seems consistent with the app.

It introduces schema modification, but there is not yet any mechanism to migrate the schema (see #83 ) so how do we do ?


almet:
I'm sorry, I've just seen this pull request right now, as I was doing some cleaning. I guess that #83 is a blocker of this very pull request, and nobody volonteered to implement it so far.

JocelynDelalande:
@almet Hurray ! That's now 

- using migrations provided by #141
- providing up-to-date translation files

budget/forms.py
```diff
@@ -152,27 +152,35 @@ def validate_amount(self, field):
...
         self.project = project
+        self.edit = edit
 
     def validate_name(form, field):
         if field.data == form.name.default:
             raise ValidationError(_("User name incorrect"))
-        if Person.query.filter(Person.name == field.data)\
-                .filter(Person.project == form.project)\
-                .filter(Person.activated == True).all():
+        if not form.edit and \
```
almet: Rather than using `\`, can you use parenthesis?

almet: Rather than using `\`, can you use parenthesis?

almet:
This looks good to me! (with small nits adressed)

JocelynDelalande:
@almet ok, nitfixed :-)

budget/models.py
```diff
@@ -217,9 +223,10 @@ def delete(self, project, id):
     archive = db.Column(db.Integer, db.ForeignKey("archive.id"))
 
     def pay_each(self):
-        """Compute what each person has to pay"""
+        """Compute what each share has to pay"""
 	if self.owers:
-		return self.amount / len(self.owers)
+                # FIXME: SQL might dot that more efficiently
```
almet: Extra space here.

almet:
Aaaand, it's merged!

PR #141, opened 2016-05-21
Alembic DB migrations

### LINKED ISSUE #83: Handle db migration?

JocelynDelalande:
fix #83 

I'm a bit frustrated with sqlalchemy migrations compared to django ones, but well… Flask-Migrations + Alembic seems the suckless solution at the moment.


JocelynDelalande:
@almet gentle bump :-) that's a blocker for #131 (that I will need to rewrite to take db migrations into account).

almet:
I have almost no experience with migrations for Flask, so let's consider this is the better solution right now :-)

budget/migrations/env.py
```diff
@@ -0,0 +1,87 @@
+from __future__ import with_statement
+from alembic import context
+from sqlalchemy import engine_from_config, pool
+from logging.config import fileConfig
+import logging
+
+# this is the Alembic Config object, which provides
+# access to the values within the .ini file in use.
```
almet: nit: please start comments with a capital and end them with a dot. Wrap them at 80 chars (not less).

budget/migrations/env.py
```diff
@@ -0,0 +1,87 @@
...
+config = context.config
+
+# Interpret the config file for Python logging.
+# This line sets up loggers basically.
+fileConfig(config.config_file_name)
+logger = logging.getLogger('alembic.env')
+
+# add your model's MetaData object here
```
almet: nit: ditto

almet:
Cool, this looks good to me apart from the nitpicks. Thanks for tackling this.

JocelynDelalande:
@almet thanks for feedback :-).

I nitfixed (note: that was default/generated code from *flask-Migrations*), is it better like that ?

almet:
This is now in master.

PR #144, opened 2016-06-18
Fix migrations upgrade path for MySQL

JocelynDelalande:
For some reason, the migration path from unmanaged db (from alembic
point-of-view) to managed db, through the initial migration works well with
sqlite… But not with mysql where the db system tries to re-create the existing
tables.

This commit is a way to detect if we are migrating from pre-alembic era and
skip the first migration (which would do nothing anyway), marking it as already
executed.

It's quite hackish but that's the best I found so far to get it working with
both MySQL and SQLite.

### DIFF
--- a/budget/run.py
+++ b/budget/run.py
@@ -3,7 +3,7 @@ import warnings
 
 from flask import Flask, g, request, session
 from flask.ext.babel import Babel
-from flask.ext.migrate import Migrate, upgrade
+from flask.ext.migrate import Migrate, upgrade, stamp
 from raven.contrib.flask import Sentry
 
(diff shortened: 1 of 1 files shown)


JocelynDelalande:
@almet bump. Without this PR, new installs are fine but upgrades are curently broken on master (I feel a bit ashamed about that :x).

budget/run.py
```diff
@@ -15,6 +15,15 @@
 app = Flask(__name__)
 
 
+def pre_alembic_db():
+    """ Checks if we are migrating from a pre-alembic ihatemoney
```
almet: Check

PR #151, opened 2016-09-13
Change ihatemoney.notmyidea.org → ihatemoney.org

almet:
As the old domain seams dead and not even redirecting.


almet:
I've updated all the links to be https.

PR #155, opened 2016-09-25
Fix unit test rounding

aavenel:
Test was Ok but failed due to some floating point errors.

### DIFF
--- a/budget/tests.py
+++ b/budget/tests.py
@@ -498,7 +498,10 @@ class BudgetTestCase(TestCase):
         result[models.Project.query.get("raclette").members[0].id] = 8.12
         result[models.Project.query.get("raclette").members[1].id] = 0.0
         result[models.Project.query.get("raclette").members[2].id] = -8.12
-        self.assertDictEqual(balance, result)
+        # Since we're using floating point to store currency, we can have some rounding issues that prevent test from working.
+        # However, we should obtain the same values as the theorical ones if we round to 2 decimals, like in the UI.
+        for key, value in balance.iteritems():
+            self.assertEqual(round(value, 2), result[key])
 
     def test_edit_project(self):
         # A project should be editable


JocelynDelalande:
@aavenel instead of fixing the test, would using `from __future__ import division` solve the issue ?

(it's a bit of a wild guess…)

budget/tests.py
```diff
@@ -498,7 +498,8 @@ def test_rounding(self):
         result[models.Project.query.get("raclette").members[0].id] = 8.12
         result[models.Project.query.get("raclette").members[1].id] = 0.0
         result[models.Project.query.get("raclette").members[2].id] = -8.12
-        self.assertDictEqual(balance, result)
+        for key, value in balance.iteritems():
```
JocelynDelalande: Maybe just add a comment to explain why you have to do that weird (but required) thing.

JocelynDelalande:
> @aavenel instead of fixing the test, would using from __future__ import division solve the issue ?
> 
> (it's a bit of a wild guess…)

Auto-answering me : no (just tested), so your way seem the right one :+1:

aavenel:
@JocelynDelalande The root cause is that we're using floating point to store currency instead of fixed point (Decimal in python). Replacing all the fields would probably be a big change so we're better off with a few rounding for now.

JocelynDelalande:
> @JocelynDelalande The root cause is that we're using floating point to store currency instead of fixed point (Decimal in python). Replacing all the fields would probably be a big change so we're better off with a few rounding for now.

The comment was not for me but to put above your code in source.

almet:
Hi, this is a friendly reminder: Alexandre, could you add the comment to the code in another commit? We would then be good to merge!

aavenel:
Added comment and rebased ! :)
All tests should be green !

JocelynDelalande:
All right, thanks @aavenel ; it's good to be green :-)

PR #156, opened 2016-09-25
Optim balances computation

aavenel:
Adding a new bill on a project with hundreds of bills is terribly slow.
This commit should improve performance, but it's still not as fast as I've hopped.

Strange thing : if I understand correctly, the query on `for bill in bills.all()` seems to be executed for each bill ??

I'm not really used to profile code on python, so please check if you see any regression :)

### DIFF
--- a/budget/models.py
+++ b/budget/models.py
@@ -35,10 +35,11 @@ class Project(db.Model):
             for time in (1, 2, 3))
 
         # for each person
+        bills_in_project = self.get_bills()
         for person in self.members:
             # get the list of bills he has to pay
-            bills = Bill.query.filter(Bill.owers.contains(person))
-            for bill in bills.all():
+            bills = bills_in_project.filter(Bill.owers.contains(person)).all()
+            for bill in bills:
                 if person != bill.payer:
                     share = bill.pay_each() * person.weight
                     should_pay[person] += share


JocelynDelalande:
> Strange thing : if I understand correctly, the query on for bill in bills.all() seems to be executed for each bill ??

Which query ?

Did you actually measured a performance progress ? Because the requesting spirit is the same before and after your patch (to my understanding). Using `get_bills` only adds some more `WHERE` SQL conditions (which may lead to small speed improvments I guess).

aavenel:
I've actually measured some progress (~50%) but my test case is not that good (something like 50 bills, only one project).
I think that in the `for` loop, the end condition seems to be evaluated at each step, causing the sql query to be computed multiple times ?
Without adding more SQL conditions, just by moving the bills.all() outside the loop, I've measured significant improvement.

JocelynDelalande:
> I've actually measured some progress (~50%) but my test case is not that good (something like 50 bills, only one project).

Ok. Out of curiosity ; could you just dump the SQL commands listing issued by each of the version of `balance()` (original/patched). (SQLAlchemy might offer a way to do that). It could offer a way to investigate why it's slow.

> I've actually measured some progress (~50%) 

That's still interesting :-)

> but my test case is not that good (something like 50 bills, only one project).

May be due to the WHERE clause.

> I think that in the for loop, the end condition seems to be evaluated at each step, causing the sql query to be computed multiple times ?

I don't think so, that's not how it is supposed to work in Python.

If you try with:
```
all_bills = bills.all()

for bill in all_bills:
    ...
```

It should produce same timings as original code.

JocelynDelalande:
@aavenel bump ?

zorun:
@aavenel can you check that #161 improves performance for you?  Can you still see a difference with your change?

aavenel:
Closing this, #161 fix this issue in a better way.

PR #158, opened 2016-10-19
Py3 port

FabriceSalvaire:


Glandos:
Is it backward compatible with python 2.7? It seems not, given the substitution from 'unicode' to 'str', but my question is more about Flask itself, since I don't know if those changes are for python 3.x only.

FabriceSalvaire:
Tests are broken for py3 and a256bca39854b197829c51d426c42084d5cc0637 should be rewritten so as to be backward compatible with py2.

Other commits are fine.

almet:
Thanks for taking the time to do this! I definitely want to have python 2.7 compatibility here, so if you manage to make it work, I'll gladly merge this.

FabriceSalvaire:
The incompatibility comes from the poor unicode support of Python 2. We could manage these few lines with:

```
import sys
if sys.version_info < (3,):
    ...
```

else we have to use helpers. I don't have so much experience on this since I stopped to write and support Python 2 codes.

almet:
That's one way to do it yes! I have no idea what's the best way to handle this, to be honnest. @natim is my python 3/2 expert here, so he might have good advice?

JocelynDelalande:
>else we have to use helpers. I don't have so much experience on this since I stopped to write and support Python 2 codes.

You way want to use six for that job :)

Natim:
> You way want to use six for that job :)

+1

JocelynDelalande:
@FabriceSalvaire gentle bump :)

almet:
Hi, thanks @FabriceSalvaire for your efforts here. Would you like to take this further, or should we close the issue instead? Or maybe someone would like to take this over?

FabriceSalvaire:
I don't have time to work on this actually. It could be probably achieved in few hours but unicode and 2/3 portability are headache stuffs.

JocelynDelalande:
@FabriceSalvaire Thanks for taking the time to explain you won't have time, that is appreciated.

Inaugurating a new label for that case to nofity people, this PR is looking for some love.

almet:
Superseded by #178

PR #162, opened 2017-01-02
Sort bills by (date.desc, ID.desc) instead of just date.desc

zorun:
When viewing the list of bills, bills are (correctly) sorted by date.  But
the order of all bills for a given day is not intuitive: I would expect
bills to be sorted by reverse order of insertion.  That is, the last bill
to be added for a given day should appear first, not last.  Otherwise,
when adding several bills in a row for a given day, it's confusing to see
that the new bills do not appear on top of the list.

Fix this by sorting by decreasing ID after sorting by date.


aavenel:
:+1: Agree, this is much better and the changes are straightforward.

PR #164, opened 2017-01-02
Fix zero-amount transfers and other rounding issues

zorun:
These are workarounds for rounding issues.

### DIFF
--- a/budget/models.py
+++ b/budget/models.py
@@ -61,9 +61,9 @@ class Project(db.Model):
         credits, debts, transactions = [],[],[]
         # Create lists of credits and debts
         for person in self.members:
-            if balance[person.id] > 0:
+            if round(balance[person.id], 2) > 0:
                 credits.append({"person": person, "balance": balance[person.id]})
-            elif balance[person.id] < 0:
+            elif round(balance[person.id], 2) < 0:
                 debts.append({"person": person, "balance": -balance[person.id]})
         # Try and find exact matches
         for credit in credits:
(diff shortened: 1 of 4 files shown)


budget/models.py
```diff
@@ -61,9 +61,9 @@ def get_transactions_to_settle_bill(self):
         credits, debts, transactions = [],[],[]
         # Create lists of credits and debts
         for person in self.members:
-            if balance[person.id] > 0:
+            if round(balance[person.id], 2) > 0:
```
aavenel: I think it'll be better to do the rounding in the `balance()` function instead.

zorun: This is what I was also thinking, but I decided against it since `balance()` seems to be used in a lot of other places.  Some calling functions may need the full floating-point precision to work.

It's a bit like unicode vs. bytestrings: always use unicode internally, and only decode/encode to UTF-8 bytestrings when handling inputs/ouputs.

aavenel: Ok, looks good to me.

JocelynDelalande:
Thanks a lot, welcome here :-)

:+1:  (tested & reviewed), although I don't know how to trigger the problematic cases, so my test is barely "it does not breaks the existing".

Should be easy to provide a test, if you have time, at least for 72f177afd1cd9103cd8fd3b98e9fcb5c695d5902 ?

zorun:
@JocelynDelalande good idea :)

I have added a test for the zero-transfer issue, this is the output before applying b507a5afa1b4e3594a3a35f3aae08f7f38989338 :

```
======================================================================
FAIL: test_settle_zero (__main__.BudgetTestCase)
----------------------------------------------------------------------
  File "tests.py", line 622, in test_settle_zero
    msg='%f is equal to zero after rounding' % t['amount'])
AssertionError: 0.003333 is equal to zero after rounding
```

almet:
Thanks :)

PR #166, opened 2017-01-09
Fix ServerTestCase.test_unprefixed test

### LINKED ISSUE #163: 'test_unprefixed' test fails
One of the test of the testsuite fails to pass:

```
======================================================================
FAIL: test_unprefixed (__main__.ServerTestCase)
----------------------------------------------------------------------
  File "tests.py", line 938, in test_unprefixed
    self.assertStatus(303, req)
  File "tests.py", line 614, in assertStatus
    "%s expected %s, got %s" % (url, expected, resp.status_code))
AssertionError:  expected 303, got 200
```

What's strange is that when I do the same query manually, I do get a 303 HTTP code:

```
~$ curl  -v localhost:5000/foo/
* Connected to localhost (127.0.0.1) port 5000 (#0)
> GET /foo/ HTTP/1.1
> Host: localhost:5000
> User-Agent: curl/7.52.1
> Accept: */*
> 
* HTTP 1.0, assume close after body
< HTTP/1.0 303 SEE OTHER
< Content-Type: text/html; charset=utf-8
< Content-Length: 251
< Location: http://localhost:5000/create?project_id=foo
< Server: Werkzeug/0.11.13 Python/2.7.13
< Date: Mon, 02 Jan 2017 12:31:34 GMT
```

@JocelynDelalande any idea?

JocelynDelalande:
The test was always failing, actual reason is the `app.run.configure()` fails
to reset the `APPLICATION_ROOT` setting which `ServerTestCase.test_prefixed`
overloads (side effect).

This patch *do not* fix app.run.configure as it seems uneasy, but takes a
different approach which has the advantage of making the test more explicit.

Would still be a good thing to investigate more on configure().

Fix #163

### DIFF
--- a/budget/tests.py
+++ b/budget/tests.py
@@ -934,6 +934,7 @@ class ServerTestCase(APITestCase):
         super(ServerTestCase, self).setUp()
 
     def test_unprefixed(self):
+        run.app.config['APPLICATION_ROOT'] = '/'
         req = self.app.get("/foo/")
         self.assertStatus(303, req)
 


almet:
LGTM.

JocelynDelalande:
@almet So… merge time ? :-)

JocelynDelalande:
@almet So… merge time ? :-)

almet:
Yep ! btw, I've given you admin access to the repository, in case I'm not around. Let me know if that works for you!

JocelynDelalande:
@almet ok, cool ; I will try use it sensibly :)

Issue #168, opened 2017-01-29
Update to bootstrap 4


JocelynDelalande:
we are currently using bootstrap2. Bootstrap 4 is currently alpha, but close to beta, which will be followed by a final release.

So I propose to jump directly to bootstrap 4 (of coulrse, we may wait for final release of BS4 to merge such a PR). BS4 introduce nice things like the use of flexbox for positioning.

I started to work on it ; nothing that I'd be proud enough to show code yet, but here are some screenshots, as apetizers ;-)

![screenshot-localhost 5000 2017-01-29 17-10-44](https://cloud.githubusercontent.com/assets/429633/22405626/f0fd5d90-e645-11e6-8a65-391c6e3065ab.png)
![screenshot-localhost 5000 2017-01-29 17-11-48](https://cloud.githubusercontent.com/assets/429633/22405627/f1007764-e645-11e6-8d62-1773e7934d62.png)

I'm trying to be quite neutral on that upgrade, not revamping the whole design, keeping the spirit of previous designs. But you'll notice some changes (see forms and their labels on home page), this is because I have the responsive version in mind that require some changes (see #113 ).

More comming soon…

aavenel:
I strongly approve this ! If I can help with something, let me know.

JocelynDelalande:
@aavenel I'll let you know when it will come to testing :)

Note that BS4 means dropping support for some old browsers like IE9.

JocelynDelalande:
Notes for myself, what will be changed with the BS4 PR:

- better factorization of templates
- fix moving elements (ex: try switching from project page to home page, things in sidebar are moving)
- improve readability
- fix misaligned elements
- make more use of BS grid system
- responsive design

I'm afraid my git log will be quite non-atomic :-s

almet:
That looks exciting, thanks for your work on this :)

JocelynDelalande:
@aavenel actualy, I'm pretty close to something on this topic, but responsive could come in a second time, so I won't include responsive things in this PR ; however it will be made easier by BS4.

So, to reconsider your question:

> I strongly approve this ! If I can help with something, let me know.

Is the adding of responsive code for small srceens (using BS4) something you would like to code ?

aavenel:
@JocelynDelalande I've close to no experience in this kind of development, but I can give it at try. Can you push a WIP PR so we can have a look at the code ?

almet:
This is now fixed by #172

Issue #113, opened 2013-10-12
Go responsive !


aavenel:
ihatemoney is currently broken on mobile display. It's quite unusable even just to add a bill.

With all the improvements in bootstrap 3.0, it should be possible to provide a better experience !

Any thoughts about this ?

ametaireau:
I believe it's quite easy to accomplish, it's just a matter of updating the theme in order to have the latest bootstrap which is responsive.

Vincent14:
I agree, I need this!

JocelynDelalande:
@Vincent14 Feel free to submit a PR :-)

almet:
In case anyone is wondering, we now have bootstrap 4 support, but are not fully rezponsive yet!

JocelynDelalande:
@aavenel still interested to responsibilize ihatemoney now that it's using BS4 ? :-)

Toover:
As a new user, I was also deceived by the impossibility to use the web app on my mobile with the default configuration. I had to display it as if I was a computer, and in landscape mode, to workaround the issue where we cannot add a bill. However I don't care about seeing a "responsive design": fixing the issue is all this web app needs.

almet:
Thanks for taking the time to comment in here, feedback is always valuable.

You're right, this is problematic. The thing is that currently, nobody volunteered to make this happen, which is why you're facing the issue you're describing. If you want to help fix the issue or if you know someone who will be willing to, don't hesitate to propose some code changes, I'll happily review and eventually merge them.

Toover:
I have had a look around the code and it seems that a good solution would be to let the navigation bar to be to stay on top of the page, instead of floating on top of the view. Another one would be to make sure all buttons of the navigation bar are dropdown to save space, so that the current space calculation is right. Is there one of the solutions better than the other for you?

almet:
thanks for the proposition, I don't understand the proposition about the dropdown in order to save space, though.

The idea to not have the navigation bar on top might work, yes! It might also be possible to hack something together that would let it on mobile, but thiner in order to not get in the way. What do you think?

Toover:
About the drop down: the top navigation bar was always expanded, so it was covering the forms at the top of the page (like "add bill"). I figured out that with "bootstrap", you it is quite easy to have a drop-down menu that is collapsed by default, so that it only covers the top of the page when needed.

[...]

PR #172, opened 2017-02-03
WIP: Upgrade to bootstrap 4

JocelynDelalande:
Upgrade to Bootstrap 4 (see #168 )

## Included
To summarize :

- better factorization of templates
- fix moving elements (ex: try switching from project page to home page, things in sidebar are moving)
- improve readability
- fix misaligned elements
- make more use of BS grid system
- better screen readers/aria support 

## Still on my TODO before it could get mergeable.

- [x] the calendar for date selection is appearing under the bill adding modal
- [x] fix datepicker language loading race condition
- [x] fix datepicker arrow icons
- [x] Add a scrollbar if the content is too big for the modal (intead of overflowing)
- [x] (chrome browser only) When you click on "add a bill" and then close the modal window, everything in the background shift a little bit.
- [x] Weird color changing of the font color when clicking « Add bill » button
- [x] Hovering a member « Reactivate » button shifts everything towards right on the sidebar
- other items reviewers will find :-)

## May come later in a separate PR

- responsive design (this PR only offers decent display for huge/medium display, not for small/tiny ones) @aavenel offered its help.
- rethink the navbar so that its contents is placed less weirdly (I tried to keep the previous philosophy, with minor adjustements).

Reviewers/testers are more than welcome :-) (poke @aavenel & others)


JocelynDelalande:
Finaly got a working datepicker :-)

… Anyway, I ran into something strange : the translation js file (for french translation of datepicker) seems not loaded yet when the `<script>` tag containing the call to `$('.datepicker').datepicker()` happens. Resulting in the calendar not being displayed in french.

I cannot manage to get sure that the translation is loaded *before* `$('.datepicker').datepicker()` is called.

What works is putting the content of the translation file directly in the `<script>` tag before `$('.datepicker').datepicker()`. But it's not something I want to do.

**Help apreciated** on this point :-)

aavenel:
Awesome work :)

I just did a quick review of the UI, here are some points : 

- The modal window for "Add a bill" is too short :
image

- When you click on "add a bill" and then close the modal window, everything in the background shift a little bit.

JocelynDelalande:
Ok, fixed my own bugs with datepicker (the bug was mainly me being stupid :-)

@aavenel thanks, added your bugs to my todo-list (the second seems a chrome-only one)

JocelynDelalande:
@aavenel I think my latest commit fixed your two issues ; could you confirm ?

Meanwhile, I found other CSS bugs…

JocelynDelalande:
@aavenel My TODO is empty :-) could you validate that it seems ok to you ?

@almet Would you have time for a second review before merging, just to get sure that I didn't go accidentally against some choices you made for example ?

Exciting!

almet:
I just tried it on my machine and it looks great. Everything seems a bit bigger, which I'm not sure we want or not. 

A few things that we might want to change (maybe not in this PR though):
- "flash" messages seem not really readable. We might want to use a different way to show them;
- On a small screen, the homepage doesn't fit in one screen (it did previously). Not sure if we want that. I like the idea of having everything on one page, but that doesn't really matter I guess?

In general, thanks for the work :)

JocelynDelalande:
Thanks for taking time to review :-)

> I just tried it on my machine and it looks great. Everything seems a bit bigger, which I'm not sure we want or not.

My feeling is that is seems a bit more in sync with current "fashions" (technically, is not only fashion, but also better usability on touch devices, more DPI on recent screens…). Side-note : I won't personally start modifying bootstrap defaults for margin/font-sizes/button-size/paddings… etc. 

Bootstrap allows to be lazy and to get something nice both to the eye and for ergonomy, that is its point of using it, isn't it ? :-)



> "flash" messages seem not really readable. We might want to use a different way to show them;

Just re-tested, I don't see how they are less readable than before, could you detail/screenshot the issue ?

> On a small screen, the homepage doesn't fit in one screen (it did previously). Not sure if we want that. I like the idea of having everything on one page, but that doesn't really matter I guess?

The homepage seems totally ok down to 1000px, which seems acceptable IMHO… At least for the moment, and until #113 gets done :). At what screen size did you find it non-fitting ?

aavenel:
Looks great ! Good to merge for me, just a few nitpicks here : 

On firefox, I've a strange thing on "Payer" field. There is 2 kind of arrows for the dropdown menu.
image

I've still some king of shifting when I click on "Add a bill" button. I think this is due to the scrollbar being hidden in modal view, such changing the width.

almet:
My screen is a 13'' with a resolution of 1366*768, but as you say, it should be fixed by #113 .

For the flash messages, I believe it's not introduced by your changes, I just noticed that it could be better. It should be handled in another issue, don't bother :)

Great work, thanks!

JocelynDelalande:
> My screen is a 13'' with a resolution of 1366*768, but as you say, it should be fixed by #113 .

weird, could you screenshot ?

almet:
selection_001
Here we are :)

JocelynDelalande:
> weird, could you screenshot ?

I genuinly don't see what is wrong on the screenshot :-/

almet:
Two things: there is a scrollbar, and the "getting started" button on the bottom right doesn't show up (not enough space).

almet:
here is the part that's not show on the first screenshot
selection_001

JocelynDelalande:
> here is the part that's not show on the first screenshot

Oook, it's a vertical issue :-). 

So, from my point of view, that vertical scrolling is fine, I'm afraid we disagree :x.

Are you ok for merge as-is and maybe find something else later if it's important to you ? :-/

almet:
> Are you ok for merge as-is and maybe find something else later if it's important to you ? :-/
Yes, sure ! As stated previously, I'm nitpicking here, thanks for your efforts :-)

almet:
This is now deployed at https://ihatemoney.org/ Thanks for your work :-)

JocelynDelalande:
Cool, thanks ! I'm proud to bring things to ihatemoney :))

@aavenel you turn on #113  ;-)

PR #174, opened 2017-02-16
Fix xss 173

### LINKED ISSUE #173: XSS vulnerability on list_bills
It seems that someone happily & successfully tested a XSS on the list_bills view. See screenshot.

xss

I disabled the XSS-ing buddy and removed it from the bills where it was on the demo project to avoid bothering visitors, but the bug remains.

aavenel:
Should fix #173 

I've rewritten `select_multi_checkbox` as an html template instead of a function. 
Arguments weren't properly escaped and this caused an XSS vulnerability.


JocelynDelalande:
@aavenel thanks a lot, tested & merged ! I also added a regression test to ensure it does not come back !

I think there are other possible XSS that we should hunt.

PR #178, opened 2017-03-13
Py2/3 compatibility, 2nd round

0livd:
Ihatemoney is now py2/3 compatible thanks to the six library
Based on what has been started in #158 

Porting the tests was a pain...

### DIFF
--- a/budget/models.py
+++ b/budget/models.py
@@ -255,11 +255,11 @@ class Bill(db.Model):
 
     def pay_each(self):
         """Compute what each share has to pay"""
-	if self.owers:
-                # FIXME: SQL might dot that more efficiently
-		return self.amount / sum(i.weight for i in self.owers)
-	else:
-		return 0
+        if self.owers:
+            # FIXME: SQL might dot that more efficiently
+            return self.amount / sum(i.weight for i in self.owers)
+        else:
+            return 0
 
     def __repr__(self):
         return "<Bill of %s from %s for %s>" % (self.amount,
--- a/budget/tests.py
+++ b/budget/tests.py
@@ -1,4 +1,5 @@
  # -*- coding: utf-8 -*-
+from __future__ import unicode_literals
 try:
     import unittest2 as unittest
 except ImportError:
@@ -8,6 +9,7 @@ import base64
 import os
 import json
 from collections import defaultdict
+import six
 
 os.environ['FLASK_SETTINGS_MODULE'] = 'default_settings'
 
@@ -23,7 +25,7 @@ class TestCase(unittest.TestCase):
         run.app.config['TESTING'] = True
 
         run.app.config['SQLALCHEMY_DATABASE_URI'] = "sqlite:///memory"
-        run.app.config['CSRF_ENABLED'] = False  # simplify the tests
+        run.app.config['WTF_CSRF_ENABLED'] = False  # simplify the tests
         self.app = run.app.test_client()
         try:
             models.db.init_app(run.app)
@@ -56,7 +58,7 @@ class TestCase(unittest.TestCase):
         })
 
     def create_project(self, name):
-        models.db.session.add(models.Project(id=name, name=unicode(name),
+        models.db.session.add(models.Project(id=name, name=six.text_type(name),
             password=name, contact_email="%s@notmyidea.org" % name))
         models.db.session.commit()
 
@@ -96,7 +98,7 @@ class BudgetTestCase(TestCase):
             response = self.app.post("/raclette/invite",
                                      data={"emails": "toto"})
             self.assertEqual(len(outbox), 0)  # no message sent
-            self.assertIn("The email toto is not valid", response.data)
+            self.assertIn("The email toto is not valid", response.data.decode('utf-8'))
 
         # mixing good and wrong adresses shouldn't send any messages
         with run.mail.record_messages() as outbox:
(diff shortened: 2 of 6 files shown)


almet:
Neat, thanks for your work on this :-) I would love to have someone else have a look at this code before we merge it (I'm bad with Python 3 ports ;)) @JocelynDelalande or @Natim, could you have a look? Thanks!

almet:
Okay, I'm leaving this open a few days more, and I'll probably merge it if I don't hear about anyone else, since it looks good to me :-)

budget/models.py
```diff
@@ -255,11 +255,11 @@ def delete(self, project, id):
 
     def pay_each(self):
         """Compute what each share has to pay"""
-	if self.owers:
-                # FIXME: SQL might dot that more efficiently
-		return self.amount / sum(i.weight for i in self.owers)
-	else:
-		return 0
+        if self.owers:
+            # FIXME: SQL might dot that more efficiently
```
Natim: might do

budget/tests.py
```diff
@@ -96,7 +97,7 @@ def test_notifications(self):
             response = self.app.post("/raclette/invite",
                                      data={"emails": "toto"})
             self.assertEqual(len(outbox), 0)  # no message sent
-            self.assertIn("The email toto is not valid", response.data)
+            self.assertIn("The email toto is not valid".encode(), response.data)
```
Natim: I would decode response.data rather than encoding the constant string.

budget/tests.py
```diff
@@ -669,18 +670,18 @@ def test_export(self):
...
         resp = self.app.post("/raclette/edit", data={
             'export_format': 'csv',
             'export_type': 'bills'
         })
-        expected = ["date,what,amount,payer_name,payer_weight,owers",
-                    "2017-01-01,refund,13.33,tata,1.0,fred",
-                    "2016-12-31,red wine,200.0,fred,1.0,\"alexis, tata\"",
-                    "2016-12-31,fromage à raclette,10.0,alexis,2.0,\"alexis, fred, tata, pépé\""]
-        received_lines = resp.data.split("\n")
+        expected = [u"date,what,amount,payer_name,payer_weight,owers",
```
Natim: I would rather import unicode_literals in the beginning: ``from __future__ import unicode_literals``

budget/tests.py
```diff
@@ -758,7 +759,7 @@ def api_add_member(self, project, name, weight=1):
     def get_auth(self, username, password=None):
         password = password or username
         base64string = base64.encodestring(
-                '%s:%s' % (username, password)).replace('\n', '')
+            ('%s:%s' % (username, password)).encode()).decode().replace('\n', '')
```
Natim: This encode() decode() doesn't looks right, it should be fixed once you import ``unicode_literals``

budget/utils.py
```diff
@@ -16,13 +17,14 @@ def slugify(value):
 
     Copy/Pasted from ametaireau/pelican/utils itself took from django sources.
     """
-    if type(value) == unicode:
+    if type(value) == six.text_type:
```
Natim: We should use ``isinstance`` instead

budget/utils.py
```diff
@@ -1,11 +1,12 @@
 import re
 import inspect
 
+from io import BytesIO, StringIO
```
Natim: I guess you need to use six there too. https://pythonhosted.org/six/#six.BytesIO

budget/utils.py
```diff
@@ -86,25 +88,33 @@ def list_of_dicts2json(dict_to_convert):
...
-    bytes_io.seek(0)
-    return bytes_io
+    return BytesIO(dumps(dict_to_convert).encode('utf-8'))
 
 def list_of_dicts2csv(dict_to_convert):
     """Take a list of dictionnaries and turns it into
     a csv in-memory file, assume all dict have the same keys
     """
-    bytes_io = BytesIO()
+    # CSV writer has a different behavior in PY2 and PY3
```
Natim: That's not true if you are opening the file with 'rb' or 'wb', we should use BytesIO in both cases.

almet:
Thanks a lot for the review @Natim ! @0livd if you have the time to look at these comments, please do, and soon enough we'll be python 3 compat on master :-)

0livd:
Thanks for the review :)
I'll make the changes

budget/tests.py
0livd: I think you saw encode().decode() and not encode()).decode(), no ? because for me there is no other way of doing this, the encode is necessary because base64.encodestring excepts bytes and the decode is necessary because bytes are returned and we want to use replace.

budget/utils.py
0livd: The thing is that in py3, even if you pass a file opened in binary mode to csv.writer and then try to write bytes with csv.writerow, you get a TypeError. 
```
Python 3.5.2 (default, Sep 14 2016, 11:28:32)
>>> import csv
>>> from io import BytesIO
>>> csv_file = BytesIO()
>>> writer = csv.writer(csv_file)
>>> writer.writerow([b"test", b"test2"])
  File "<stdin>", line 1, in <module>
TypeError: a bytes-like object is required, not 'str'
```
Six.BytesIO is not a good candidate neither because it is equivalent to StringIO in py2 and BytesIO in py3 and we need the opposite. So for me, there is no other choice than using StringIO in py3 and BytesIO in py2.

budget/tests.py
Natim: Oh I missed the base64 call :)

0livd:
Changes made, I also removed all the u"" since we are now importing unicode_literals

almet:
It's all good, thanks!

PR #179, opened 2017-03-13
Fix bad export filename

0livd:
Use project.id in export filename as it is already slugified

### DIFF
--- a/budget/web.py
+++ b/budget/web.py
@@ -225,7 +225,7 @@ def edit_project():
 
             return send_file(file2export,
                              attachment_filename="%s-%s.%s" %
-                             (g.project.name, export_type, export_format),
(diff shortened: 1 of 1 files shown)


almet:
Not sure why the tests are failing, it seems unrelevant to your changes.

0livd:
Yeah don't know why this was not failing before .... I actually fixed this failing test in #178 with 10a16a3b5cf13ca6ca2a1f060901b2bc943098ca . The first time I saw it failing was after 108970eb503f503b764408b0f93fde87d67a776d so I thought it was not executed because of the tab indentation but apparently it is something else ...

almet:
Neat, this can be rebased now :-)

almet:
Thanks again for your energy on the projet, that is very appreciated!

PR #188, opened 2017-03-28
Make travis use tox

JocelynDelalande:
So that we are consistent between dev test env and CI test env.
The use of `TOXENV` is to keep travis running the several envs in parallel.


almet:
Seems good to me!

PR #191, opened 2017-03-29
Add the possibilty to run ihatemoney via Apache mod_wsgi

0livd:
ihatemoney.wsgi is the entry point for mod_wsgi.
A virtualenv can be activated if its path is specified as
an env var in the apache virtual host file

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -33,28 +33,18 @@ You can also set the `TESTING` flag to `True` so no mails are sent
 Deploy it
 =========
 
+You have multiple options to deploy ihatemoney. Two of them are documented at the moment:
+
 With Apache and mod_wsgi
 ------------------------
 
-1. Install Apache and mod_wsgi::
+1. Install Apache and mod_wsgi - libapache2-mod-wsgi(-py3) for Debian based and mod_wsgi for RedHat based distributions -
 
-    $ sudo apt-get install apache2
-    # If you use Python3
-    $ sudo apt-get install libapache2-mod-wsgi-py3
-    # Or if you use Python2
-    $ sudo apt-get install libapache2-mod-wsgi
-
-2. Create an Apache virtual host based on the sample configuration file
-
-    $ sudo cp conf/apache-vhost.conf /etc/apache2/sites-available/ihatemoney.conf
+2. Create an Apache virtual host based on the sample configuration file in conf/apache-vhost.conf
 
 3. Adapt it to your paths and specify your virtualenv path if you use one
 
-4. Activate the virtual host and reload Apache ::
-
-    $ sudo a2ensite ihatemoney
-    $ sudo service apache2 restart
-
+4. Activate the virtual host if needed and restart Apache
 
 With Nginx, Gunicorn and Supervisord
 ------------------------------------
--- a/README.rst
+++ b/README.rst
@@ -33,7 +33,31 @@ You can also set the `TESTING` flag to `True` so no mails are sent
 Deploy it
 =========
 
-To deploy it, I'm using gunicorn and supervisord.
+With Apache and mod_wsgi
+------------------------
+
+1. Install Apache and mod_wsgi::
+
+    $ sudo apt-get install apache2
+    # If you use Python3
+    $ sudo apt-get install libapache2-mod-wsgi-py3
+    # Or if you use Python2
+    $ sudo apt-get install libapache2-mod-wsgi
+
+2. Create an Apache virtual host based on the sample configuration file
+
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -33,7 +33,31 @@ You can also set the `TESTING` flag to `True` so no mails are sent
 Deploy it
 =========
 
-To deploy it, I'm using gunicorn and supervisord.
+With Apache and mod_wsgi
+------------------------
+
+1. Install Apache and mod_wsgi::
+
+    $ sudo apt-get install apache2
```
almet: It seems that on new debian versions you can use `apt install` directly, should we propose that instead?

almet: We should specify that these instructions work for debutuntu.

conf/apache-vhost.conf
```diff
@@ -0,0 +1,17 @@
...
+    ErrorLog /var/log/apache2/ihatemoney.example.com_error.log
+    CustomLog /var/log/apache2/ihatemoney.example.com_access.log combined
+<Directory /path/to/ihatemoney>
+    WSGIProcessGroup ihatemoney
+    WSGIApplicationGroup %{GLOBAL}
+    Order deny,allow
+    Allow from all
+</Directory>
+Alias /static/ /path/to/ihatemoney/budget/static/
+</VirtualHost>
```
almet: I don't really know apache how vhost feature works, but it looks legit.

README.rst
0livd: Maybe the best here is to stay distro agnostic and just say 'Install Apache and activate mod_wsgi' and maybe just add a "e.g. libapache2-mod-wsgi(-py3) for Debian based and mod_wsgi for RH based distributions" because everyone knows how to install apache :)

almet: That looks like the best option, yes!

conf/apache-vhost.conf
```diff
@@ -0,0 +1,17 @@
+<VirtualHost *:80>
+    # Uncomment the following line if you use a virtualenv
+    # SetEnv IHATEMONEY_VENV_PATH /path/to/your/venv
+    ServerAdmin admin@example.com
+    ServerName ihatemoney.example.com
+    WSGIDaemonProcess ihatemoney user=www-data group=www-data threads=5
```
JocelynDelalande: *disclaimer: that may be a silly suggestion, as I'm not an apache user ;-)*

 Why not just using [`python-home=/path/to/venv`](http://modwsgi.readthedocs.io/en/develop/configuration-directives/WSGIDaemonProcess.html?highlight=python-home) instead of an environment variable ? This would simplify *ihatemoney.wsgi*. Calling *activate* in *ihatemoney.wsgi* seems a bit weird to me.

README.rst
```diff
@@ -33,7 +33,31 @@ You can also set the `TESTING` flag to `True` so no mails are sent
 Deploy it
 =========
 
-To deploy it, I'm using gunicorn and supervisord.
```
JocelynDelalande: The Maybe add some introduction so that  the reader understands she is facing an alternative. Something like 

> You have numerous options to deploy ihatemoney. Two of them are documented at the moment:

conf/apache-vhost.conf
0livd: Ah nice didn't know this one ! That would make ihatemoney.wsgi less hackish yes

0livd:
@almet @JocelynDelalande I pushed some changes to use the python-home parameter and updated the readme, should be good to merge :)

ihatemoney.wsgi
```diff
@@ -0,0 +1,9 @@
+import sys
+import os
+
+__HERE__ = os.path.dirname(os.path.abspath(__file__))
+
+# Add the budget directory to the path so we can then import from run
+sys.path.insert(0, os.path.join(__HERE__, 'budget'))
```
JocelynDelalande: Once #193 is ok, and when we will be pypi released, that `sys.path` extension will become useless. Maybe you could add some reminder comment about it in the source code.

README.rst
```diff
@@ -33,7 +33,21 @@ You can also set the `TESTING` flag to `True` so no mails are sent
...
+You have multiple options to deploy ihatemoney. Two of them are documented at the moment:
+
+With Apache and mod_wsgi
+------------------------
+
+1. Install Apache and mod_wsgi - libapache2-mod-wsgi(-py3) for Debian based and mod_wsgi for RedHat based distributions -
+
+2. Create an Apache virtual host based on the sample configuration file in conf/apache-vhost.conf
```
almet: nit: use double back-quotes here for files.

PR #193, opened 2017-03-29
Externalize the settings

### LINKED ISSUE #187: Have a way to externalise the settings
Currently, the settings file needs to be in the same directory as the source, which is a problem when you want to package it properly.

We've yet to find a solution for this, any help is welcome on the topic :)

0livd:
Default settings from app root path are loaded first
Settings are then overriden by /etc/ihatemoney/ihatemoney.cfg
or by another file which path is set in an env var
Fixes #187 

Readme should be updated if this gets merged


0livd:
yup it's nice to see that there is such a dynamic around a little project like this :)

Yes it's correct, maybe we can ditch the environment var thing and just try to load /etc/ihatemoney/ihatemoney.cfg but in this case I think we should create this directory by default on a production installation

almet:
I actually quite like the current approach with the env settings, as it enables some customisation for the users :-) My question was more about testing here :-)

0livd:
Oh ok I thought you said having multiple ways to load settings was too complicated :)

I'll try to get some inspiration from the flask test suite to write some tests
https://github.com/pallets/flask/blob/master/tests/test_config.py

JocelynDelalande:
> Yes it's correct, maybe we can ditch the environment var thing and just try to load /etc/ihatemoney/ihatemoney.cfg but in this case I think we should create this directory by default on a production installation

No, in many cases, the user installing ihatemoney won't have the right to create something in /etc/.

0livd:
@JocelynDelalande @almet I added some comments a configuration section in the readme, what do you think ?

0livd:
It might also be a good idea to make the instance path configurable so we can change it in a dev environment

0livd:
@almet I added some tests for this new configuration process, there are still 2 todos related to the default config file though

README.rst
```diff
@@ -43,13 +33,43 @@ To deploy it, I'm using gunicorn and supervisord.
...
-use a different database for instance. You can also set `APPLICATION_ROOT` if
-you want to prefix your URLs to serve ihatemonney in the *folder* of a domain,
-e.g:
+Configure it
+============
+
+In a production environment
+---------------------------
+
+Make a copy of budget/default_settings.py and name it ihatemoney.cfg.
```
almet: nitpick: use double backquotes here to make file locations appear inside `pre` html tags. This applies to all file locations.

README.rst
```diff
@@ -43,13 +33,43 @@ To deploy it, I'm using gunicorn and supervisord.
...
+Then adjust the settings to your needs and move this file to
+`/etc/ihatemoney/ihatemoney.cfg`.
```
almet: nit: you should use double backquotes here instead of single ones.

budget/run.py
```diff
@@ -27,8 +27,18 @@ def pre_alembic_db():
...
     """
-    config_obj = os.environ.get('FLASK_SETTINGS_MODULE', 'merged_settings')
-    app.config.from_object(config_obj)
+    default_config_file = os.path.join(app.root_path, 'default_settings.py')
+    config_file = os.environ.get('IHATEMONEY_SETTINGS_FILE_PATH')
+
+    # Load default settings first
+    # Then load the settings from the path set in IHATEMONEY_SETTINGS_FILE_PATH var
+    # If not set, default to /etc/ihatemoney/ihatemoney.cfg
+    # If the latter doesn't exist no errors are raised and the default settings are used
```
almet: nit: no error *is* raised.

budget/tests.py
```diff
@@ -10,14 +10,18 @@
 from collections import defaultdict
 import six
 
-os.environ['FLASK_SETTINGS_MODULE'] = 'default_settings'
-
 from flask import session
 
+# Unset configuration file env var if previously set
```
almet: Good catch ;)

almet:
This looks good to me!

If possible, we should address the TODO before merging to not add technical debt to the code, but all the rest looks great to me :-)

0livd:
I won't have time to push the last changes to the 2 pending PRs in the coming week so maybe you should merge them as is if you want to move forward in the meantime

almet:
Okay, thanks for letting us know, and thanks again for your work on the project!

almet:
I fixed the conflict in the README file.

0livd:
I rebased the PR and fixed the backquote nitpicks. For the TODOs, I need to create a fake ``ihatemoney.cfg`` so I'd like to do it after #196 is addressed

JocelynDelalande:
@0livd I don't understand how it depends on #196. Did you mean #193 ?

(Welcome back here:-))

0livd:
@JocelynDelalande These tests would add a new file :  ``budget/ihatemoney.cfg``, I would rather wait until we decide where to put the tests and put it directly in ``test/ihatemoney.cfg``. If I put it in the ``budget`` directory, users might think it is something actually used by the application.

0livd:
No more TODOs ! :)

almet:
Thanks! Seeing these changes, we probably should bump the major when releasing because it will break compatibility. Would that be okay for you? (cc @JocelynDelalande)

Issue #194, opened 2017-04-04
`make release` is broken on dependencies


JocelynDelalande:
Just tried to issue a `make release` (see https://github.com/spiral-project/ihatemoney/issues/180#issuecomment-290907313).

It's failing with :
```
[...]
Installing collected packages: zest.releaser, tox, colorama, six, setuptools, twine, virtualenv, pluggy, py, appdirs, packaging, clint, requests, requests-toolbelt, pkginfo, pyparsing, args
  Found existing installation: setuptools 5.5.1
    Uninstalling setuptools:
      Successfully uninstalled setuptools
  Running setup.py install for clint
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/home/jocelyn/dev/ihatemoney/.venv/lib/python3.4/site-packages/setuptools/__init__.py", line 12, in <module>
        import setuptools.version
      File "/home/jocelyn/dev/ihatemoney/.venv/lib/python3.4/site-packages/setuptools/version.py", line 1, in <module>
        import pkg_resources
      File "/home/jocelyn/dev/ihatemoney/.venv/lib/python3.4/site-packages/pkg_resources/__init__.py", line 72, in <module>
        import packaging.requirements
      File "/home/jocelyn/dev/ihatemoney/.venv/lib/python3.4/site-packages/packaging/requirements.py", line 9, in <module>
        from pyparsing import stringStart, stringEnd, originalTextFor, ParseException
    ImportError: No module named 'pyparsing'
    Complete output from command /home/jocelyn/dev/ihatemoney/.venv/bin/python3 -c "import setuptools, tokenize;__file__='/tmp/pip-build-xcx4i2et/clint/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-8h3dwuaw-record/install-record.txt --single-version-externally-managed --compile --install-headers /home/jocelyn/dev/ihatemoney/.venv/include/site/python3.4:
    Traceback (most recent call last):

  File "<string>", line 1, in <module>

  File "/home/jocelyn/dev/ihatemoney/.venv/lib/python3.4/site-packages/setuptools/__init__.py", line 12, in <module>

    import setuptools.version

  File "/home/jocelyn/dev/ihatemoney/.venv/lib/python3.4/site-packages/setuptools/version.py", line 1, in <module>

    import pkg_resources

[...]

PR #197, opened 2017-04-10
Install venv before using pip

almet:
Related to #194 even though I'm not sure it solves the problem. In any case it's good to fix this :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ TEMPDIR := $(shell mktemp -d)
 
 all: install
 install: $(INSTALL_STAMP)
-$(INSTALL_STAMP):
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -9,7 +9,7 @@ TEMPDIR := $(shell mktemp -d)
 
 all: install
 install: $(INSTALL_STAMP)
```
Toover: in my other commit, I propose rather to add virtualenv (the target) here instead. Do you know which one is better?

JocelynDelalande:
Sadly, does not fix #194.

almet:
Lets use the other fix instead, thanks :-)

PR #201, opened 2017-04-15
Fix makefile: Python lookup and virtualenv setup

Toover:
An alternative to the PR "fix-makefile", with an additional fix for looking up Python.


almet:
Thanks

PR #202, opened 2017-04-15
Update the README with additional requirements and information

Toover:
Without this information, I really could not understand where to go next.


almet:
Thanks !

PR #203, opened 2017-04-15
Add a missing file to prevent a systematic 404 error

Toover:
Bootstrap 4 was apparently not properly installed.


Toover:
The 404 appears in the JavaScript console. I admit I don't know the point of that file, but a 404 response simply looked like a problem to me, so I fixed it.

JocelynDelalande:
> The 404 appears in the JavaScript console. I admit I don't know the point of that file, but a 404 response simply looked like a problem to me, so I fixed it.

@Toover I'm against merging as-is (at most, we are hiding something, but fixing nothing).

So, what I suggest,  either :

- you have time and want to read about map files and you modify the PR to use them fully (it's pretty easy, only a matter of including files from parent projects), 
- or we close this PR.

Anyway, thanks for taking care of details and giving love to ihatemonney, that is appreciated :-)

almet:
I concur with @JocelynDelalande here, and I believe we should probably open another pull request with the map files if that's what we want.

I'm going to close this now, but I want to thank you again for taking the time to do the proposition changes here, that **is** appreciated.

Issue #108, opened 2013-08-06
Minimalist index page


astorije:
It would be nice to be able to disable the demo project, as well as the public ability to create new projects. This could be handled in `settings.py`.
I would love to implement this feature but I prefer to have the general opinion about this. Any suggestions?

ametaireau:
So, if I understand correctly, the idea would be to only have one project, and disable the possibility to create new projects. Is this correct?

If yes, then I believe we should add new settings (`DISABLE_PROJECT_CREATION` and `DEFAULT_PROJECT_NAME`). and change the routing based on that, plus disable some endpoints depending on the settings.

@astorije do you still need this?

almet:
Gentle bump on this. Anyone would like to volunteer on this?

astorije:
Hi @almet! Sorry, there is an extremely low chance I'd be working on this in any foreseeable future as I'm already involved in way more projects than I can handle at the moment.
Anyone willing to contribute, I'd be happy to discuss/add details/give 2c if that's any useful.

Feel free to close this if you think it's not a reasonable idea anymore.

0livd:
I'd like to work on this one but I'm not really sure of the implementation. In my opinion the main issue is that if you host ihatemoney on a public server, anyone can create new projects and manage their expenses on your instance. As @astorije said, this could be solved by adding a new parameter in settings.py so that new project creation will require a password (also specified in settings.py). It could also automatically disable the demo project. What do you guys think ?

0livd:
I made a WIP commit to show how i would implement it. This adds 2 new settings, PUBLIC_SERVER and PUBLIC_SERVER_PASSWORD. When PUBLIC_SERVER is set to True, the server password must be provided to create a new project and the demo project is no longer a special project.
Here PUBLIC_SERVER is set to True for the example but it should set to False by default.
Tests are broken in this commit
Any comment on this ?

almet:
Hi, thanks for taking the time to work on this, I believe it will be useful to others :)

Here are a few comments:

I find the name "public server" unclear. A name like "AUTHORIZE_PROJECT_CREATION" seems to better  explain what the intent is.

to be honest, I'm not so clear about what we are trying to achieve exactly here. If we want to have a way for the administrator to control the creation of new projects, I see different options in order to accomplish just this:

- Have an email sent to the administrator when a new project is created, and having him validate them manually;
[...]

PR #206, opened 2017-05-04
WIP Public project creation and admin permissions

0livd:
This is a wip PR based on what has been proposed in #108, it adds the possibility to remove the demo project and the public project creation, it also adds the possibility to protect any endpoint with an admin password (only ``/create`` is protected for the moment). We still don't have any admin dashboard but it make its future addition easier.

a5856a3b0565f6e1791273290fdebb0dc6b5cfcd adds a decorator (@require_admin) that will ask for ADMIN_PASS to access any decorated endpoint. If ADMIN_PASS is left empty in the settings(which it is by default) the decorator has no effect.

84033f9e7786fc47805df7ef68f08827e25bf269 Introduce PUBLIC_PROJECT_CREATION setting. When set to False (True by default), the project form on the home page is the replaced by a link to ``/create``, which is protected by @require_admin and the ``Try out the demo`` button is removed. As stated above, if ADMIN_PASS is left to the default value @require_admin has no effect so ``/create`` will still be publicly available

The empty admin password trick is the only way I found to remain compatible with the previous behavior, if someone has a more elegant solution...

I'll write some tests if it doesn't get rejected :)


budget/default_settings.py
```diff
@@ -8,3 +8,7 @@
 SECRET_KEY = "tralala"
 
 MAIL_DEFAULT_SENDER = ("Budget manager", "budget@notmyidea.org")
+
+ADMIN_PASS = ""
```
almet: nit: `ADMIN_PASSWORD`.

0livd:
Actually I'm not really satisfied with this ``PUBLIC_PROJECT_CREATION`` setting because if one forget to set ``ADMIN_PASSWORD`` public project creation is still possible and it does 2 unrelated things at the same time: it deactivates the demo project and the project creation form on the home page.

I have a better design in mind, I'll submit another PR so we can compare both approaches

0livd:
Closing this in favor of #210 and #209

PR #209, opened 2017-05-13
Add ACTIVATE_DEMO_PROJECT setting

0livd:
When set to False (True by default), it deactivates
the demo project

### DIFF
--- a/budget/web.py
+++ b/budget/web.py
@@ -121,8 +121,11 @@ def authenticate(project_id=None):
 def home():
     project_form = ProjectForm()
     auth_form = AuthenticationForm()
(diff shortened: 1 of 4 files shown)


budget/web.py
```diff
@@ -121,8 +121,11 @@ def authenticate(project_id=None):
 def home():
     project_form = ProjectForm()
     auth_form = AuthenticationForm()
+    activate_demo_project = current_app.config['ACTIVATE_DEMO_PROJECT']
```
almet: nit: we could call this `is_demo_project_activated`.

0livd:
@almet Change pushed

almet:
Merged, thanks!

PR #210, opened 2017-05-13
Public project creation and admin permissions

0livd:
Add a @require_admin decorator to protect specific endpoints with ``ADMIN_PASSWORD``(a password that is stored unencrypted in the settings). The decorator has no effect if ``ADMIN_PASSWORD`` is an empty string (default value)
The create project endpoint is the only one to be protected by @require_admin at the moment. When ``ADMIN_PASSWORD`` is not empty, project creation form on the home page will be replaced by a link to the create project endpoint so one is able to enter the admin password before filling the form


budget/web.py
```diff
@@ -66,6 +80,23 @@ def pull_project(endpoint, values):
...
+    if request.method == "POST":
+        if form.validate():
+            if form.admin_password.data == current_app.config['ADMIN_PASSWORD']:
+                session['admin_password'] = form.admin_password.data
+                session.update()
+                return redirect(goto)
+            else:
+                msg = _("This admin password is not the right one")
+                form.errors['admin_password'] = [msg]
+    return render_template("authenticate.html", form=form, admin=True)
```
almet: I find it a bit confusing to have a variable named `admin`. Should we call it `is_user_admin` or `is_admin` instead?

budget/web.py
```diff
@@ -66,6 +80,23 @@ def pull_project(endpoint, values):
                     url_for(".authenticate", project_id=project_id))
 
 
+@main.route("/authenticate_admin", methods=["GET", "POST"])
```
almet: Should the route be just `admin`? Otherwise, the URL could be joined with `-` rather than `_` (it seems more URLish to me)

budget/tests/tests.py
```diff
@@ -366,6 +368,21 @@ def test_authentication(self):
...
+        self.assertIn('<a href="/authenticate_admin?goto=%2Fcreate">', resp.data.decode('utf-8'))
+
+        # test right password
+        resp = self.app.post("/authenticate_admin?goto=%2Fcreate", data={'admin_password': 'pass'})
+        self.assertIn('<a href="/create">/create</a>', resp.data.decode('utf-8'))
+
+        # test wrong password
+        resp = self.app.post("/authenticate_admin?goto=%2Fcreate", data={'admin_password': 'wrong'})
+        self.assertNotIn('<a href="/create">/create</a>', resp.data.decode('utf-8'))
+
```
almet: Adding a test which validates what happens in case of an empty password would be great.

budget/web.py
```diff
@@ -119,13 +150,21 @@ def authenticate(project_id=None):
 
 @main.route("/")
 def home():
-    project_form = ProjectForm()
+    # If ADMIN_PASSWORD is empty we consider that we are not in admin mode
+    admin = bool(current_app.config['ADMIN_PASSWORD'])
```
almet: I honestly find that this logic is a bit odd — Maybe this is because of the naming. We might want to name the `admin` variable something else, like `is_admin_mode_enabled` or `admin_mode`.

budget/web.py
```diff
@@ -119,13 +150,21 @@ def authenticate(project_id=None):
...
     auth_form = AuthenticationForm()
+
+    if admin:
+        # If we are in admin mode we don't need to pass a project form to home
```
almet: The comment is a bit confusing, I thought we were in a different view than "home". Maybe state "We don't need to pass the project form to the template".

budget/web.py
```diff
@@ -119,13 +150,21 @@ def authenticate(project_id=None):
...
+        return render_template("home.html", auth_form=auth_form, session=session, admin=admin)
```
almet: Do we actually gain something out of not passing the project_form to the template? If not, it might be more interesting to keep the logic simple and pass it without displaying it instead.

budget/web.py
0livd: IMO ``admin`` would be a better fit for an administration page but since we already have a ``/dashboard`` that we could also use for administrative tasks, it makes sense to use ``admin`` here

budget/web.py
0livd: A few KB of memory :D ...yep you're right we'd better keep the logic simple

budget/web.py
almet: Gotcha. Do whatever you feel is best!

0livd:
I've just pushed the changes :)

almet:
This looks good to me, but it seems to have conflicts with the current master. Could you rebase?

0livd:
Yep rebase done

almet:
And we are done, thanks again!

PR #213, opened 2017-05-30
More responsive layout

Phyks:
This PR fixes the lack of responsive layout on mobile.

It should be ok on mobile devices for homepage and the main page of a given project, so that at least users could add new bills from a mobile device. Navbar on mobile device could still be enhanced though.

### DIFF
--- a/budget/static/css/main.css
+++ b/budget/static/css/main.css
@@ -74,7 +74,12 @@ body {
     background-repeat: no-repeat;
     height: 100%;
     color: black;
-    position: fixed;
+}
+
+@media (min-width: 768px) {
+    .sidebar {
+        position: fixed;
+    }
 }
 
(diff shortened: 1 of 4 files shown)


almet:
Hi, and thanks for this! It will be helpful to a bunch of people, I'm sure :) I've reviewed and it seems to work like a charm. Let's merge !

Phyks:
Thanks!

JocelynDelalande:
@Phyks seems cool, thank you ! 

@almet should we close #113  ?

Phyks:
My patch makes IHateMoney "responsive" in the sense that it should now be usable on mobile devices. I would not consider fully responsive though, especially since mobile UI is still very similar to desktop one and not particularly tailored for the mobile abilities. Not sure what was exactly wanting #113.

PR #215, opened 2017-06-15
Tell Apache to add the budget directory to the python path

0livd:
The path no longer needs to be modified in the .wsgi file


almet:
Cool, this seems better :-)

PR #218, opened 2017-06-18
Move the wsgi file to the budget dir

0livd:
The wsgi file needs to live in the budget dir
if we want it to be installed by setuptools


almet:
Not really, I'll just trust you on this one. I'm not using Apache on my side :-)

Issue #180, opened 2017-03-18
start making releases


JocelynDelalande:
I was thinking, what about starting making tagged releases  (or even just tagging on github) ?

It's a bit of work, but it would :

- make the install process a bit easier
- simplify third-party packaging (I'm thinking about https://github.com/YunoHost-Apps/ihatemoney_ynh )
- force us to keep track of changes in a ChangeLog

What do you think ?

almet:
I'm in favor of releases as well :-)

JocelynDelalande:
We now have a mechanism for making realeases (#182) … But we still didn't produce any.

@almet ok if I release a 1.0 with current code ? I would use the tagging mechanism you set but I won't release this version on pypi (for reasons we discussed [before](https://github.com/spiral-project/ihatemoney/pull/182#issuecomment-289809693)).

To tell the full story, I need a tag/release to improve [the yunohost package for ihatemoney](https://github.com/YunoHost-Apps/ihatemoney_ynh)

almet:
I think the thing we need now is a roadmap :-)

if you need a release, make one, for sure! I would probably wait us to merge the settings change before 1.0 though. So maybe 0.9 ? And we can issue 1.0 shortly after once ready.

JocelynDelalande:
Ok, just released [0.9](https://github.com/spiral-project/ihatemoney/releases/tag/0.9) despite #194 .

I leave this issue open for a few days, to see if this release process is ok for now or if we have comments/improvement suggestion. 

I followed strictly what our README.md release section. The *zest.releaser* wizard is pretty clear (I answered *yes* everywhere except for pypi release).

What I'd like to improve on next release is signing (gpg) the git releases.

0livd:
Should we release version 1.0 ? Now that the design on mobile has been fixed and that #193 is merged

JocelynDelalande:
I'd vote yes, let's 1.0 :-)

almet:
Yes ! I'll let you the honor @JocelynDelalande, if that's okay for you?

JocelynDelalande:
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA512

Alexis Metaireau <notifications@github.com> wrote:
> Yes ! I'll let you the honor @JocelynDelalande, if that's okay
> for you?

With great pleasure, I will be able to update my my yunohost
package to back on that 1.0 right after :-)

-----BEGIN PGP SIGNATURE-----
Version: GnuPG v1

[...]

PR #222, opened 2017-06-20
Prepare changelog & contrib list for v1.0

JocelynDelalande:
Not that I want to micro-review every single thing, but as this is the first time we put something in the changelog, that's to be sure you guys are ok with that.

I inspired from http://keepachangelog.com/en/0.3.0/,

The items list is made by hand, largely inspired (and copy/pasted) from the commit log.

What do you think ?

relates #180

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,11 +3,27 @@ Changelog
 
 This document describes changes between each past release.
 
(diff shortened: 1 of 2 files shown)


almet:
I think this is great! I don't see anything more to say :-) Let's ship it!

PR #226, opened 2017-06-21
Rename delete to deactivate for members. Fixes #177

### LINKED ISSUE #177: Rename delete to deactivate
I suggest renaming 'delete' / 'supprimer' a person by 'deactivate' or 'disable' / 'désactiver' it. Otherwise, it is very hard to know that it is possible to reactivate it afterwards.
By the way, this feature is very nice!

almet:


almet:
Thanks for the review. 

However, I believe the message shouldn't be changed in both places. Both information are valuable, see http://github.com/spiral-project/ihatemoney/blob/512de82d4c2ba8c4f4f2f7f224d02edd871b8585/budget/web.py#L391-L394
I've updated the text to contain some more information about what "deactivated" means.

JocelynDelalande:
> However, I believe the message shouldn't be changed in both places. Both information are valuable, see in the code;

I agree.

So all good, tested, works as expected, merging, thanks :-)

almet:
Thanks for the merge!

PR #229, opened 2017-06-21
Make all imports relative

almet:
I believe it will ease our life when it comes to packaging.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ remove-install-stamp:
 update: remove-install-stamp install
 
 serve: install
-	cd budget; $(PYTHON) run.py
+	$(PYTHON) -m budget.manage runserver
 
 test: $(DEV_STAMP)
 	$(VENV)/bin/tox
--- a/budget/api.py
+++ b/budget/api.py
@@ -2,9 +2,9 @@
 from flask import Blueprint, request
 from flask_rest import RESTResource, need_auth
 
-from models import db, Project, Person, Bill
-from forms import (ProjectForm, EditProjectForm, MemberForm,
-                   get_billform_for)
+from .models import db, Project, Person, Bill
+from .forms import (ProjectForm, EditProjectForm, MemberForm,
+                    get_billform_for)
 
 
 api = Blueprint("api", __name__, url_prefix="/api")
--- a/budget/forms.py
+++ b/budget/forms.py
@@ -7,10 +7,11 @@ from flask_babel import lazy_gettext as _
 from flask import request
 
 from wtforms.widgets import html_params
-from models import Project, Person
 from datetime import datetime
 from jinja2 import Markup
-from utils import slugify
+
+from .models import Project, Person
+from .utils import slugify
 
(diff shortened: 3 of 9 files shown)


0livd:
In development mode, the ``make serve`` command doesn't work anymore. I thought it would be working with ``python3 -m budget.run`` inside a venv but I get the same error : 
>ihatemoney$ python3 -m budget.run
INFO  [alembic.runtime.migration] Context impl SQLiteImpl.
INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
  File "ihatemoney/budget/run.py", line 10, in <module>
    from .web import main, db, mail
SystemError: Parent module '' not loaded, cannot perform relative import

@almet How do you run it on your side ?

JocelynDelalande:
- Ok on the idea, that is elegant :-)
- I haven't tested myself (thanks @0livd !) Also, the CI is good for testing that kind of change (subtle py2 vs py3 stuff may happen with relative imports).
- By the way you switched to *py.test*. I'm not against, but then here is your TODO  ;-)
  - [ ] update dev-requirements.txt
  - [ ] update the contributor doc

almet:
Thanks you both for the review! I've updated the pull request accordingly.

@JocelynDelalande I don't understand your comment about the `dev-requirements.txt` file, as it's already updated. Also, the command in the tox env has now changed so `make test` continues to work. What is it missing? Which documentation should I update?

JocelynDelalande:
> @JocelynDelalande I don't understand your comment about the dev-requirements.txt file, as it's already updated. Also, the command in the tox env has now changed so make test continues to work. What is it missing? Which documentation should I update?

Nevermind, I missed the *dev-requirements.txt*. Regarding the `make test`, you're right. I forgot that this is the official way to run tests. So… all good for me.

JocelynDelalande:
@almet I let you merge because perhaps you want to handle #229 beforehand.

almet:
Yeah, I've got the problem locally as well, and it might cause breaking changes for people using the default configuration for the database location (which is indeed not recommended for production use).

Since we're introducing a bunch of backward incompatible changes, I think now is a good time to change the default location for the data directory. Defaulting to `/var/lib/ihatemoney/database.sqlite` seems like a sane default.

I would love to see some feedback from @JocelynDelalande on this.

JocelynDelalande:
> Since we're introducing a bunch of backward incompatible changes, I think now is a good time to change the default location for the data directory. Defaulting to /var/lib/ihatemoney/database.sqlite seems like a sane default.

Why not :) . But that is a quite big break, and that makes a git clone installation not usable right after.

So, given that for a production use, you *need* a custom config file (for secret key, at least), why not making */var/lib/ihatemoney/database.sqlite* path the default in our example settings file, and not in the default one ? (right now, default and example settings files are the same, but this may change).

That would keep dev/test setup straightforward, and provide sane defaults for production use.

That aside, I'm not sure this PR is the good place do discuss/implement that. Is it ?

almet:
I agree, this is not related to this very PR.

PR #231, opened 2017-06-21
Add meta viewport tag

### LINKED ISSUE #219: Need to unzoom on firefox mobile to add bills
Firefox for Android 53.0.2

When I want to add a new bill on firefox mobile, the popover is opening as expected, I enter somehhing in the "What?" field (producing an automatic zoom on the popover). Then when I want to enter something on another field (like the payer field), nothing is happening. I can't even validate or close the popover.

Solution found: unzoom to max, by unpinching on the screen, then target carefully the wanted field.

Phyks:
I've just realized it was missing a `meta` tag.

Still not perfect, but incidentally it fixes #219.

Footer can be truncated on small screen, I do not have quick fix for this :/

### DIFF
--- a/budget/templates/layout.html
+++ b/budget/templates/layout.html
@@ -4,6 +4,7 @@
 <head>
     <title>{{ _("Account manager") }}{% block title %}{% endblock %}</title>
     <meta http-equiv="content-type" content="text/html; charset=utf-8">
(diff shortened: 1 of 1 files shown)


almet:
Awesome :-) thanks !

Phyks:
Thanks for the merge!

PR #236, opened 2017-06-23
Use a hashed password for ADMIN_PASSWORD

### LINKED ISSUE #233: Hash the admin password
The admin password is currently stored in clear text in the settings, it would be more secure to hash it and maybe provide a makefile recipe to hash a given password. The (new) documentation would need to be updated too

0livd:
A generate_password_hash manage.py command is provided
Fixes #233

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,7 +8,7 @@ This document describes changes between each past release.
 
 ### Changed
 
-- **BREAKING CHANGE** Use a hashed ADMIN_PASSWORD instead of a clear text one, see [doc](https://ihatemoney.readthedocs.io/en/latest/installation.html#configuration) for migration (#236)
+- **BREAKING CHANGE** Use a hashed ``ADMIN_PASSWORD`` instead of a clear text one, ``./budget/manage.py generate_password_hash`` can be used to generate a proper password HASH (#236)
 
 
 1.0 (2017-06-20)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,7 +8,7 @@ This document describes changes between each past release.
 
 ### Changed
 
-- Use a hashed ADMIN_PASSWORD instead of a clear text one (#236)
+- **BREAKING CHANGE** Use a hashed ADMIN_PASSWORD instead of a clear text one, see [doc](https://ihatemoney.readthedocs.io/en/latest/installation.html#configuration) for migration (#236)
 
 
 1.0 (2017-06-20)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,9 @@ This document describes changes between each past release.
 1.1 (unreleased)
 ----------------
 
-- Nothing changed yet.
+### Changed
+
+- Use a hashed ADMIN_PASSWORD instead of a clear text one (#236)
 
 
 1.0 (2017-06-20)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,7 +3,7 @@ Changelog
 
 This document describes changes between each past release.
 
-1.1 (unreleased)
+2.0 (unreleased)
 ----------------
 
 ### Changed
--- a/budget/run.py
+++ b/budget/run.py
@@ -52,6 +52,15 @@ def configure():
         if not 'MAIL_DEFAULT_SENDER' in app.config:
             app.config['MAIL_DEFAULT_SENDER'] = DEFAULT_MAIL_SENDER
 
+    if "pbkdf2:sha256:" not in app.config['ADMIN_PASSWORD'] and app.config['ADMIN_PASSWORD']:
+        # Since v1.1
+        warnings.warn(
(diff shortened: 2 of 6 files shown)


JocelynDelalande:
@0livd thanks :-) Currently reviewing & testing.

docs/installation.rst
```diff
@@ -85,7 +85,9 @@ properly.
 +----------------------------+---------------------------+-----------------------------------------------------------------------------+
 | ACTIVATE_DEMO_PROJECT      |  ``True``                 | If set to `True`, a demo project will be available on the frontpage.        |
 +----------------------------+---------------------------+-----------------------------------------------------------------------------+
-| ADMIN_PASSWORD             |  ``""``                   | If not empty, the specified password must be entered to create new projects |
+|                            |  ``""``                   | If not empty, the specified password must be entered to create new projects.|
+| ADMIN_PASSWORD             |                           | It needs to be hashed with the following command :                          |
+|                            |                           |     budget/manage.py generate_password_hash                                 |
```
JocelynDelalande: I'd reword a bit : 

```rest
If not empty, the specified password must be entered to create new projects. 

To generate the propper password HASH, use ``./budget/manage.py generate_password_hash`` and copy its output into the value of *ADMIN_PASSWORD*.
```

0livd:
@JocelynDelalande I went for a dirty option b) but it should be enough IMO

budget/web.py
```diff
@@ -13,6 +13,8 @@
     render_template, request, session, url_for, send_file
 from flask_mail import Mail, Message
 from flask_babel import get_locale, gettext as _
+from werkzeug.security import generate_password_hash, \
```
almet: nit: let's try to use parenthesis and carriage return instead of backslashes if possible :)

0livd:
@almet What kind of tests are you thinking about ? For me the tests performed by ``test_admin_authentication`` should be enough but I might be missing some other cases

budget/web.py
0livd: It might better to make another PR to remove all the backslashes instead of using 2 different syntaxes ?

almet: Works for me!

almet:
@0livd sorry, I though you provided an auto-migration facility, but I misread. I think we're good to merge!

budget/run.py
```diff
@@ -52,6 +52,15 @@ def configure():
         if not 'MAIL_DEFAULT_SENDER' in app.config:
             app.config['MAIL_DEFAULT_SENDER'] = DEFAULT_MAIL_SENDER
 
+    if "pbkdf2:sha256:" not in app.config['ADMIN_PASSWORD'] and app.config['ADMIN_PASSWORD']:
+        # Since v1.1
+        warnings.warn(
```
JocelynDelalande: Stop me if I'm wrong, but the message seems misleading: it's not a primary matter of safety reasons: without using an hashed password, **the previously set password won't work anymore**. 
Do you agree me ?

JocelynDelalande:
I modified slightly the changelog to make clear it is a breaking change. Do we all agree that [semver](http://semver.org/) is a good thing for ihatemoney ? @almet @0livd  ? (I'm not sure we stated anything on this topic yet).

*side note: I wanted to do it as a PR, but github seems broken (getting 404 errors trying to do so), so I did direct commit, sorry. Feel free to delete/reword/squash my commit if you want.*

CHANGELOG.rst
```diff
@@ -8,7 +8,7 @@ This document describes changes between each past release.
 
 ### Changed
 
-- Use a hashed ADMIN_PASSWORD instead of a clear text one (#236)
+- **BREAKING CHANGE** Use a hashed ADMIN_PASSWORD instead of a clear text one, see [doc](https://ihatemoney.readthedocs.io/en/latest/installation.html#configuration) for migration (#236)
```
almet: I think this is a good idea to state it's a breaking change. I would also write a "how to migrate" here, without having to follow a link.

CHANGELOG.rst
```diff
@@ -8,7 +8,7 @@ This document describes changes between each past release.
```
almet: It also seems that you didn't bump to 2.0 just yet.

JocelynDelalande:
>  ### Changed
>  
> -- Use a hashed ADMIN_PASSWORD instead of a clear text one (#236)
> +- **BREAKING CHANGE** Use a hashed ADMIN_PASSWORD instead of a clear text one, see [doc](https://ihatemoney.readthedocs.io/en/latest/installation.html#configuration) for migration (#236)
> 
> I think this is a good idea to state it's a breaking change. I
> would also write a "how to migrate" here, without having to
> follow a link.

Why not. No strong opinion on that. @0livd, if you agree on that
suggestion, I let you edit. I definititely don't like to commit
on somebody else PR :x

0livd:
@almet @JocelynDelalande Modified the warning message and mentioned the ``generate_password_hash`` command in the Changelog

almet:
This seems to work. Let's merge!

JocelynDelalande:
Thanks :-)

PR #237, opened 2017-06-23
Include the wsgi files in the source distribution

0livd:
#218 moved the wsgi file to the budget dir
but it was still not distributed


CHANGELOG.rst
```diff
@@ -6,8 +6,9 @@ This document describes changes between each past release.
 1.1 (unreleased)
 ----------------
 
-- Nothing changed yet.
+### Fixed
 
+- Add *.wsgi in the manifest file (#237)
```
almet: I would rephrase ans say what problem this fixes. e.g. «Include the wsgi files in the source distribution».

setup.py
```diff
@@ -59,10 +59,6 @@ def read_file(filename):
       author_email='alexis@notmyidea.org',
       url='https://github.com/spiral-project/ihatemoney',
       packages=find_packages(),
-      package_data={'': ['*.rst', '*.py', '*.yaml', '*.po', '*.mo', '*.html',
```
almet: if I understand properly, if we add it in the `recursive-include` we don't have to have the `package_data` and `include_package_data` entries?

0livd: I'm trying to understand the difference between the entries in ``MANIFEST.in`` and the ``package_data`` entries, it's a real mess but using only the manifest file seems to be the recommended way to go (apparently ``package_data`` only works for binary distribution and not source distribution).

0livd: And I finally didn't remove the ``include_package_data`` entry as it is not clear if it used or not ...even though its works without

CHANGELOG.rst
almet: It's here in the CHANGELOG entry that I was thinking about the rephrasing (I've seen you changed the title of the issue instead).

setup.py
almet: Heh. It's MAGIC :p Thanks!

0livd:
Actually I think I am going to close this PR for the moment. I've just realized that this wsgi file should not only be used by apache but by all the wsgi servers that want to serve the application, including gunicorn. Turns out gunicorn needs this file to be a python module so we might be better off just renaming it to ``wsgi.py`` (I am still wondering why I named it ``ihatemoney.wsgi`` in the first place :D).
The ``package_data`` entry still needs to be removed though
@almet What's your opinion on this ?

almet:
With gunicorn, I'm not using the wsgi file you've provided, instead I'm running it as follows:

```path/to/your/app/venv/bin/gunicorn -c /path/to/your/app/conf/gunicorn.conf.py run:app``` and it works properly.

That put aside, it might be better to provide a `wsgi.py` module instead of what's currently proposed. don't hesitate to open another Pull Request with the changes to the `setup.py` file, and we can address the WSGI this in yet another PR.

I thought that the WSGI file was only needed for Apache, but if it's not the case, then I need to read some documentation to better understand what problem it solves. What makes you say it's needed for gunicorn?

0livd:
Actually my point was that it would be better to have only one entrypoint for wsgi servers. Here we are using 2 entrypoints: the wsgi file for apache and the run module for gunicorn (even though they point to the same app under the hood). If I remember correctly Apache doesn't particularly need a wsgi file, it only needs the app to be named ``application`` whereas gunicorn doesn't have any name constraint.

almet:
Yes, it would indeed be a good practice!

0livd:
Closing this for the moment, I'll open new PRs to address the different issues separately

PR #239, opened 2017-06-27
Remove unused option in the setup script

0livd:
Additionnal files to be distributed along
with the sources are already specified in
the MANIFEST.in file thus the package_data
entry in the setup script can be safely removed


almet:
Thanks!

PR #242, opened 2017-06-28
Show some respect to pep8 and fellow readers.

almet:
I've spent some time reading the code and updated a few aspects of it whenever it had sense:

- updated the documentation when links were missing;
- Used the `autopep8` command to ease the formatting of the code;
- Added a few information in the default settings module;

I believe these changes are pretty harmless, put aside the fact that there is a repetition between the docs and the default settings module.

Let me know what you think!


almet:
Done in #243 to avoid merging headaches.

PR #243, opened 2017-06-28
Absolute imports & some other improvements

almet:
So, right after merging the work I've done with relative imports (#229) I've started experiencing some weird issues with it, especially when setting up the `debug` flag to `True`.

After some investigation, it seems that absolute imports [would be more reliable](https://stackoverflow.com/a/28154841/147077).

So I started taking a stab at these, and expanded the work to rename the `budget` package to `ihatemoney`, added a `ihatemoney` command and factorized some logic along the way.

I'm sorry to submit a pull request with all these changes at once, but I think they are all related. I hope you'll find these useful!

Happy review :-)


almet:
Oh; I need to state that this is still a work in progress. I mostly published it as a pull request to get early feedback and notify fellow contributors that I started hacking on something about this.

ihatemoney/run.py
```diff
@@ -0,0 +1,143 @@
...
+            "The way Ihatemoney stores your ADMIN_PASSWORD has changed. You are using an unhashed"
+            + " ADMIN_PASSWORD, which is not supported anymore and won't let you access your admin"
+            + " endpoints. Please use the command './budget/manage.py generate_password_hash'"
+            + " to generate a proper password HASH and copy the output to the value of"
+            + " ADMIN_PASSWORD in your settings file.",
+            UserWarning
+        )
+
+
+def create_app(configuration=None, instance_path='/etc/ihatemoney',
```
0livd: This new ``create_app`` function needs to be called in the wsgi module too

almet:
I believe this is ready, but I would like to get a review from someone before merging, cc @Phyks @0livd @JocelynDelalande (thanks!)

Phyks:
I did not review https://github.com/spiral-project/ihatemoney/pull/243/commits/8229b41bff2a241dcad1c7eb3119cde48c2f96f3 and https://github.com/spiral-project/ihatemoney/pull/243/commits/62436da75586c7bed80b5ab56b0e7c8cf8c90d26 extensively, but the rest of the commits looks nice!

PR #244, opened 2017-06-29
Some power and visibility for the dashboard

0livd:
This is a big PR whose main purpose is to add the possibility to delete and edit any project from the dashboard.
* The first breaking change to be introduced is that the dashboard is now disabled by default. A new setting is added to activate it. When activated, one has to enter the ``ADMIN_PASSWORD`` to access the dashboard and a link to it is added in the navigation bar. A requirement for the edit and delete actions to work on any project was that an admin could access every project, this is now the case.

* The first commit (d9ae72f) is a redesign of the admin authentication mechanism. The current design where we rely on the ``ADMIN_PASSWORD`` being defined or not to ask for admin permissions is not flexible enough. We need to be able to ask admin permissions for each endpoint independently.
With this new design the @requires_admin decorator can be conditionally bypassed based on the value of an application setting. To keep the old behavior where it was possible to ask for admin permissions to access  ``/create`` a new setting ``ALLOW_PUBLIC_PROJECT_CREATION`` is introduced, it is used to bypass the admin authentication  on ``/create`` when set to True. The second breaking change to be introduced is that ``ADMIN_PASSWORD`` has now a non-empty default value, a warning will be displayed to users who have set the ``ADMIN_PASSWORD`` to an empty string in a custom setting file and the application will automatically set it to the default password until they change it in their setting file. The default admin password is written in clear text in the documentation.


CHANGELOG.rst
```diff
@@ -11,7 +11,16 @@ This document describes changes between each past release.
 - **BREAKING CHANGE** Use a hashed ``ADMIN_PASSWORD`` instead of a clear text one, ``./budget/manage.py generate_password_hash`` can be used to generate a proper password HASH (#236)
 - **BREAKING CHANGE** Turn the WSGI file into a python module, renamed from budget/ihatemoney.wsgi to budget/wsgi.py. Please update your Apache configuration!
 - Changed the recommended gunicorn configuration to use the wsgi module as an entrypoint
+- **BREAKING CHANGE** The default value of ``ADMIN_PASSWORD`` has changed. If you have a custom settings file which set ``ADMIN_PASSWORD`` to an empty string (""), the application will use the default admin password until you update your settings.
+- **BREAKING CHANGE** Admin privileges are required to access the dashboard
```
almet: I wonder: maybe we should add a section "Breaking changes" instead of repeating it? What do you think?

CHANGELOG.rst
```diff
@@ -11,7 +11,16 @@ This document describes changes between each past release.
 - **BREAKING CHANGE** Use a hashed ``ADMIN_PASSWORD`` instead of a clear text one, ``./budget/manage.py generate_password_hash`` can be used to generate a proper password HASH (#236)
 - **BREAKING CHANGE** Turn the WSGI file into a python module, renamed from budget/ihatemoney.wsgi to budget/wsgi.py. Please update your Apache configuration!
 - Changed the recommended gunicorn configuration to use the wsgi module as an entrypoint
+- **BREAKING CHANGE** The default value of ``ADMIN_PASSWORD`` has changed. If you have a custom settings file which set ``ADMIN_PASSWORD`` to an empty string (""), the application will use the default admin password until you update your settings.
```
almet: I'm commenting here because it summarizes the intent: I'm not sure I like this behaviour. I'm saying this because having a default password in here seems like a security issue: instances with the default password will be easy targets.

One potential solution is to issue a warning and don't have any administration tasks available until a password is defined.

budget/default_settings.py
```diff
@@ -11,4 +11,8 @@
 
 ACTIVATE_DEMO_PROJECT = True
 
-ADMIN_PASSWORD = ""
+ADMIN_PASSWORD = "pbkdf2:sha256:50000$jc3isZTD$b3be8d04ed5c2c1ac89d5eb777facc94adaee48d473c9620f1e0cb73f3dcfa11"
+
+ALLOW_PUBLIC_PROJECT_CREATION = True
+
+ACTIVATE_DASHBOARD = False
```
almet: We should probably name this better. What about `ACTIVATE_ADMIN_DASHBOARD`?

0livd:
> Should we have the ability for the bearer of the admin password to connect to any project?
> The first question might be something worthy, but I'm not sure what to think about it yet. What's the use case, by the way?

The primary use case is to be able to have a private instance with an administration page, see #108 . Having access to any project with an admin password has never been asked as a feature, I added it because it greatly simplifies the implementation of the delete and edit actions in the dashboard but it could be implemented otherwise

>How should we handle the default password?
For the default password, here is an idea: If no pass is defined (or if we have a default one) then no administration tasks are available. If a password is defined, then it must be entered to access protected areas.

Good idea and this is closer to the previous design so this won't introduce any breaking change concerning the admin password :)

almet:
> I added it because it greatly simplifies the implementation of the delete and edit actions in the dashboard but it could be implemented otherwise.

Thanks for the precision :-) I think we should wait to have some more input from other contributors before going ahead. I know how frustrating this can be, but I believe it's for the greater good !

In the meantime, don't hesitate to introduce the change we've discussed about the default password.

0livd:
Changes pushed !

JocelynDelalande:
&lt;off topic&gt;
Hi all, thanks for your work discussions <3.

Just to let you know I probably won't take time to read/review/discuss pull requests next weeks.

Happy summer :-).
&lt;/off-topic&gt;

almet:
I took your changes and created a branch with them, available at #247. Let's continue the discussion there.

almet:
(The reason being that I didn't wanted to let you with all the rebase work to do, so I just did it but I cannot push to your branch, so here we go).

PR #247, opened 2017-07-06
Enhance the dashboard.

almet:
* Update to a more flexible admin authentication
* Admin can now access any project
* Add delete and edit options in the dashboard
* Add a link to the dashboard in the nav bar

This is a rework of the changes proposed by @0livd, so they can apply on top of
the latest master without trouble. All credit goes to him for the code.

Supersedes #244

PR #249, opened 2017-07-09
Protect admin endpoints against brute force attacks

### LINKED ISSUE #245: Protection against brute force attacks
Along with #244, it would be nice to have some throttling mechanism as a protection against brute force attacks, at least for the admin authentication

0livd:
Add a throttling mechanism to prevent a client brute
forcing the authentication form, based on its ip address
Closes #245 

When using multiple workers, the remaining number of attempts can get inconsistent but will still be limited to num_workers * max_attempts. The solution would be to use redis to share the state between the workers but I guess we'd rather keep it simple...


ihatemoney/translations/fr/LC_MESSAGES/messages.po
```diff
@@ -163,13 +163,25 @@ msgid "Export file format"
...
 msgid "This private code is not the right one"
 msgstr "Le code que vous avez entré n'est pas correct"
 
+#: web.py:106
+msgid "This admin password is not the right one. Only 2 attempts left."
+msgstr "Le mot de passe administrateur que vous avez entré n'est pas correct. Plus que 2 tentatives."
+
+#: web.py:106
+msgid "This admin password is not the right one. Only 1 attempts left."
+msgstr "Le mot de passe administrateur que vous avez entré n'est pas correct. Plus q'une seule tentative."
```
almet: typo: «plus qu'une »

ihatemoney/translations/fr/LC_MESSAGES/messages.po
```diff
@@ -163,13 +163,25 @@ msgid "Export file format"
...
-#: web.py:95
 msgid "This private code is not the right one"
 msgstr "Le code que vous avez entré n'est pas correct"
 
+#: web.py:106
+msgid "This admin password is not the right one. Only 2 attempts left."
+msgstr "Le mot de passe administrateur que vous avez entré n'est pas correct. Plus que 2 tentatives."
+
+#: web.py:106
+msgid "This admin password is not the right one. Only 1 attempts left."
```
almet: nit: is it possible to localize the beginning of the sentence just once?

ihatemoney/utils.py
```diff
@@ -131,3 +132,38 @@ def list_of_dicts2csv(dict_to_convert):
 
 # base64 encoding that works with both py2 and py3 and yield no warning
 base64_encode = base64.encodestring if six.PY2 else base64.encodebytes
+
+
+class LoginThrottler():
+    """Simple login throttler used to limit authentification attempts based on client's ip address
```
almet: It's « Authentication » not « Authentification ».

ihatemoney/utils.py
```diff
@@ -131,3 +132,38 @@ def list_of_dicts2csv(dict_to_convert):
...
+    When using multiple workers, remaining number of attempts can get inconsistent
+    but will still be limited to num_workers * max_attempts
```
almet: nit: finish your sentences with a dot.

ihatemoney/utils.py
```diff
@@ -131,3 +132,38 @@ def list_of_dicts2csv(dict_to_convert):
...
+    """
+    def __init__(self, max_attempts=3, delay=1):
+        self._max_attempts = max_attempts
+        # Delay in minutes before attempts counter is reset
```
almet: nit: «before resetting the attempts counter »

ihatemoney/utils.py
```diff
@@ -131,3 +132,38 @@ def list_of_dicts2csv(dict_to_convert):
...
+    def increment_attempts_counter(self, ip):
+        if self._attempts.get(ip) is None:
+            # Store first attempt date and number of attempts since
+            self._attempts[ip] = [datetime.now(), 0]
+        self._attempts.get(ip)[1] += 1
+
+    def is_login_allowed(self, ip):
+        if self._attempts.get(ip) is None:
+            return True
+        # Reset counter when the delay since first attempt is over
```
almet: We can rephrase this: « When the delay is expired, reset the counter. »

ihatemoney/utils.py
```diff
@@ -131,3 +132,38 @@ def list_of_dicts2csv(dict_to_convert):
...
+        self._delay = delay
+        self._attempts = {}
+
+    def get_remaining_attempts(self, ip):
+        return self._max_attempts - self._attempts.get(ip, [datetime.now(), 0])[1]
+
+    def increment_attempts_counter(self, ip):
+        if self._attempts.get(ip) is None:
+            # Store first attempt date and number of attempts since
+            self._attempts[ip] = [datetime.now(), 0]
```
almet: I'm a bit concerned with this behavior. I might be too worried for nothing, but this makes it possible to overwhelm the memory of the server by faking a lot of requests from different IP addresses. In order to fix this, we could free the memory (reset the dict) once we hit a big number of IP addresses (let's say 10k IPs ?). What do you think?

ihatemoney/web.py
```diff
@@ -89,14 +91,24 @@ def admin():
...
+                # Valid password
+                if (check_password_hash(current_app.config['ADMIN_PASSWORD'],
+                                        form.admin_password.data)):
+                    session['is_admin'] = True
+                    session.update()
+                    login_throttler.reset(client_ip)
+                    return redirect(goto)
+                # Invalid password
+                login_throttler.increment_attempts_counter(client_ip)
+                msg = _("This admin password is not the right one. Only %d attempts left." %
```
almet: [gettext](https://pythonhosted.org/Flask-Babel/#flask.ext.babel.gettext) can be passed the second argument directly to ease the work of translators.

almet:
Also this might be conflicting with #247, can you build your changes on top of this branch there? Thanks!

ihatemoney/utils.py
0livd: To fake random IP addresses, a client would manipulate the X-FORWARDED-FOR header. ProxyFix offers a protection against that because by default, if there are multiple proxies, it only trusts the headers set by the last one. In non-proxy setups, it means that the headers set by the client would be trusted and that the OOM killer might pay us a visit.
TLDR: Yeah let's add an auto-reset :)

0livd:
Changes pushed !

0livd:
>Also this might be conflicting with #247, can you build your changes on top of this branch there? Thanks!
Why not simply validate and merge this and then rebase #247 ?

almet:
Simply because I already took the time to rebase it once and I don't really want to do it again :) If you feel like doing it, please go ahead, that will be appreciated!

0livd:
Ok I'll do it but I won't have time in the coming weeks

almet:
Thanks a lot, and happy holidays (?) ;)

0livd:
>Thanks a lot, and happy holidays (?) ;)
Yep, flying to kirghistan, hope they have some decent vodka :)

0livd:
Back from the other side of the planet...Should we merge this? I'll then take care of the grunt work on #247.

almet:
I won't have time to look at this in the upcoming weeks. I'll be back at the end of August!

almet:
Yes, this looks good ! Lets merge :-)

PR #250, opened 2017-07-09
Revert configuration tests deletion

0livd:
Some supernatural power erased the configuration
tests, they're now back !


almet:
Thanks, I believe this is my fault !

PR #252, opened 2017-07-26
Streamline installation doc, now using pip

### LINKED ISSUE #238: Better document how to download ihatemoney
The recommended way to get the code in the installation doc is ``git clone https://github.com/spiral-project/ihatemoney.git`` which does not correspond to the latest stable release.
We could :

- tag a release as the ```latest release``` so ihatemoney could be downloaded following this link ``https://github.com/spiral-project/ihatemoney/releases/latest``
- provide the link to pypi ``https://pypi.python.org/pypi/ihatemoney``
- create a ``stable`` branch which points to the latest release
- keep ``git clone`` and add a ``git checkout -b stable 1.x`` step (sounds complicated...)

IMO ``git clone https://github.com/spiral-project/ihatemoney.git`` should go in the contributing section

JocelynDelalande:
Fix #238 

As discussed, pip will become the "official" installation tool for ihatemoney. Let's document that :-).

In the same row, I harmonized doc/conf files naming, it's the same goal: have a consistent installation procedure.

## Remaining issue

One issue remains, (I think it could be handled after this PR got merged, but before we release the next version) : static assets paths with setuptools packaging. I'm going to open an issue on this topic (later today, hopefuly), so that we discuss solutions.


almet:
I won't have time to look at this in the upcoming weeks. I'll be back at the end of August!

Issue #136, opened 2016-02-03
Display statistics as a graph


Natim:
It could be nice to have a summary tab with information about what everybody did pay and did spend as well as the total of expenses for the group.

almet:
What would that look like?

Natim:
![solde-demo](https://cloud.githubusercontent.com/assets/229453/12854164/5baa1206-cc38-11e5-8603-cba6a4a14ad7.png)

I was thinking of something like that.

almet:
That's really cool! I think we could even provide this information [as a graph](https://wpoely86.github.io/D3.js-Diverging-Stacked-Bar-Chart/) (Diverging stacked bar chart)

zorun:
Having a graph for statistics would be very nice, and probably not that hard to do, I can give it a try :)

I think this kind of simple line chart would be easier to read: http://www.chartjs.org/samples/latest/charts/line/basic.html
There would be one line for each participant, and one point per month showing the balance for the month.  There could be an option to show the cumulative balance instead of the "instantaneous" balance. The timescale could be changed by the user to display data by day, or week, or year.

One practical question though: would it be better to use the API with Ajax queries, or just write a view that generates the necessary data?
Generating javascript variables in a Jinja template (like [here](https://pythonspot.com/flask-and-great-looking-charts-using-chart-js/)) is simple but not very clean, while Ajax queries are cleaner but involve dealing with API authentication.

Natim:
@zorun thank you for your message and welcome to this project.

I don't think the project already have a REST API, so if you were to build one, it would be a bit of work. (CORS, Auth, utilities, documentation).

Also it might be a interesting project to build one.

For now, I guess having a django rendered view would be good enough.
I also wish to have a table explaining how much money users spend.

I am usually using a project per trip (with family and friends) and I don't need the monthly feature but more a total of all the expenses.

Last time I also wanted some metrics about how much was spend for the car, how much for the accomodation and how much for food and tourism.

zorun:
Hi @Natim 

There are already such statistics shown in a table, from #257 :) see for instance https://ihatemoney.org/demo/statistics

Regarding the API, I was thinking of using this one https://ihatemoney.readthedocs.io/en/latest/api.html

almet:
@zorun we never actually answered you! I think the best way to go is to integrate the Javascript in a view to be served directly by the python app. While this method is subject to discussion, it's the way all the project is currently built, so we better stick with it for clarity :-)

[...]

Issue #103, opened 2013-05-08
Add a stats tab


almet:
Having a "statistics" tab would probably be valuable, so we can see who spent what, and what's the evolution for the last months, for instance.

almet:
Duplicate of #136

PR #257, opened 2017-08-03
Add a statistics tab

0livd:
Add a basic statistics tab

ref #136 #103


almet:
this needs a small conflict resolution before merge, but it seems okay otherwise!

0livd:
Conflict resolved, I updated the changelog too :)

PR #258, opened 2017-08-06
Fix misplaced datepicker configuration

### LINKED ISSUE #256: Deleting a project
Hello,

When going to the edit options of the project on `https://ihatemoney.org/[name_of_the_project]/edit` and **triying to delete it throught the appropriate button**, nothing happens.

0livd:
Bootstrap-datepicker is only included in the
list_bills template but its configuration was
living in the layout template, leading to a
javascript error on every page except list_bills.

Fixes #256

### DIFF
--- a/ihatemoney/templates/layout.html
+++ b/ihatemoney/templates/layout.html
@@ -21,13 +21,6 @@
                 });
             }, 4000);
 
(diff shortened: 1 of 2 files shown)


almet:
Thanks for the debug :-)

PR #262, opened 2017-08-22
Enhance the dashboard - bis -

0livd:
* Update to a more flexible admin authentication
* Admin can now access any project
* Add delete and edit options in the dashboard
* Add a link to the dashboard in the nav bar

Supersedes #244 and #247.


almet:
I think we're almost ready to go. The only thing that bugs me is the fact that the administrator can access any project with its admin password. I think we should at least provide an option for this behavior in the settings, making it possible to disable the super admin password (but keeping the ability to connect to an administration dashboard).

Currently, it's not possible to want to deactivate the admin-password-for-all-projects *and* have an administrative dashboard at the same time.

I understand this is due to the fact that it's easier to implement, but before merging I would like to have the opinion of @all people implicated here. What do you think about the proposed design? Should we go ahead with this and merge?

almet:
@JocelynDelalande do you have any idea on this?

almet:
Okay, it's been a long time since we're waiting for this to be merged, and I think we should go ahead, and we can improve on this later on if needed.

JocelynDelalande:
Sorry, I was quite unavailable for reviewing that. Going ahead was the good option, thanks all :-)

PR #265, opened 2017-09-05
Normalize changelog

JocelynDelalande:
To bring some consistence / clarity to the changelog before releasing 2.0.

Please squash before merging, this meta-changes do not deserve such a big commit history :-)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,17 +6,20 @@ This document describes changes between each past release.
 2.0 (unreleased)
 ----------------
 
-### Breaking changes
+Breaking changes
+================
 
 - ``ADMIN_PASSWORD`` is now hashed rather than plain text. The ``ihatemoney generate_password_hash`` command can now be used to generate a proper password HASH (#236)
 - Turn the WSGI file into a python module, renamed from budget/ihatemoney.wsgi to ihatemoney/wsgi.py. Please update your Apache configuration!
 - Admin privileges are now required to access the dashboard
 
-### Changed
+Changed
+=======
 
 - Changed the recommended gunicorn configuration to use the wsgi module as an entrypoint
 
-### Added
+Added
+=====
 
 - Add a statistics tab (#257)
 - Add python3.6 support (#259)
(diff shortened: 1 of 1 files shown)


CHANGELOG.rst
```diff
@@ -6,34 +6,38 @@ This document describes changes between each past release.
...
-- If admin credentials are defined, they can be used to access any project.
-- It is now possible to edit and delete projects directly from the dashboard.
-- The dashboard can now be deactivated using the ACTIVATE_ADMIN_DASHBOARD setting.
-- When activated, a link to the dashboard appears in the navigation bar.
+- Statistics tab (#257)
+- Python3.6 support (#259)
+- ALLOW_PUBLIC_PROJECT_CREATION setting (#262)
+- Projects can be edited/deleted from the dashboard (#262)
+- ACTIVATE_ADMIN_DASHBOARD setting (#262)
+- Link to the dashboard in the navigation bar (#262)
```
almet: Thanks for doing this. Having the issue number is a good thing. However, it seems that some replaced items are clearer than their newer counterparts. Maybe a rule of thumb here could be to have a self-explanatory sentence at least?

almet: Ah, I understand now that this was done for consistency with the oldest changelog items. I think that's because implicitely I decided to make the items more descriptive. We probably should decide on what we want together. Waiting for your viewpoint on this :)

almet:
Thanks for taking the time to do these pull requests / housekeeping. Let me know what path you prefer for this one!

JocelynDelalande:
> Thanks for taking the time to do these pull requests / housekeeping. Let me know what path you prefer for this one!

Actually, I made this PR so that we discuss :-)

I personally prefer shorter sentences, changelog is not git log neither complete upgrade documentation (teaser: working on it :-)). 

Also, we have section names "*Added*" for example. They provide context, thus, IMHO, no need to say "Added link …" in the "Added" section.

almet:
I think you're right. Being able to parse this rather quickly is a good thing, and for any detail it's possible to look at the pull request. Let's go ahead with this :-)

Issue #251, opened 2017-07-26
Do not expose settings in an hard-to-read table


JocelynDelalande:
It's unreadable on most screens (evil horizontal scroll).

![image](https://user-images.githubusercontent.com/429633/28619706-f9b9c982-7209-11e7-8663-ceba1d6b2107.png)

Maybe we should use sections to present/explain the different settings.

Issue #266, opened 2017-09-05
Fix installation doc to avoid storing db in /tmp/


JocelynDelalande:
Default database location [is in /tmp](https://github.com/spiral-project/ihatemoney/blame/master/ihatemoney/default_settings.py#L11). Which is good for testing but bad for production.

We must mention the need to tweak this setting in installation documentation, as */tmp* often gets flushed on boot, this default is dangerous for production.

Has to be fixed prior to 2.0 release (#264)

almet:
Also, we could issue a warning in the console when the default database path is /tmp, with a link to the docs :-)

PR #267, opened 2017-09-05
Fix default SQLALCHEMY_DATABASE_URI value in doc

JocelynDelalande:
Reality-sync with default_settings.py

Not fixing the whole table layout because

- my text editor doesn't do it ;
- I'm too lazy ;
- #251 is IMHO the real way to this issue :)


JocelynDelalande:
(tiny relation to #266)

almet:
LGTM.

Issue #232, opened 2017-06-21
Use unique URIs to share projects


0livd:
A nice enhancement would be to be able to share a project with a link of the following form https://ihatemoney.org/projects/-uuid-, just like on framadate.org. IMO it would be a much better design than the actual one where we store an unhashed password in the database (even if there is a warning). For backward compatibility, we could keep the password authentication but it would be hashed in the database.
Related to #140

edit: readability

0livd:
@almet @JocelynDelalande Any thoughts on this ? I might be better off submitting a WIP PR to show what I have in mind...

almet:
Thanks for this proposition.

Why not having the ability to share a project with an unique URI, but I think that not having to remember anything that looks like an UUID is a better design. Having to remember a project name + password is something that currently works versus having to go to the emails to find the link that was used to share the project.

This would add a new feature to the existing feature set, and I would like to keep the overall number of features small (mainly because we're rather bad at UX).

Now, if we want to have this feature, we could add it to the current feature set, but I would not replace the current behaviour with it.

0livd:
The main idea behind this is to increase security by not storing unhashed passwords anymore. In the current design we store cleartext passwords to be able to :
- recover a lost password
- share a project by sending its credentials by email

The first one could be easily avoided by sending a token link to the user so he can reset his password via a dedicated form.
One way to avoid the second is sharing projects with an unique URI.

So yeah what I propose is not to add another feature but to replace the current behavior of the project sharing system. I am in favor of increased security even at the cost of a less convenient UX (going to the emails to find the project link).

almet:
Hi, I'm back :-) I dont like to break features like that. It's never à good idea for the users.

Why not having this as a way to share links ! As to storing passwords in clear, we can probably just change that and hash it with a salt, which would change less the existing feature. What do you think?

Also using a link sent by e-mail to connect seems as insecure as sending the password in clear, but is less easy to remember. 

Happy to be back, lets move this forward !

0livd:
>Why not having this as a way to share links ! As to storing passwords in clear, we can probably just change that and hash it with a salt, which would change less the existing feature. What do you think?

The thing is that if we want to hash the passwords in the database, we have to reimplement the 2 features that need clear text passwords.
[...]

PR #269, opened 2017-09-06
Use token based authentication to reset passwords

0livd:
Send a mail containing a password reset
token link instead of sending a clear text
password.

Ref #232


almet:
The approach seems good to me, we probably should go ahead with this.

Python 3.5 and python 3.6 seems to be failing, though. From the logs, I believe it's related to tox and an incompatibility with virtualenv. Since this was triggered a while ago, the conflicts might simply be gone, so I triggered another build to see if that makes any difference.

0livd:
Seems we are not the only ones hitting this issue, a work around is proposed here https://github.com/pytest-dev/pytest/issues/2779#issuecomment-334393688

ihatemoney/templates/password_reminder.fr
```diff
@@ -1,7 +1,7 @@
 Salut,
 
-Vous avez demandez des informations sur votre mot de passe pour "{{ project.name }}".
+Vous avez demandé à réinitialiser votre mot de passe pour "{{ project.name }}". Ce lien est seulement valide pendant30 minutes.
```
almet: missing a space here!

almet: In the french and english version, I believe the `it's only valid for X minutes` should be *after* the actual link.

ihatemoney/templates/reminder_mail.en
```diff
@@ -2,7 +2,7 @@ Hi,
 
 You have just (or someone else using your email address) created the project "{{ g.project.name }}" to share your expenses.
 
-You can access it here: {{ config['SITE_URL'] }}{{ url_for(".list_bills") }} (the identifier is {{ g.project.id }}),
+You can access it here: {{ url_for(".list_bills", _external=True) }} (the identifier is {{ g.project.id }}),
```
almet: Yes, it's better with the use of `_external=True`, thanks!

ihatemoney/translations/fr/LC_MESSAGES/messages.po
```diff
@@ -500,6 +532,10 @@ msgstr "Rappel du code d'accès"
 msgid "Your projects"
 msgstr "Vos projets"
 
+#: templates/reset_password.html:2
+msgid "Reset your password"
+msgstr "Changez votre de passe"
```
JocelynDelalande: typo

0livd:
@almet @JocelynDelalande Just pushed the requested changes. I extended the token lifetime to one hour.

almet:
Okay, since you've addressed all the comments from @JocelynDelalande , I'll take the liberty to merge this :) Thanks again for this, it'll be a lot better now without sending these passwords by email !

JocelynDelalande:
Excellent, thanks to you two!

PR #270, opened 2017-09-06
Make authentication logic simpler and safer

0livd:
The project password was set in clear text in the session cookie. The cookie payload is only base64 encoded so it must not be used to store private information. The password is simply replaced by a boolean.

The authenticate method is also refactored to be more readable.


CHANGELOG.rst
```diff
@@ -17,6 +17,7 @@ Changed
 =======
 
 - Logged admin can see any project (#262)
+- Simpler and safer authentication logic
```
almet: nit: When adding a line in the changelog, if possible specify the pull-request number.

0livd:
That's weird, tests are failing for py35 and py36 with some ``Error creating virtualenv.`` error...travis CI issue ?

almet:
I've restarted the builds, let's see.

almet:
I've restarted the builds once again, let's see if it's better :)

PR #271, opened 2017-09-06
Do not import the whole werkzeug package

0livd:


almet:
Thanks, that's the best practice actually :-)

PR #272, opened 2017-09-16
Add a DockerFile

0livd:
Proposition to add a DockerFile that can be used to deploy the latest version from PyPI in a production environment or from the master branch in a dev environment.


almet:
Thanks for the proposal. I have however no experience whatsoever with Docker files, so I wont be able to review this properly.

Let's summon some friends on this one. @bnjbvr @jankeromnes @Natim maybe do you have insights / can review this better than me? Thanks for your help :)

bnjbvr:
Nice work! Two questions:

- does it make sense to have the default sql backend be sqlite **and** the sqlite db to be stored in the `/tmp` directory? I think at restart and/or respawn, all the data would be lost. I'd rather have better defaults, with the sqlite database in a volume that's exposed to the host (it's also simpler then because one doesn't need to set up a whole database).
- is there a way to require the user setting some env variables? For instance, keeping the defaults for `MAIL_DEFAULT_SENDER` would be a spam disaster, as well as keeping the default `ADMIN_PASSWORD`.

0livd:
>does it make sense to have the default sql backend be sqlite and the sqlite db to be stored in the /tmp directory? I think at restart and/or respawn, all the data would be lost. I'd rather have better defaults, with the sqlite database in a volume that's exposed to the host (it's also simpler then because one doesn't need to set up a whole database).

Yeah you're right, I always tend to think that sqlite is only for quick test and dev but that's not true, I'll add an optional volume for it

>is there a way to require the user setting some env variables? For instance, keeping the defaults for MAIL_DEFAULT_SENDER would be a spam disaster, as well as keeping the default ADMIN_PASSWORD.

Hum, didn't see anything talking about that in the dockerfile doc...made me think that in the current state of the image, we are going to get an ``Error 500`` when trying to send mails with flask_mail because no local mail server is running...

almet:
Hey, do we have any progress on this?

0livd:
Hey, the main issue is the mail server. I've checked the ``flask_mail`` doc and it's not possible to use the system's mail client, a mail server address has to be specified. So we could :
* run a mail server with supervisord in the same container but it breaks the one app per container principle
* run a mail server in another container but that would require the use of docker-compose
* use some free external smtp service

bnjbvr:
Is it possible to use an external email service by providing a few variables (or by configuring it in the settings), through an SMTP user/password/host/scheme etc? If so, then the decision of running an host or docker container for SMTP would be left to the administrator, which seems more flexible and simpler than forcing a choice upon them.

0livd:
Yep it's possible, see https://pythonhosted.org/flask-mail/#configuring-flask-mail. Seems that letting the administrator decide is the way to go indeed.
So I'll just add the possibility to pass these extra variables and this should be good to merge. The doc update on how to configure the mail server should be done in another PR as it is not only docker related.

almet:
Thanks for your contribution on this topic, @bnjbvr, it's appreciated :)
@0livd I'll wait for your last changes and do another review of it all, and we'll be good to merge!

almet:
Thanks for your help on this topic, @bnjbvr, that's appreciated!
@0livd, I'll wait for your latest changes and we should be good to merge!

almet:
Thanks for your help on this topic, @bnjbvr, that's appreciated!
@0livd, I'll wait for your latest changes and we should be good to merge!

0livd:
I added the possibility to pass the mail server configuration and documented an example

almet:
Let's ship it, thanks to everyone involved :)

Issue #274, opened 2017-10-23
Add upgrade doc


JocelynDelalande:
see 

- https://github.com/spiral-project/ihatemoney/issues/264#issuecomment-327219245
- https://github.com/spiral-project/ihatemoney/issues/264#issuecomment-338497339

Issue #195, opened 2017-04-04
Better document SECRET_KEY tunning on a production instance.


JocelynDelalande:
Currently, we don't warn boldly people installing ihatemoney to change `SECRET_KEY` with a random string. Thus, blindly following README leads to insecure installation with a SECRET_KEY set to *tralala*.

We must improve doc on that topic.

almet:
What we did on kinto was to provide à tool to generate the settings file maybe we should do the same here ?

JocelynDelalande:
> What we did on kinto was to provide à tool to generate the settings file maybe we should do the same here ?

Sounds cool :-)

We could use a `manage.py` subcommand.

PR #275, opened 2017-10-23
Enhance install process by generating config files from templates

JocelynDelalande:
Motivations are:

- https://github.com/spiral-project/ihatemoney/issues/195#issuecomment-291607821
- frustration (re)writing installation and upgrade docs

Writing docs is a good exercise : if something is too much complicated to explain, maybe it is simply too much complicated.

So this is a simplification attempt of the install process.

This may be extended in the future (the script may ask questions or receive arguments), but that is out of the scope of this PR.

This is (IMHO) a requirement for #274 (put another, I ended up with this PR trying to write the PR for #274 :-p)

What do you think ?


almet:
Great, I think it's better this way too. One small adjustment might be to change the command name to `generate-config` as I tend to prefer verbs in the command actions.

Also, some people might prefer to copy/paste the settings files (even with the Jinja2 variable inside them), it might be good to point their location in the docs as well?

JocelynDelalande:
Thanks for fast answer !

> Great, I think it's better this way too. One small adjustment might be to change the command name to generate-config as I tend to prefer verbs in the command actions.

Done

> Also, some people might prefer to copy/paste the settings files (even with the Jinja2 variable inside them), it might be good to point their location in the docs as well?

I tend to think it bloats the documentation, but well… DONE also :-). I added a note with a link to the github folder.

ihatemoney/conf-templates/ihatemoney.cfg.j2
```diff
@@ -0,0 +1,36 @@
...
+# Enter an absolute path here.
+SQLALCHEMY_DATABASE_URI = 'sqlite:///var/lib/ihatemoney/ihatemoney.sqlite'
+SQLACHEMY_ECHO = DEBUG
+
+# Will likely become the default value in flask-sqlalchemy >=3 ; could be removed
+# then:
+SQLALCHEMY_TRACK_MODIFICATIONS = False
+
+# This secret key is random and auto-generated, it protects cookies and user sessions
+SECRET_KEY = "{{ secret_key }}"
```
almet: I especially like this as it's now possible to generate a secret key without even the user noticing it :)

ihatemoney/default_settings.py
```diff
@@ -1,37 +1,10 @@
...
-# documentation, available online at
-# http://ihatemoney.readthedocs.io/en/latest/installation.html#configuration
-
-# Turn this on if you want to have more output on what's happening under the
-# hood.
-DEBUG = False
-
-# The database URI, reprensenting the type of database and how to connect to it.
-# Enter an absolute path here.
+# Verbose and documented settings are in conf-templates/ihatemoney.cfg.j2
```
almet: Thanks for not duplicating this :)

PR #276, opened 2017-10-23
Attempt to fix Travis-CI

almet:


JocelynDelalande:
We are back to green, thanks :-)

PR #280, opened 2017-10-27
Use token based auth in invitation e-mails

0livd:
Invitation e-mails no longer contain the clear text project password.

There is a small regression in the sense that once someone has been sent a token link it is not possible to revoke his access even by changing the project password. I think it is an issue we can address later to make things go faster.


JocelynDelalande:
I have little time right now, I have not done a complete code reading, but just a quick question 

Does this PR means that we change the way we abandon the shared per-project password ?

0livd:
I forgot to reference #232 where we discussed this with @almet. So yeah it would mean that if users want to share the project password they have to do it via another communication mean.

almet:
I actually quite like the shared password thing, as it makes it very easy to use and share for everyone. If we decide to go for a different approach, I believe we should encourage the users to share the password using a different mean (ideally via voice).

As this is a big change in how the user flow works, I would like to get @QuentinRoy's input and at least someone else (@leplatrem ?).

0livd:
Yeah sharing passwords is for sure the easiest way to share projects but it is also a very big security issue. Imagine someone that creates a project with no intent of sharing it and without paying attention to the warnings. He might reuse his e-mail password, putting his own e-mail account at risk if the ihatemoney database gets hacked...

almet:
That's right :) I think we should take a stand against these practices !

I agree with your changes, but I also think that we should tell the users to share this passcode between members clearly, maybe in the email we send to them. In the user flow, we could also not send an email to everyone involved *but instead* tell the creator of the project to share the passcode with the members of his choosing.

If we decide to go this way, we could make the 2nd screen optional, and create a button in the UX to create a "sharing link". What do you think? 

Also, in my opinion, we should make a big difference between a password (that is unique and not shared) and a passcode (not sure about the denomination ?) that is shared among users of a same project.

JocelynDelalande:
> Also, in my opinion, we should make a big difference between a password (that is unique and not shared) and a passcode (not sure about the denomination ?) that is shared among users of a same project.

Maybe we could word it « shared password » everywhere.

Explicit is better than implicit ;-)

On global spirit agree with @almet that shared password −being a bad security practice− is easier to use for that kind of project. On the details… Well, I have little time and I I trust the collective you are :-).

almet:
Small bump on this: @0livd can we help you with anything?

0livd:
>I agree with your changes, but I also think that we should tell the users to share this passcode between members clearly, maybe in the email we send to them. In the user flow, we could also not send an email to everyone involved but instead tell the creator of the project to share the passcode with the members of his choosing.
If we decide to go this way, we could make the 2nd screen optional, and create a button in the UX to create a "sharing link". What do you think?

I think this is a good compromise between the two approaches :). I'll try to propose something going this way.

0livd:
Here is a new proposition that includes the following changes:

- Skip the invite page after project creation
- Replace ``The project identifier is demo, remember it!`` by ``Invite other people to join this project!`` (linking to the invite page)
- Encourage users to share the project password via other communication means in the reminder email

ihatemoney/templates/reminder_mail.fr
```diff
@@ -4,5 +4,7 @@ Vous venez de créer le projet "{{ g.project.name }}" pour partager vos dépense
 
 Vous pouvez y accéder ici: {{ url_for(".list_bills", _external=True) }} (l'identifieur est {{ g.project.id }}),
 et le code d'accès "{{ g.project.password }}".
+Si vous voulez partager ce projet avec vos amis, vous pouvez partager son identifiant et son mot de passe avec eux ou leur envoyer une invitation avec leline suivant :
```
almet: typo :)

ihatemoney/templates/reminder_mail.en
```diff
@@ -4,6 +4,8 @@ You have just (or someone else using your email address) created the project "{{
 
 You can access it here: {{ url_for(".list_bills", _external=True) }} (the identifier is {{ g.project.id }}),
 and the private code is "{{ g.project.password }}".
+If you want to share this project with your friends, you can share the identifier and the private code with them or send them invitations with the following link:
```
almet: We should name this "shared password" instead.

ihatemoney/templates/list_bills.html
```diff
@@ -92,7 +92,7 @@
 {% endblock %}
 
 {% block content %}
-<div class="identifier">{{ _("The project identifier is") }} <a href="{{ url_for(".list_bills") }}">{{ g.project.id }}</a>, {{ _("remember it!") }}</div>
+<div class="identifier"><a href="{{ url_for(".invite") }}">{{ _("Invite") }}</a> {{ _("people to join this project!") }}</div>
```
almet: Should all the text be the link?

ihatemoney/templates/send_invites.html
```diff
@@ -1,17 +1,15 @@
...
-    <li><a href="{{ url_for(".list_bills") }}">{{ _("Use it!") }}</a></li>
-</ol>
-{% endblock %}
 {% block content %}
 <h2>{{ _("Invite people to join this project") }}</h2>
 <p>{{ _("Specify a (comma separated) list of email adresses you want to notify about the
 creation of this budget management project and we will send them an email for you.") }}</p>
-<p>{{ _("If you prefer, you can") }} <a href="{{ url_for(".list_bills") }}">{{ _("skip this step") }}</a> {{ _("and notify them yourself") }}</p>
+<p>{{ _("If you prefer, you can share the project identifier and the project
+password by other communication means. Or even directly share the following link:") }}</br>
```
almet: "shared password"?

0livd:
Changes pushed :)

0livd:
Hey, what's up with this one ? If this is too big of a change we can close it for the moment and give it some more thoughts.

almet:
I actually just missed your last changes, sorry for that ! I'll review everything once again soon and I think I'll be able to merge this time :-)

Thanks for the efforts, and sorry for the delay !

almet:
Hey, from what I understand, there is only non-timed tokens used around, so I don't understand why we need a timed token concept here. Could you clarify? Thanks! Otherwise it looks good to me :)

almet:
Actually, I just got it. It's for the password reminder option you introduced earlier. That's all good to me, let's merge! Thanks :)

PR #282, opened 2017-10-30
Update installation doc to include doc about production values

### LINKED ISSUE #266: Fix installation doc to avoid storing db in /tmp/

Phyks:
Also add a warning at runtime if database is stored under `/tmp`.

This fixes #266.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -157,28 +157,43 @@ properly.
 Defaults given here, are those for development mode. To know defaults on your
 deployed instance, simply look at your *ihatemoney.cfg*.
 
+Production values are recommended values for use in production.
+
 
 +-------------------------------+---------------------------------+----------------------------------------------------------------------------------------+
 | Setting name                  |  Default                        | What does it do?                                                                       |
 +===============================+=================================+========================================================================================+
 | SQLALCHEMY_DATABASE_URI       | ``sqlite:///tmp/ihatemoney.db`` | Specifies the type of backend to use and its location. More information on the         |
 |                               |                                 | format used can be found on `the SQLAlchemy documentation`_.                           |
+|                               |                                 |                                                                                        |
+|                               |                                 | **Production value:** Set it to some path on your disk. Typically                      |
+|                               |                                 | ``sqlite:///home/ihatemoney/ihatemoney.db``. Do *not* store it under ``/tmp`` as this  |
+|                               |                                 | folder is cleared at each boot.                                                        |
 +-------------------------------+---------------------------------+----------------------------------------------------------------------------------------+
-| SECRET_KEY                    |  ``tralala``                    | The secret key used to encrypt the cookies. `ihatemoney conf-example ihatemoney.cfg`   |
-|                               |                                 | sets it to something random, which is good.                                            |
+| SECRET_KEY                    |  ``tralala``                    | The secret key used to encrypt the cookies.                                            |
+|                               |                                 |                                                                                        |
+|                               |                                 | **Production value:** `ihatemoney conf-example ihatemoney.cfg` sets it to something    |
+|                               |                                 | random, which is good.                                                                 |
 +-------------------------------+---------------------------------+----------------------------------------------------------------------------------------+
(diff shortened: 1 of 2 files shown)


ihatemoney/run.py
```diff
@@ -28,6 +28,12 @@ def _pre_alembic_db():
         alembic_setup = db.engine.dialect.has_table(con, 'alembic_version')
         return tables_exist and not alembic_setup
 
+    sqlalchemy_url = app.config.get('SQLALCHEMY_DATABASE_URI')
+    if sqlalchemy_url.startswith('sqlite:////tmp'):
+        warnings.warn(
+            'Database is stored in /tmp and might be lost at next reboot.'
```
almet: nit: "The database is currently stored [...]"

docs/installation.rst
```diff
@@ -157,35 +157,57 @@ properly.
...
+|                               |                                 |                                                                                        |
+|                               |                                 | **Production value:** Up to you, according to your needs.                              |
 +-------------------------------+---------------------------------+----------------------------------------------------------------------------------------+
 | APPLICATION_ROOT              |  ``""``                         | If empty, ihatemoney will be served at domain root (e.g: *http://domain.tld*), if set  |
 |                               |                                 | to ``"foo"``, it will be served from a "folder" (e.g: *http://domain.tld/foo*)         |
+|                               |                                 |                                                                                        |
+|                               |                                 | **Production value:** Up to you, according to your needs.                              |
```
almet: I don't think this adds much value, we could just remove the "production value" for this setting.

docs/installation.rst
```diff
@@ -157,35 +157,57 @@ properly.
...
+|                               |                                 | **Production value:** Usually, you will want to set it to ``False`` for a personal     |
+|                               |                                 |  instance.                                                                             |
 +-------------------------------+---------------------------------+----------------------------------------------------------------------------------------+
 | ACTIVATE_ADMIN_DASHBOARD      |  ``False``                      | If set to `True`, the dashboard will become accessible entering the admin password     |
 |                               |                                 | If set to `True`, a non empty ADMIN_PASSWORD needs to be set                           |
+|                               |                                 |                                                                                        |
+|                               |                                 | **Production value:** Up to you, according to your needs.                              |
```
almet: nit: this is probably not useful :)

docs/installation.rst
```diff
@@ -157,35 +157,57 @@ properly.
...
-| ALLOW_PUBLIC_PROJECT_CREATION |  ``True``                       | If set to `True`, everyone can create a project without entering the admin password    |
-|                               |                                 | If set to `False`, the password needs to be entered (and as such, defined in the       |
+| ALLOW_PUBLIC_PROJECT_CREATION |  ``True``                       | If set to ``True``, everyone can create a project without entering the admin password  |
+|                               |                                 | If set to ``False``, the password needs to be entered (and as such, defined in the     |
 |                               |                                 | settings).                                                                             |
+|                               |                                 |                                                                                        |
+|                               |                                 | **Production value:** Usually, you will want to set it to ``False`` for a personal     |
```
almet: We have different profile for the admins: some of them will have a personal instance, some of them will offer a service for others. I don't think this should be a recommended setting for production value. (we should provide no hint for this)

almet:
Thanks for the changes! After reading the whole thing again, I like how this is separated with the examples described in the "production value" section. I've commented some lines where I think there is no real value in adding some text (for instance, when we say "it's up to you"). I believe we should just point out what is the recommended value and keep it simple.

Thanks again for taking the time to propose these changes. I'll wait for an update to them before reviewing again and hopefully merging this :)

Phyks:
Thanks for the review! I should have updated my PR accordingly :)

almet:
Great, let's merge this then :-)

Phyks:
Thanks!

PR #285, opened 2017-11-11
Fix some typos using codespell

aavenel:
Fix some typos with the help of codespell (`pip install codespell`)


almet:
Thanks, nice tool! Should we include it in our Travis jobs?

PR #286, opened 2017-12-17
Use hashed passwords for projects

### LINKED ISSUE #232: Use unique URIs to share projects

0livd:
- Remove all occurences of clear text project passwords.
- Migrate the database to hash the previously stored passwords.
Closes #232

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,6 +21,7 @@ Changed
 - Use token based auth to reset passwords (#269)
 - Better install doc (#275)
 - Use token based auth in invitation e-mails (#280)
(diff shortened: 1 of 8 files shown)


almet:
And, it's now merged! Thanks for the work :-)

PR #287, opened 2017-12-21
Add missing properties in API docs

JocelynDelalande:
Because this is reality :-)


almet:
Thanks :) Nice to see that this API evolves easily without us noticing ;)

PR #291, opened 2017-12-23
Remove un-needed sidebar block

JocelynDelalande:
{% block sidebar %} is used by sidebar_table_layout.html and its children, not by
layout.html nor its direct children.

This is dead code removal.


almet:
I SEE DEAD CODE.

PR #292, opened 2017-12-23
Position flash messages with CSS rather than JS

JocelynDelalande:
Because I'm in the mood of killing some JS today :-)

### DIFF
--- a/ihatemoney/static/css/main.css
+++ b/ihatemoney/static/css/main.css
@@ -245,8 +245,13 @@ tr.payer_line .balance-name{
     margin-top: 10px;
 }
 
-.flash {
+.messages {
+    display: flex;
+    justify-content: center;
     position: absolute;
+    top: 0.6rem;
+    width: 100%;
+    pointer-events: none;
 }
 
 .light {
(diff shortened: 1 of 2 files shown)


JocelynDelalande:
Besides, I'm not satisfied with where the flash messages are positioned in ihatemoney, but that is another story :-).

JocelynDelalande:
Thanks for your review :-)

> There is a small regression though. When I add a new user, I have to wait for the flash message to disappear to be able to click the Add new bill button.

Well seen, I missed this one. I believe this is now fixed :-).

I used a CSS property that is [unsuported on IE <= 10](https://caniuse.com/#feat=pointer-events). IMHO, this is OK (IIRC, IE is auto-updating by default since some time. Result is IE10 have ridiculous usage share).

So, IMHO, no need to use the polyfill (or do somehow else).

WDYT ?

0livd:
It's now working fine on both Firefox and Chromium. I didn't bother testing on IE but yeah I guess dropping IE 10 support won't hurt much :)

almet:
I agree: we don't really care about internet explorer versions before 10. Thanks for the work on this one, it works perfectly locally, let's merge!

PR #293, opened 2017-12-25
Silent stderr output during app init for some commands

JocelynDelalande:
Avoid confusing the user for commands outputing text.

That is quite of a hack, but at least it is simple.

I tried several other approaches, but they proved overly complicated (involving big refactoring, inheritance madness, or removing commands, or even having to deal with metaclasses…).

### DIFF
--- a/ihatemoney/manage.py
+++ b/ihatemoney/manage.py
@@ -3,6 +3,7 @@
 import os
 import pkgutil
 import random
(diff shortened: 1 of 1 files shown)


almet:
Yeah, it's kind of hackish, I agree. But that seems legitimate. Let's do it like that.

Issue #298, opened 2017-12-27
Use a different dependency for the REST API.


almet:
@JocelynDelalande pointed out recently a bug that is related to a package (flask-rest) I published back in the day (6 years ago) to do the REST API.

I've published a change on the package to fix the issue, but it's interesting to note that this package is not really maintained, and doesn't have tests. I wrote this at a point where I wasn't aware of the best practices for software development, I fear.

Since that, the ecosystem had changed quite a lot and I think we should consider changing the dependency to something else (https://flask-restful.readthedocs.io/ ?)

What do you think?

JocelynDelalande:
Seems legit :)

PR #301, opened 2017-12-27
Allow to disable/enable member via API

JocelynDelalande:
*EDIT: since then, this is working*

This is not working yet, the `activated` state pushed via API seems not to be saved to DB. But I cannot figure out why.

Any ideas ? Help more than welcome.

I wrote a test, it shows how I would expect it to work.

### DIFF
--- a/ihatemoney/api.py
+++ b/ihatemoney/api.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 from flask import Blueprint, request
 from flask_rest import RESTResource, need_auth
+from wtforms.fields.core import BooleanField
 
 from ihatemoney.models import db, Project, Person, Bill
 from ihatemoney.forms import (ProjectForm, EditProjectForm, MemberForm,
@@ -58,6 +59,18 @@ class ProjectHandler(object):
         return 400, form.errors
 
 
+class APIMemberForm(MemberForm):
+    """ Member is not disablable via a Form.
+
+    But we want Member.enabled to be togglable via the API.
+    """
+    activated = BooleanField(false_values=('false', '', 'False'))
+
+    def save(self, project, person):
+        person.activated = self.activated.data
+        return super(APIMemberForm, self).save(project, person)
+
+
 class MemberHandler(object):
 
     def get(self, project, member_id):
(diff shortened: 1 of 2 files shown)


almet:
Hey. I managed to make it work with the following changes:

```diff
+++ b/ihatemoney/api.py
@@ -64,7 +64,7 @@ class APIMemberForm(MemberForm):
 
     But we want Member.enabled to be togglable via the API.
     """
-    activated = BooleanField()
+    activated = BooleanField(false_values=('false', '', 'False'))
```

It seems for whatever reason that the `False` value is not converted properly. I'm unsure if it's during the sending by the test client, or if it's when received by the server's view.

almet:
I've digged a bit more on this: it seems that the data we're receiving is not properly formatted: we receive direct strings, rather than JSON-encoded strings.

So, I've read some documentation about how the Flask test-client works, and [it seems](http://werkzeug.pocoo.org/docs/0.13/test/#werkzeug.test.EnvironBuilder) that by default it only sends form-encoded data, which explains why we were not getting the proper values.

The patch for this will probably be bigger than we think, because we will need to change all the calls to the API to specify the `Content-Type` and encode the data to JSON.

```python
res = self.app.post('/api/call, 
                    data=json.dumps(data),
                    content_type='application/json')
```

almet:
Some more thoughts: I believe this API is starting to show its current limits. Instead of accepting JSON as an input, it seems to only support form-encoded data, which doesn't seems right (and probably never was the intent).

I think we should discuss this in the issue about changing how the API is currently generated (#298).

JocelynDelalande:
> Hey. I managed to make it work with the following changes:

Hurray ! thanks a lot for this help ! My eyes slipped over this part of the doc without stopping while debugging.

> Some more thoughts: I believe this API is starting to show its current limits. Instead of accepting JSON as an input, it seems to only support form-encoded data, which doesn't seems right (and probably never was the intent).

I globally agree (API lib change required), but I discovered (by testing) that the API actually accepts JSON request body.

JocelynDelalande:
Ready for merge :-). I let someone else giving a look and merging.

PR #303, opened 2018-01-04
Enable basic auth passthrough for API

donkers:
Added the couple of apache configuration lines necessary to get the basic authentication working for the API.


ihatemoney/conf-templates/apache-vhost.conf.j2
```diff
@@ -4,13 +4,15 @@
...
     WSGIScriptAlias / {{ pkg_path }}/wsgi.py
+    WSGIPassAuthorization On
 
     ErrorLog /var/log/apache2/ihatemoney.example.com_error.log
     CustomLog /var/log/apache2/ihatemoney.example.com_access.log combined
 
     <Directory {{ pkg_path }}>
        WSGIProcessGroup ihatemoney
        WSGIApplicationGroup %{GLOBAL}
+       AuthType Basic
```
JocelynDelalande: With this line, no risk that it prompts user for auth even on non-API page ? (just asking).

donkers:
Hey @0livd you are totally correct and I after doing some testing here I have pushed another commit to the pull request.

PR #308, opened 2018-01-05
Include all .j2 files in the packaged version.

### LINKED ISSUE #305: Templates are not included in the packaged version
When summoning the `ihatemoney generate-config gunicorn.conf.py` command, I'm prompted with this error:

```python
IOError: [Errno 2] No such file or directory: '/home/www/ihatemoney.org/venv/local/lib/python2.7/site-packages/ihatemoney/conf-templates/gunicorn.conf.py.j2'
```

The same happens when I try to run the script with `supervisord.conf, for instance`.

almet:
I've also renamed the templates to *.j2 in order to make things clearer
to others. Having extensions with the name of the locale doesn't seem to
be a good practice, and would need us to add the locales in the MANIFEST
file each time we add one.

Fix #305


JocelynDelalande:
Also, I think we should mention that in changelog.

almet:
I've updated with an entry in the CHANGELOG.

CHANGELOG.rst
```diff
@@ -6,7 +6,10 @@ This document describes changes between each past release.
 2.1 (unreleased)
 ----------------
 
-- Nothing changed yet.
+Fixed
+=====
+
+- Include all .j2 files in the packaged version (#308)
```
JocelynDelalande: IMHO, the changelog audience is users/admins, not devs.
it should contain entries on what actually change for users/admin.

Thus, I'd suggest a wording like:

> Fix "IOERrorr" crash running `ihatemoney generate-config`

(sorry if it seems like nitpicking)

almet: That's an interesting point. I'll update the branch, thanks for the comments, they're greatly appreciated, don't worry :)

PR #309, opened 2018-01-05
Fix the supervisord template.

### LINKED ISSUE #306: Supervisord generated config file contains a `None` value
Here is the output of the `./venv/bin/ihatemoney generate-config supervisord.conf` command:

```ini
[program:ihatemoney]
command=None/bin/gunicorn -c /etc/ihatemoney/gunicorn.conf.py ihatemoney.wsgi:application
user=ihatemoney
autostart=true
autorestart=true
redirect_stderr=true
```

I find the `None` highly suspicious. It's supposed to be the `{{ venv_path }}`.

almet:
The script was relying on the presence of an environment variable, which
is only set when the virtualenv is activated. But a virtualenv does not
have to be activated to work (it's possible to call the python command
directly).

This fixes it by relying on `sys.executable` which should be correct at
all times.

Fixes #306


ihatemoney/manage.py
```diff
@@ -49,9 +49,11 @@ def run(self, config_file):
             os.path.join('conf-templates/', config_file) + '.j2'
         ).decode('utf-8')
 
+        bin_path = os.path.join(os.path.dirname(sys.executable))
```
JocelynDelalande: Why `os.path.join` ?

almet: Good catch. I added this while playing around, but we can get rid of it now!

almet:
The branch is now updated.

JocelynDelalande:
> The branch is now updated.

Thanks :)

I think there is still the apache conf to fix (see my previous comment).

almet:
Ah I missed this one. I'll let it for the next days or so as I won't be able to spend much time on this. Feel free to fix it in the meantime, otherwise I'll get back to it whenever I can :)

almet:
I've updated the pull request, and created an Jinja2 environment that enforces the use of defined template variables.

ihatemoney/conf-templates/apache-vhost.conf.j2
```diff
@@ -1,8 +1,7 @@
 <VirtualHost *:80>
     ServerAdmin admin@example.com      # CUSTOMIZE
     ServerName ihatemoney.example.com  # CUSTOMIZE
-
-    WSGIDaemonProcess ihatemoney user=www-data group=www-data threads=5 python-path={{ pkg_path }} {% if venv_path %}python-home={{ venv_path }}{% endif %}
+    WSGIDaemonProcess ihatemoney user=www-data group=www-data threads=5 python-path={{ pkg_path }} {% if bin_path %}python-home={{ bin_path }}{% endif %}
```
JocelynDelalande: I think this simple « drop-in » replacement might not be good, as `venv_path` does not point the same thing as `bin_path` (to my understanding).

almet: Yeah, I'm not sure what this should point to, actually!

We probably should add another set of variables when à venv is in use.

JocelynDelalande: Bump :)

I think we break some use case here (`generate-config apache` within venv), so IMHO, better fixing that before merging :)

almet: yeah, sorry, I lack some time these days with the opening of the brewery. I'll give it a look when I can, but don't hesitate to take over otherwise!

JocelynDelalande: @almet I did :-). I PR-ed your PR (crazy !), see #359 I let you the care to merge it and rebase.

*Hint: the ChangeLog might have to be fixed since 2.1 has been released since.*

Not that this should work but is untested on a real apache (I do not use apache), . @0livd, as the official reference ihm-with-apache2 user if you have some time to test the apache2 setup outputed by `ihatemoney ihatemoney generate-config apache-vhost.conf`, feel free :-).

JocelynDelalande:
@0livd leaving open a few days to let you a chance to review/test apache conf, as explained in https://github.com/spiral-project/ihatemoney/pull/309#discussion_r202818021

PR #315, opened 2018-01-09
API: Migrate from flask-rest to flask-restful

### LINKED ISSUE #298: Use a different dependency for the REST API.

0livd:
The flask-rest custom json encoder is still needed and thus was added to ihatemoney's utils.

Closes #298

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,11 @@ This document describes changes between each past release.
 2.1 (unreleased)
 ----------------
 
+Changed
+=======
+
+- Use flask-restful instead of deprecated flask-rest for the REST API (#315)
+
 Fixed
 =====
 
--- a/ihatemoney/api.py
+++ b/ihatemoney/api.py
@@ -1,62 +1,68 @@
 # -*- coding: utf-8 -*-
 from flask import Blueprint, request
-from flask_rest import RESTResource, need_auth
+from flask_restful import Resource, Api, abort
 from wtforms.fields.core import BooleanField
 
 from ihatemoney.models import db, Project, Person, Bill
 from ihatemoney.forms import (ProjectForm, EditProjectForm, MemberForm,
                               get_billform_for)
 from werkzeug.security import check_password_hash
+from functools import wraps
 
 
 api = Blueprint("api", __name__, url_prefix="/api")
+restful_api = Api(api)
 
 
-def check_project(*args, **kwargs):
+def need_auth(f):
     """Check the request for basic authentication for a given project.
 
-    Return the project if the authorization is good, False otherwise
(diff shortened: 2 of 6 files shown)


CHANGELOG.rst
```diff
@@ -6,6 +6,11 @@ This document describes changes between each past release.
 2.1 (unreleased)
 ----------------
 
+Changed
+=======
+
+- Use flask-restful for the REST API (#315)
```
almet: I think it would make sense to explicit a bit why we're doing this move?

ihatemoney/api.py
```diff
@@ -1,62 +1,68 @@
...
-class ProjectHandler(object):
-
-    def add(self):
+    @wraps(f)
+    def wrapper(*args, **kwargs):
+        auth = request.authorization
+        project_id = kwargs.get("project_id")
+
+        if auth and project_id and auth.username == project_id:
+            project = Project.query.get(project_id)
```
almet: that's probably me being silly, but in the previous version of this code, I was using `auth.username` here rather than the direct input from the user, because I tend to trust less the input from the user.

If for whatever reason the `==` part evaluates to `True`, it incidentally opens a breach to do queries with other unintended `ids`. I cannot think of an use for this right now, but I thought I would share my thoughts on this. What's your take on this?

ihatemoney/api.py
```diff
@@ -71,98 +77,92 @@ def save(self, project, person):
         return super(APIMemberForm, self).save(project, person)
 
 
-class MemberHandler(object):
+class MembersHandler(Resource):
```
almet: Ah, inheritance rather than composition! I liked how this old design looked :-)

ihatemoney/api.py
0livd: Yeah I see your point but ``auth.username`` is also a direct user input, isn't it ?

0livd:
Hey, I reworded the changelog line and reverted to ``auth.username`` in the authentication wrapper.

For the custom JSON encoder testing, it is kind of handled by all the API tests that return JSON but maybe you were thinking about some kind of unit test for the ``IhmJSONEncoder`` class ?

almet:
Nope, it's all fine by now :) Thanks! Let's ship it!

PR #316, opened 2018-01-23
Make sidebar scrollable

rscoates:
Make sidebar scrollable when there is an overflow.


rscoates:
Done!

JocelynDelalande:
This PR introduced a regression : the scrollbar was always present, even if there was nothing to scroll, thus, I allowed myself to fix this regression directly on master.

See issue:
image

rscoates:
Thanks, that’s much cleaner.

PR #321, opened 2018-01-30
Fix some anti patterns in docker deployment

0livd:
- Use exec to run gunicorn and avoid creating a new process.
- Add the possibility to pass any additional parameters to
gunicorn.
- Use only one gunicorn worker by default as the usual way to
scale the app in production would be to use the scale command
of the cluster scheduler. Additional workers could still be
added by passing the "-w" gunicorn parameter to docker run.


0livd:
@bnjbvr If you have some time for a review :)

PR #324, opened 2018-02-03
Add an API endpoint for statistics

JocelynDelalande:


JocelynDelalande:
> Tested locally and LGTM, the failing test is only a linting one.

… But is worth fixing :). 

So this is fixed, I added a ChangeLog entry, resolved  merging conflicts and most of all, you approoved the PR. So, all green, merging !

Issue #296, opened 2017-12-27
Design choices for Single-page app


JocelynDelalande:
Thinking about #149 and in the need on some pretext to learn angular (already known angularjs, needed to upgrade), I started working on an Angular version. @Phyks is also thinking about vue.js.

At this stage, no one knows if one implementation or the other will reach something worth it.

Anyway… I wanted to share some thoughs / questions.

## A. What should be covered by rich web app ?

The value of rich web app is IMHO mostly for the bills listing view. The rest of the app (project creation, login form…) might  or might not keep the current classic python-backed non-ajax interface. 

The templates between the bill listing view and the front-door pages have little in common, so might not be too difficult to split between python+jinja templates and js+whatever templates.

IMHO: match, up to the implementer.

## B. Waiting for server response or not ?

One of the interests of single-page apps is to be very fast & reactive on user interactions. One way to be even faster is to update displayed data without waiting server response.

So, there are several approaches to do (example of adding a user) :

### 1. Synchronous

1. user fills name input and clicks *Add* button
2. send `POST` request  to server
3. receive `POST` response (valid or error)
4. update client-side data with new user (or not, if error)

### 2. Fully confident

1. user fills name input and clicks *Add* button
2. send `POST` request  to server
3. update client-side data with new user
3. receive `POST` response (valid or error)
  - if valid : consolidate client-side data (ex: server sent the `id`)
  - if invalid : undo client-side data modification

#### Pros

- minimal network and CPU usage
- very reactive UI

#### Cons

- What if some other user modifies made modifications meanwhile ?
- some data logic have to occur on client-side AND server-side
- some edge cases are tricky (ex: add a bill on a new member that has not yet been acknowledged by server)

### 3. Confident but loyal to server

1. user fills name input and clicks *Add* button
2. send `POST` request  to server
3. update client-side data with new user
3. receive `POST` response (valid or error)
4. either valid or invalid : refresh the whole project data from server

#### Pros

- very reactive UI
- data logic enforced from server
- receiving 3rd-party data updates

#### Cons

- "important" network (refetch) and cpu usage (redraw)
- some edge cases are tricky (ex: add a bill on a new member that has not yet been acknowledged by server), and the « dangerous time frame » is longer than with 2. (as there is no data consolidation)

JocelynDelalande:
> B. Waiting for server response or not ?

[...]

PR #327, opened 2018-02-19
Add bill.imported_on field

### LINKED ISSUE #325: Import date as well as bill date
Hi,

There is currently a single date field on bills, reflecting the date of the bill. 

It is usually helpful to have the date of addition as well. This is especially useful when adding long past bills so that the date reflects the real date of the bills, but to allow a way to sort and see last additions. 

What do you think about adding a new date field on the bill to store date of last update? I can write a PR for this if you are interested :)

Thanks,

Phyks:
Should fix #325


ihatemoney/templates/list_bills.html
```diff
@@ -98,10 +98,11 @@ <h3 class="modal-title">{{ _('Add a bill') }}</h3>
 
     {% if bills.count() > 0 %}
     <table id="bill_table" class="col table table-striped table-hover">
-        <thead><tr><th>{{ _("When?") }}</th><th>{{ _("Who paid?") }}</th><th>{{ _("For what?") }}</th><th>{{ _("For whom?") }}</th><th>{{ _("How much?") }}</th><th>{{ _("Actions") }}</th></tr></thead>
+        <thead><tr><th>{{ _("Added on") }}</th><th>{{ _("When?") }}</th><th>{{ _("Who paid?") }}</th><th>{{ _("For what?") }}</th><th>{{ _("For whom?") }}</th><th>{{ _("How much?") }}</th><th>{{ _("Actions") }}</th></tr></thead>
     <tbody>
     {% for bill in bills %}
     <tr owers="{{bill.owers|join(',','id')}}" payer="{{bill.payer.id}}">
+            <td>{{ bill.import_date }}</td>
```
almet: This looks good to me, except the name: should we name this `creation_date` instead?

Phyks: Probably :)

I was lacking a bit of imagination here.

ihatemoney/migrations/alembic.ini
```diff
@@ -7,6 +7,7 @@
 # set to 'true' to run the environment during
 # the 'revision' command, regardless of autogenerate
 # revision_environment = false
+script_location = .
```
Phyks: I had to put this to be able to use `alembic`. Not sure if this is coming from me not using it the right way or not.

Phyks:
Note that there should probably be a better UI counterpart (sorting by creation date for instance), but that should most likely be done as part of #296. I might have some time in the coming weeks to handle it.

ihatemoney/migrations/versions/afbf27e6ef20_add_bill_import_date_field.py
```diff
@@ -0,0 +1,26 @@
...
+revision = 'afbf27e6ef20'
+down_revision = 'b78f8a8bdb16'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('bill', sa.Column('import_date', sa.Date(), nullable=True))
```
Phyks: Maybe I should fill the new field to mimmick the `date` value for already existing bills?

Phyks:
Should have been updated according to comments and with tests fixed :)

ihatemoney/migrations/alembic.ini
JocelynDelalande: what are the commands you used for alembic ?

JocelynDelalande:
@Phyks thanks for the PR :-). I agree that keeping record of the creation date is a good thing and might be useful. 

However, I would like that we discuss some points before we potentially merge :

- sorting by creation date is opinionated and is not relevant for all use cases (I think it goes against most of my own use cases: eg: shared house : even if I log the September internet bill in December, I it still want to see it next to the Geptember gas bill within the listing)
- I hate money is damn simple and quite easy to use. Having two dates in the table view might introduce some rarely needed complexity in the interface.

To tell more on my last point, I used a lot of time (and with large groups) IHM in a realy dumb fashion : really minimal labels, and do care taken about putting the right date. Having a limited amount of information helps, and a single date helps, for that use case.

I don't think my remarks are incompatible with your own use case, but we have to discuss how (eg: configurable sorting).

Side note : if you had time to expose that information through the REST API, that would be lovely :)

ihatemoney/migrations/alembic.ini
Phyks: Not sure, but I think it was something like `alembic revision -m "add bill.import_date field"`, ran in the `migrations` dir.

Phyks:
> sorting by creation date is opinionated and is not relevant for all use cases (I think it goes against most of my own use cases: eg: shared house : even if I log the September internet bill in December, I it still want to see it next to the Geptember gas bill within the listing)

I have a rather similar use case, however I quite disagree on this one. I'd like to have a way to easily see newly added bills, without having to diff a long list and dig in the past. With the current setup my roommate might add a bill for september which will simply go unnoticed. I'll see my balance change, and figuring out why is a really difficult task.

I think I should emphasize this is a creation date, linked to the date of addition of the bill. The date field is still there and with the same purpose. Then, it is just a matter of sorting in the UI to order in the way you want.

I did not want to implement configurable sorting for now, as there is the ongoing discussion about #296 and I thought this might solve everything at once.

Concerning the REST API, what do you mean exactly? I think it is already exposed in the REST API (see tests), but I might have missed something.

ihatemoney/migrations/alembic.ini
JocelynDelalande: Ok. The command `./manage.py db`¹ (sub)command should work out-of-the-box without requiring patch ; could you confirm ?

If so… I think we should document it better :)

¹ or `ihatemoney db` if using an installed version.

JocelynDelalande:
> Concerning the REST API, what do you mean exactly? I think it is already exposed in the REST API (see tests), but I might have missed something.

That is me who missed something on this point, my bad :x (I got tricked by too fast reading and seeng the `del` lines as a way to ignore totaly the dates).

(answering other parts later, little time now)

ihatemoney/migrations/alembic.ini
Phyks: Indeed, it seems to work. I completely missed the doc about this command though :/

JocelynDelalande: I think there simply is no doc, that is why you missed it  :-)

almet:
I think what we have here should be merged. I will try to first issue a point release to fix everything that's broken, and then come back to this, merge and issue a new major release (since we have a database API change here).

Phyks:
Great! Let me know if you need anything from my part.

Best,

Phyks:
Any news on this one? Thanks!

almet:
Sorry for the time it took us to go over this, without any real reason. I have a little time now so I will try to look at the few pull requests that are still open.

PR #335, opened 2018-05-09
Add more human "for who?" description.

Glandos:
The limit for displaying "Everyone except …" instead of the ower's list is completely empirical.

I don't know how to update translations template, I haven't used `gettext` for a long time. But it should be run right after merging. Or you can tell me what to do, I will push in this branch.


almet:
Thanks! For the translations, there is a target in the makefile. I don't remember how that works (we should update the docs!), but updating the .po file and running the makefile target could be a good start :)

https://github.com/spiral-project/ihatemoney/blob/master/Makefile#L40

Glandos:
OK. I found the way to do it. It is clearly out of the scope of this request. I'll open a new issue.

ihatemoney/templates/list_bills.html
```diff
@@ -105,7 +105,13 @@ <h3 class="modal-title">{{ _('Add a bill') }}</h3>
             <td>{{ bill.date }}</td>
             <td>{{ bill.payer }}</td>
             <td>{{ bill.what }}</td>
-            <td>{{ bill.owers|join(', ', 'name') }} </td>
+            <td>{% if bill.owers|length == g.project.members|length -%}
```
JocelynDelalande: Does this check plays nice with disabled users ?

almet: @Glandos, I'm not sure what your :+1: answer means? Is this handled correctly? Thanks :)

Glandos: Ahah, I was sure it was misleading.
No, it means: nice catch. I didn't think about disabled users, so I have to test it. Or at least, it has to be tested.

almet: Communication on the internet? Misleading? I have no idea what you're talking about ;-)

Anyway thanks for the help on this. As usual these days (launching a business also has its downsides when it comes to spare time) I probably won't take it over, but if anyone wants, go for it!

Glandos:
Let's talk about deactivated persons. For now, this change cannot be merged.

The issue is that deactivated persons are still in the database, and still have bills. So you can have:

1. Bill 1 payed by A for A,B,C_deactivated
2. Bill 2 payed by A for A,B

Since C_deactivated is not activated anymore, it has two states: either its balance is zero or not. In the first case, it **only disappears from the sidebar** and not from the bill list.

Maybe we should add a logic, somewhere, that totally removes a person from the database, including bill relationships when its balance reaches zero. In this case, the template doesn't have to mess with code like `balance[member.id]|round(2) != 0`.

Or, we can keep it like this and only show deactivated persons in bills if they belong to them. Otherwise, it should act as if it was completely deleted.

almet:
Let's get back to the intent here. Can't we just add a method in the model to know the number of active persons?

JocelynDelalande:
> Let's get back to the intent here. Can't we just add a method in the model to know the number of active persons?

+1, but that does not solve the whole question.

So, let's browse the different cases.

Considering we have users A, B, C and D. and D is disabled, the others being active. I am going to suggest arbitrary sentences so that we can discuss them.

- Bill involving everyone from A to I -> "everyone (+ old user)" (with a small icon with popover explaining on touch/hover who is that old user). In a case with more than 1 disabled user, we should pluralize
- Bill involving everyone but D -> "everyone "
- Bill involving A, B -> "A and B"
- Bill involving everyone but D and E -> "everyone but E"

We could discuss that suggestions, and add other cases maybe, until we agree on something sensible :-)

Glandos:
@almet That's a good start indeed. But I also agree with @JocelynDelalande that it just helps without solving the whole question.

Now regarding the 4 propositions, I really like the idea of mentionning old users in a popover with a simple mention. I should also add that a visual style could help, like a light gray text color, or italic, or both. And then maybe those users could be mentionned in the sidebar, at the very bottom, with the same style. And the popover could also have the full list of users, in case of the text is too long, even with this modification merged. But that's for another request.

I agree with the last 3 options: disabled users are totally invisible in the summary of bills that don't mention them.

almet:
I pretty much like these scenarios. I'm thinking at least of another one:

If we have a list of 5 of more persons, we probably should just state the number of users. What this limit number is is suggest to discussions (5 is just a number I took out of my hat).

almet:
So, to come back to the pull request here: I think we should at least introduce a concept of method in the model, which tells us how many (active) members we have active.

Once we have this we could merge it as is here, and open another request to have it fixed better, with the phrasing and all. What do you think?

almet:
Hi,

Sorry for the time it took us to go trough this. I'm willing to merge / review this, but I think the changes are lost in a lot of other changes here. I think it might be due to the fact you rebases / merged with another branch ? Not sure exactly what happened, but I can't access a readable diff with only the changes you're proposing.

Can you have a look at this, and I'll review it after :-) Thanks!

almet:
@Glandos did you got my last message? Is there a way to go ahead on this? Thanks!

Glandos:
Thanks for the poke. Yes, It's on my TODO list. I think I need to reopen a new request for that, without merged commits. I don't know when it will be done though :/

almet:
Okay, no need to hurry. Could you point out the changes you did that might motivate someone to do it instead?

almet:
(Closing because I created a clean pull request with your changes instead)

Glandos:
Thanks for that! And sorry for the unresponsiveness…

almet:
No need for apologies, really. Thanks for everything you've done so far for the project, it's appreciated!

Issue #340, opened 2018-06-24
pip 10 breaks setup.py


jhass:
The `req` module is gone. My current workaround is to `pip install pip==9.03` in the virtualenv.

JocelynDelalande:
Thanks @jhass this is pretty bad. 

https://github.com/capless/warrant/pull/101/files for a possible fix. 

Anyone to implement and test ?

JocelynDelalande:
note that this is why current travis CI is broken on master.

JocelynDelalande:
Fixed via #341

PR #341, opened 2018-06-26
fix install with pip10

mduret:
hello,

I propose a simple fix for this issue : https://github.com/spiral-project/ihatemoney/issues/340

Have a good day

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,16 +2,24 @@
 import codecs
 import os
 from setuptools import setup, find_packages
-try:
-    from pip.req import parse_requirements
-    from pip.download import PipSession
-except ImportError:
-    print('Cannot find pip.')
-    raise
 
-# Get requirements from the requirements.txt file.
-pip_requirements = parse_requirements("requirements.txt", session=PipSession())
-install_requires = [str(ir.req) for ir in pip_requirements]
+import pip
+if pip.__version__ < "10":
+    try:
+        from pip.req import parse_requirements
+        from pip.download import PipSession
+    except ImportError:
+        print('Cannot find pip.')
+        raise
+    # Get requirements from the requirements.txt file.
+    pip_requirements = parse_requirements("requirements.txt", session=PipSession())
+    install_requires = [str(ir.req) for ir in pip_requirements]
+else:
+    def parse_requirements(filename):
+        """ load requirements from a pip requirements file """
+        lineiter = (line.strip() for line in open(filename))
+        return [line for line in lineiter if line and not line.startswith("#")]
+    install_requires = parse_requirements('requirements.txt')
 
 here = os.path.abspath(os.path.dirname(__file__))
 
--- a/setup.py
+++ b/setup.py
@@ -3,24 +3,6 @@ import codecs
 import os
 from setuptools import setup, find_packages
 
-import pip
-if pip.__version__ < "10":
-    try:
-        from pip.req import parse_requirements
-        from pip.download import PipSession
-    except ImportError:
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -2,16 +2,24 @@
...
-# Get requirements from the requirements.txt file.
-pip_requirements = parse_requirements("requirements.txt", session=PipSession())
-install_requires = [str(ir.req) for ir in pip_requirements]
+
+import pip
+if pip.__version__ < "10":
+    try:
+        from pip.req import parse_requirements
+        from pip.download import PipSession
+    except ImportError:
```
almet: I think the try/except for the import should be just next to "import pip" rather than here.

setup.py
```diff
@@ -2,16 +2,24 @@
...
+        print('Cannot find pip.')
+        raise
+    # Get requirements from the requirements.txt file.
+    pip_requirements = parse_requirements("requirements.txt", session=PipSession())
+    install_requires = [str(ir.req) for ir in pip_requirements]
+else:
+    def parse_requirements(filename):
+        """ load requirements from a pip requirements file """
+        lineiter = (line.strip() for line in open(filename))
+        return [line for line in lineiter if line and not line.startswith("#")]
```
almet: If we have this utility, I think we should use it in all occasions, and throw away the use of pip's version. That's less differences and less trouble maintaining two versions.

almet: Could we have a bunch of tests for this new function?

setup.py
```diff
@@ -2,16 +2,24 @@
...
+    except ImportError:
+        print('Cannot find pip.')
+        raise
+    # Get requirements from the requirements.txt file.
+    pip_requirements = parse_requirements("requirements.txt", session=PipSession())
+    install_requires = [str(ir.req) for ir in pip_requirements]
+else:
+    def parse_requirements(filename):
+        """ load requirements from a pip requirements file """
+        lineiter = (line.strip() for line in open(filename))
```
almet: We open the file but never close it. (hint: it's handled automatically when using a `with` statement)

mduret:
Thanks for the tricks, I gonna propose a more elegant fix in few days.

mduret:
I propose a more simple way : it works with pip 9 and pip 10.
I hope it can help.

Have a good one,

almet:
Thanks! That seems to be a simpler way to handle this. Could you provide a few tests for the created function? (So we're sure it does what it should?). Thanks!

mduret:
Ok, thx I going to make this test.

JocelynDelalande:
I allowed myself to rename as your fix is broader than just docker.

mduret:
Hi,

I spend some time on this issue but I don't know where to put this test.

I am not an expert but `setup.py` is not part of the package so I do not know how to test it in the package.

Usualy, I put `-e .` in `requirements.txt` and I put requirements in the `setup` attribute `install_requires` of the `setup.py` file.

If you have an idea...

Thanks and have a good day.

JocelynDelalande:
> I am not an expert but setup.py is not part of the package so I do not know how to test it in the package.

Good point. I searched/tested things a bit myself and cannot find a way to do that (test setup.py functions).

py.test seem to fail at parsing it (`py.test setup.py` will crash). 

So unless someone comes quickly with a solution, and this bug (#340) maintains ihm in a quite broken state, I'd suggest merging as-is.

almet:
Yes, after some look around, I don't see any obvious way to do it as well. Let's merge! Thanks for your work and patience here :)

PR #357, opened 2018-07-07
Document MySQL setup

JocelynDelalande:
PyMySQL is more difficult to install since its version 0.9 since it now depends
on *cryptography* lib, which in turns depends on OpenSSL and Python dev files.

See https://github.com/PyMySQL/PyMySQL/issues/697

side-note : I am pretty sure it breaks the docker setup, but I do not use docker, anyone to check and fix ? @0livd ?

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -63,6 +63,22 @@ Once installed, you can start a test server::
 
 And point your browser at `http://localhost:5000 <http://localhost:5000>`_.
 
+Configure database with MySQL/MariaDB (optional)
+================================================
+
+Only required if you prefer MySQL/MariaDB over SQLite.
+
+1. Install PyMySQL dependencies. On Debian or Ubuntu, that would be::
+
+    apt install python3-dev libssl-dev
+
+2. Install PyMySQL (within your virtualenv)::
+
+    pip install 'PyMySQL>=0.9,<0.10'
+
+3. Create an empty database and a database user
+4. Configure :ref:`SQLALCHEMY_DATABASE_URI <configuration>` accordingly
+
 Deploy it
 =========
 
@@ -159,6 +175,8 @@ For example, use the following command to add more gunicorn workers::
 
     docker run -d -p 8000:8000 ihatemoney -w 3
 
+.. _configuration:
+
 Configuration
 =============
 
@@ -187,7 +205,7 @@ Production values are recommended values for use in production.
 |                               |                                 | **Production value:** `ihatemoney conf-example ihatemoney.cfg` sets it to something    |
 |                               |                                 | random, which is good.                                                                 |
 +-------------------------------+---------------------------------+----------------------------------------------------------------------------------------+
-| MAIL_DEFAULT_SENDER           | ``("Budget manager",            | A python tuple describing the name and email address to use when sending                |
+| MAIL_DEFAULT_SENDER           | ``("Budget manager",            | A python tuple describing the name and email address to use when sending               |
 |                               | "budget@notmyidea.org")``       | emails.                                                                                |
 |                               |                                 |                                                                                        |
 |                               |                                 |  **Production value:** Any tuple you want.                                             |
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,15 @@ This document describes changes between each past release.
 2.1.1 (unreleased)
 ------------------
 
+Fixed
+=====
+
 - Regenerate translations (#338)
+
+Added
+=====
+
+- Document MySQL setup (#357)
 
 
 2.1 (2018-02-16)


docs/installation.rst
```diff
@@ -63,6 +63,22 @@ Once installed, you can start a test server::
 
 And point your browser at `http://localhost:5000 <http://localhost:5000>`_.
 
+Configure database with MySQL/MariaDB (optional)
```
almet: Should this be "recommended" here rather than optional? SQLite is mainly intended for developers / very low usage setups.

docs/installation.rst
```diff
@@ -63,6 +63,22 @@ Once installed, you can start a test server::
...
+
+1. Install PyMySQL dependencies. On Debian or Ubuntu, that would be::
+
+    apt install python3-dev libssl-dev
+
+2. Install PyMySQL (within your virtualenv)::
+
+    pip install 'PyMySQL>=0.9,<0.10'
+
+3. Create an empty database and a database user
```
almet: That might be a good idea to give instructions here on how to do so, so the instructions are completely noob-proof (And they will help me, too ;))

docs/installation.rst
Glandos: What do you mean by "very low"?
I often see that for SQLite, but it is a very capable SQL engine that have good performance even with very huge database. I even run a test for a 8 GB database on a low-end hardware (Intel Atom + 2"5 HDD) and it performs at a subsecond.
The only drawback of SQLite is that it's not multi-user.
Backup are even easier, except if you already handle backups for your prefered database engine.

almet: I'm thinking mainly about scenarios where multiple writes can occur. That depends of course on your setup. As an example, the [default gunicorn template](https://github.com/spiral-project/ihatemoney/blob/master/ihatemoney/conf-templates/gunicorn.conf.py.j2#L4) uses 3 workers, and when these workers are used in parallel (e.g when 3 requests are held at the same time), multiple writes can occur.

It's still not clear to me how well SQLite handles multiple writes, but from what I understand, it degrades performance. Multiple reads do not cause any performance problems apparently.

One way to know more about this would be to write load tests and run them in parallel. I used to do that when I was working at mozilla, and we were using a tool named loads, which makes it possible to write tests scenarios in python, and run them on a huge infrastructure, to test if the service is handling the load.

Glandos: It seems reasonable. But with pure intuition (this is not a good metric), I'd say that with less than 10 users / projects, the performance with SQLite are really good enough.

Do you have some stats about the main instance?

almet: I could probably gather some. Which kind of information would be useful?

Glandos: For a start, let's be simple:
- Database lengths: project count, member count
- Some access log statistics, e.g. with [GoAccess](https://goaccess.io/) to know how much requests per second/minute/day.

I think that GoAccess can also display the request duration, if that information is logged, but I only played with it once :)

almet: Here are some:

- Total number of projects: 1650
- Projects with more than 10 bills: 347
- Number of active projects (in the past 3 months) with more than 10 bills: 57
- Total number of hits per day: 597
- Total number of visitors per day: 49

I don't have any way to know the number of members, since the concept doesn't exist in the software (or maybe do you have an idea on how to do so?)

zorun: @Glandos you might be interested in the performance results there: 8e5ac8ae98824b8b359cc0f020b10acb29985dab (keep in mind that's it's a low-end CPU, but still with a SSD)

For my personal instance, I switched from SQLite to PostgreSQL when the number of bills in the main project reached about 400, because it was starting to feel really slow.  I was surprised that PostgreSQL didn't improve the page loading time as much as I hoped (it still improved, but I don't have numbers, sorry).  On the other hand, the CPU usage on the server decreased really dramatically!

I *think* (not 100% sure) that this was before 8e5ac8ae98824b8b359cc0f020b10acb29985dab.

Glandos: @almet So it sounds to me like small figures. Do you actually run it with SQLite? If you don't, did you make the switch because of performance issues?

@zorun Yes, that change was a very big win. But this was about read (`SELECT`) while @almet mentioned that SQLite might cause issues with write (`UPDATE` et al.). This seems a reasonable asumption, but I don't have any real figures.

And for now, I don't have the time to run benchmark for this :disappointed:

almet: It runs on Postgres currently, probably because I don't have a high opinion about using SQLite in prod ;)

docs/installation.rst
JocelynDelalande: Well… It would help me too :-p.

There are different ways to create a MySQL database, I do use [yunohost helpers](https://github.com/YunoHost-Apps/ihatemoney_ynh/blob/master/scripts/install#L35) for that. 

Some people may use phpmyadmin or some other frontend.

All that to say I won't write that myself and now, as I do not have anything ready to copy-paste.

So IMHO, we could merge without further details (better than nothing), and open a ticket to remember providing detailed instructions with MySQL Command line.

almet:
Alright! It's merged, thanks!

PR #358, opened 2018-07-11
Remove strict_slashes for /

Glandos:
I don't know why, but on my setup (nginx + uwsgi), the `strict_slashes` (default to `True`) was causing an infinite loop.
I think it could be safely removed for this route only.

### DIFF
--- a/ihatemoney/tests/tests.py
+++ b/ihatemoney/tests/tests.py
@@ -1434,6 +1434,12 @@ class APITestCase(IhatemoneyTestCase):
 
 class ServerTestCase(IhatemoneyTestCase):
 
+    def test_homepage(self):
+        # See https://github.com/spiral-project/ihatemoney/pull/358
+        self.app.config['APPLICATION_ROOT'] = '/'
+        req = self.client.get("/")
+        self.assertStatus(200, req)
+
     def test_unprefixed(self):
         self.app.config['APPLICATION_ROOT'] = '/'
         req = self.client.get("/foo/")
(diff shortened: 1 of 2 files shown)


JocelynDelalande:
@Glandos out of curiosity, what was your nginx conf ?

Glandos:
```nginx
        location / {
                include uwsgi_params;
                uwsgi_pass unix:/run/uwsgi/ihatemoney.socket;
        }

        location /static {
                gzip_static on;
        }
```

The `uwsgi_params` is the one provided in Debian.

And in my uwsgi configuration:
```ini
wsgi-file = /var/www/example.com/ihatemoney/ihatemoney/wsgi.py
callable = application
env = IHATEMONEY_SETTINGS_FILE_PATH=/var/www/example.com/ihatemoney/data/ihatemoney.cfg
venv = /var/www/example.com/ihatemoney/venv-system/
```

mgautierfr:
I've got the same problem simply running `python run.py` in 'ihatemoney` directory (all dependencies being installed).
`http://localhost:5000/` was always redirecting to `http://localhost:5000/`
Other urls were fine.
So no nginx conf where involved. 

This PR fixed the problem.

almet:
The fix seems okay. Can we add a small test case for it? Thanks!

Glandos:
Here you are. The test is failing if `strict_slashes` is not set. And of course, it succeeds if it is set to `False`.

JocelynDelalande:
@Glandos nice !

I took the time to fetch and test.

@Glandos last thing, could you add a meaningful ChangeLog entry and squash your commits ?

Thanks for your care and patience!

PS: thanks for the nginx conf sample to satisfy my curiosity :-).

almet:
fixed!

PR #366, opened 2018-08-05
Almet/negative weights

almet:
If anyone got time to review, that would be great!


ihatemoney/models.py
```diff
@@ -346,7 +346,10 @@ def pay_each(self):
         """Compute what each share has to pay"""
         if self.owers:
             # FIXME: SQL might dot that more efficiently
-            return self.amount / sum(i.weight for i in self.owers)
+            weights = sum(i.weight for i in self.owers)
+            if weights <= 0:
```
bnjbvr: If I understand correctly, this can happen if, before this commit is merged, somebody had negative weights. If so, wouldn't it make more sense to:

- keep this code as is
- migrate former negative weights to 1?

Then we have the general invariant that weights are non-negative.

bnjbvr:
(Forgot to say: bonus points if you could squash the two commits together, since the second one fixes one error introduced in the first one.)

JocelynDelalande:
> Do not allow negative weights on users -> Please add a ChangeLog entry :)

Bumpybump on that point + a rebase is required.

almet:
Hi ! I don't have time these days for this, I'm sorry to let it in this state. Don't hesitate to take it over if you want; and release as you think is best!

almet:
I've just updated this branch with the fixes. This is my first database migration, if anyone could review this and tell me if I made something wrong, let me know!

Issue #344, opened 2018-07-05
"languages" buttons not visible enough


Nartagnan:
It is perhaps a detail, but at first I didn't saw the site was available in French.

almet:
I believe this is a dupe of #35

almet:
Ah, apparently not. Let's keep the discussion going. I think that if you found it afterwards it's good enough. What do you propose ?

Nartagnan:
Flags is not good, but perhaps adding a label :
Language : EN, FR
in bold, and in MAJ for the languages perhaps ?

almet:
I'm not a big fan about adding yet another text here that will probably clutter a bit the UI.

What about having this in the settings instead? We could:

- detect the language automatically with the browser settings (a specific HTTP headear `Accept-Language` I think is sent)
- in the settings, make it possible to change the language?

That should get us covered for 80% of the users.

almet:
We've changed that in recent versions, it should be easier to find!

PR #368, opened 2018-08-05
Move the language selector in the options menu.

almet:
Related to #344.

Here is a capture of what it looks like
languages

Any idea if this is better than before? (In which case I could add the translation strings)


eMerzh:
i kind of like the language dropdown instead of language list as of now... but maybe setting it in is own drop down like 
Selector
could make it more visible... need to test with the available space though 🤔

almet:
Yeah, it's obviously a matter of space. If you want to try, don't hesitate :-)

PR #371, opened 2018-08-05
Simplify a JS function.

almet:
Quite nothing, but still :)

### DIFF
--- a/ihatemoney/static/js/ihatemoney.js
+++ b/ihatemoney/static/js/ihatemoney.js
@@ -1,18 +1,7 @@
- // Add scripts to select all or non of the checkboxes in the add_bill form
-function selectall()
-    {
-        var els = document.getElementsByName('payed_for');
-        for(var i =0;i<els.length;i++)
-        {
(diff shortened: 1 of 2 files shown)


ihatemoney/static/js/ihatemoney.js
```diff
@@ -1,18 +1,7 @@
...
-        {
-            els[i].checked=false;
-        }
-    }
-
+ // Utility to select all or none of the checkboxes in the add_bill form.
+function selectCheckboxes(value){
+  var els = document.getElementsByName('payed_for');
+  for(var i = 0; i < els.length; i++){
+    els[i].checked = value;
```
bnjbvr: if you can use es6: `for (let node of document.getElementsByName('payed_for')) { node.checked = value; }` for even more conciseness

almet: I'm not really sure which JS version / browsers we're targeting, so I'll let as is until we make a decision on this ;)

PR #373, opened 2018-08-05
Better error message for alredy taken project id.

### LINKED ISSUE #346: UX : error message for "project's name already taken"
> Project identifier: The project identifier is used to log in and for the URL of the project. We tried to generate an identifier for you but a project with this identifier already exists. Please create a new identifier that you will be able to remember

The error message was in red, as for a technical problem. Perhaps having a friendlier one as it is not critical.
Also, when you try something else, if it is also used, absolutely nothing change. (if the message do not change, I just though "I didn't click, I'll try again", but it was not that)
I suggest :
> Project identifier: "ALREADY TAKEN PROJECT NAME HERE" The project identifier is used to log in and for the URL of the project. We tried to generate an identifier for you but a project with this identifier already exists. Please create a new identifier that you will be able to remember

If the message is dynamic, it will be also easier to understand what it is about.

almet:
Fixes #346


ihatemoney/translations/fr/LC_MESSAGES/messages.po
```diff
@@ -33,16 +33,12 @@ msgstr "Identifiant du projet"
 msgid "Create the project"
 msgstr "Créer le projet"
 
+#, python-format
 msgid ""
-"The project identifier is used to log in and for the URL of the project. "
```
JocelynDelalande: english version totally removed by accident  ?

ihatemoney/messages.pot
```diff
@@ -13,11 +13,10 @@ msgstr ""
 msgid "Create the project"
 msgstr ""
 
+#, python-format
 msgid ""
-"The project identifier is used to log in and for the URL of the project. "
-"We tried to generate an identifier for you but a project with this "
-"identifier already exists. Please create a new identifier that you will "
-"be able to remember"
+"A project with this identifier (\"%(project)s\") already exists. Please "
```
JocelynDelalande: Change « create » -> « choose ».

Maybe, that is not the right place to mention « that you will be able to remember ». Let's be more straight in that error message place, WDYT ?

PR #381, opened 2018-08-05
Add a favicon.ico route and logo.

### LINKED ISSUE #255: HTTPS and Google Chrome
Some requests make the application not usable in HTTPS. Is there a way to fix it ?

These days, HTTPS for web applications is a must have !

almet:
Might fix #255


ihatemoney/run.py
```diff
@@ -2,7 +2,7 @@
 import os.path
 import warnings
 
-from flask import Flask, g, request, session
+from flask import Flask, g, request, session, url_for
```
0livd: Url for what ? :D

almet: Heh, it's coming from when I first tried to add the route in the run.py file (rather than in the web module)

PR #387, opened 2018-08-30
doc: Improve Makefile

marienfressinaud:
Add help target to the Makefile

It is often expected to have a `help` target in a Makefile. This one is
automatically generated from comments in the Makefile so it is easier to
maintain.

This commit only documents targets that seem the most important.

---

Add the server address and port on `make serve`

Developers should not have to read the documentation to find where to
point their browsers to access ihatemoney application.

---

Add .PHONY instructions to the Makefile

`.PHONY` instructs Make to not look, for instance, for an `install` file
before executing the corresponding target. If such a file would exist,
the command would not be performed at all. This is because Make is
initially intended to create files (i.e. targets) and consider there is
nothing to do if the file already exists.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,56 +7,71 @@ DOC_STAMP = $(VENV)/.doc_env_installed.stamp
 INSTALL_STAMP = $(VENV)/.install.stamp
 TEMPDIR := $(shell mktemp -d)
 
+.PHONY: all
 all: install ## Alias for install
+.PHONY: install
 install: virtualenv $(INSTALL_STAMP) ## Install dependencies
 $(INSTALL_STAMP):
 	$(VENV)/bin/pip install -U pip
 	$(VENV)/bin/pip install -r requirements.txt
 	touch $(INSTALL_STAMP)
 
+.PHONY: virtualenv
 virtualenv: $(PYTHON)
 $(PYTHON):
 	$(VIRTUALENV) $(VENV)
 
+.PHONY: install-dev
 install-dev: $(INSTALL_STAMP) $(DEV_STAMP) ## Install development dependencies
 $(DEV_STAMP): $(PYTHON) dev-requirements.txt
 	$(VENV)/bin/pip install -Ur dev-requirements.txt
 	touch $(DEV_STAMP)
 
+.PHONY: remove-install-stamp
 remove-install-stamp:
 	rm $(INSTALL_STAMP)
 
+.PHONY: update
 update: remove-install-stamp install ## Update the dependencies
 
+.PHONY: serve
 serve: install ## Run the ihatemoney server
 	@echo 'Running ihatemoney on http://localhost:5000'
 	$(PYTHON) -m ihatemoney.manage runserver
 
+.PHONY: test
 test: $(DEV_STAMP) ## Run the tests
 	$(VENV)/bin/tox
 
+.PHONY: release
 release: $(DEV_STAMP) ## Release a new version (see https://ihatemoney.readthedocs.io/en/latest/contributing.html#how-to-release)
 	$(VENV)/bin/fullrelease
 
+.PHONY: build-translations
 build-translations: ## Build the translations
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -18,41 +18,45 @@ virtualenv: $(PYTHON)
...
 	$(VENV)/bin/pybabel extract --strip-comments --omit-header --no-location --mapping-file ihatemoney/babel.cfg -o ihatemoney/messages.pot ihatemoney
 	$(VENV)/bin/pybabel update -i ihatemoney/messages.pot -d ihatemoney/translations/
 
-create-database-revision:
+create-database-revision: ## Create a new database revision
 	@read -p "Please enter a message describing this revision: " rev_message; \
 	$(PYTHON) -m ihatemoney.manage db migrate -d ihatemoney/migrations -m "$${rev_message}"
 
-build-requirements:
+build-requirements: ## Freeze the requirements
```
JocelynDelalande: « save currently installed packages to requirements.txt »

(« freeze » gives the false impression, that it will only do a pip freeze)

Makefile
```diff
@@ -18,41 +18,45 @@ virtualenv: $(PYTHON)
...
-release: $(DEV_STAMP)
+release: $(DEV_STAMP) ## Release a new version
 	$(VENV)/bin/fullrelease
 
-build-translations:
+build-translations: ## Build the translations
 	$(VENV)/bin/pybabel compile -d ihatemoney/translations
 
-update-translations:
+update-translations: ## Update the translations
```
JocelynDelalande: « Extract new translations from source code » ?

Makefile
```diff
@@ -18,41 +18,45 @@ virtualenv: $(PYTHON)
...
+serve: install ## Run the ihatemoney server
+	@echo 'Running ihatemoney on http://localhost:5000'
 	$(PYTHON) -m ihatemoney.manage runserver
 
-test: $(DEV_STAMP)
+test: $(DEV_STAMP) ## Run the tests
 	$(VENV)/bin/tox
 
-release: $(DEV_STAMP)
+release: $(DEV_STAMP) ## Release a new version
```
JocelynDelalande: Maybe add a link to https://ihatemoney.readthedocs.io/en/latest/contributing.html#how-to-release ?

Makefile
```diff
@@ -18,41 +18,45 @@ virtualenv: $(PYTHON)
...
 	$(TEMPDIR)/bin/pip install -U pip
 	$(TEMPDIR)/bin/pip install -Ue "."
 	$(TEMPDIR)/bin/pip freeze | grep -v -- '-e' > requirements.txt
 
-clean:
+clean: ## Clean the virtual environment
 	rm -rf .venv
+
+help: ## Show the help indications
+	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
```
JocelynDelalande: Nice trick :-)

marienfressinaud: Thanks :) This one is not from me but I use it on all the projects I work on, very helpful to self-document the Makefile!

Glandos:
It is maybe a personal preference, but I'd rather write one `.PHONY` section with all targets, instead of writing it as a header for every target.

Otherwise, very nice improvements for developers.

marienfressinaud:
It is debatable indeed but it is usually considered as best practice to have `PHONY` just before the target. The benefit that I can see is it's easier to keep synchronized `PHONY` and targets (it helps to remember to add the `PHONY` when adding a target, and we don't forget to remove the `PHONY` when we delete the target). But I can understand if you still prefer to have a single `PHONY` section since it's quite usual to have it this way :+1:

Glandos:
I totally understand this point of view. You can let it like this for better maintainability.

almet:
Great, thanks for the help :-)

Issue #235, opened 2017-06-23
Float separators


Phyks:
Hi,

It seems the `input` field for the amount is of type `number`. Then, the browser enforces the use of the locale dependent comma separator. For instance, my French Firefox enforces the use of `,` as a float separator.

Problem is that on French keyboards, the `.` key in the numpad area types `.` and not `,`. Then, it is very unintuitive to type `3,2` instead of `3.2`.

Not sure if we should fix it on our side, but I was thinking some pre-processing JS code to convert `3.2` or `3,2` to the correct number representation for the browser could be useful.

EDIT: Just looking for feedbacks, I can contribute this if there is an agreement on this :)

JocelynDelalande:
> Not sure if we should fix it on our side, but I was thinking some
> pre-processing JS code to convert |3.2| or |3,2| to the correct number
> representation for the browser could be useful.
>
Agreed. I'm hitting this too rigid behavior often also.

aavenel:
In fact I think this is a regression, as I remember to write code to handle `,` or `.` without distinction in back-end validation. 

Perhaps using `<input type="number" formnovalidate />` will be enough to get rid of this. https://stackoverflow.com/a/15304154


```
class CommaDecimalField(DecimalField):
    """A class to deal with comma in Decimal Field"""
    def process_formdata(self, value):
        if value:
            value[0] = str(value[0]).replace(',', '.')
        return super(CommaDecimalField, self).process_formdata(value)
```

Phyks:
Yes, either comma and dot are working in the backend code. But the browser does not validate float with dots with French locale. 

Not validating the field works, but "24,a34" will not be discarded either. I'll try to have a look at browsers capabilities. 
-- 
Phyks

JocelynDelalande:
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA512

Possible good compromise showed in :
https://codepen.io/anon/pen/MovRBW

It seems that you cannot customize validation as long as you keep
`type="number"`

-----BEGIN PGP SIGNATURE-----

iQIzBAEBCgAdFiEECtxJPuKxaTvavr2N6h2HoC+Us1MFAllOTdsACgkQ6h2HoC+U
s1OnOxAAgKFxn/ALVRwk1rsy8q9oHIMYM+5Gpw01XY42Rua4UfH2z8WYqmKBqKYE
[...]

PR #413, opened 2018-12-30
Resolve "Basic math operations"

### LINKED ISSUE #234: Basic math operations
Hi,

Not sure how relevant it is, but often I have a bill with multiple entries and I want to enter the sum of some entries in ihatemoney.

I usually have `bc` running in a term next to my browser to handle it, but was thinking that maybe this could be handled directly by `ihatemoney`.

Ullauri:
Closes #234.


JocelynDelalande:
No time to review right now, but I can already send you a big thanks, this feature rocks :-)

PS: I put my housekeeper hat : could you already update the changelog in the PR ? þ<3

Ullauri:
Sure thing! Ill update when I get the chance.

ihatemoney/forms.py
```diff
@@ -44,6 +45,28 @@ def process_formdata(self, value):
...
+    def process_formdata(self, valuelist):
+        if valuelist:
+            error_msg = "Not a valid amount or expression"
+            value = str(valuelist[0]).replace(" ", "").replace(",", ".")
+
+            if not match(r'^[ 0-9\.\+\-\*/\(\)]{0,50}$', value) or "**" in value:
+                raise ValueError(error_msg)
+
+            try:
+                valuelist[0] = str(eval(value, {"__builtins__": None}, {}))
```
Phyks: I would rather avoid the `eval` and use a small dedicated parser instead. The current code with `eval` should be safe, but it is easy to miss it and make a mistake in a further edit of the file, contrary to a safer approach (in my opinion) with a parser.

almet: Thanks for the feature, this will be really useful :-)

After a quick research, it seems that using [`ast.literal_eval`](https://python.readthedocs.io/en/v2.7.2/library/ast.html?highlight=ast#ast.literal_eval) would avoid us exposing a broad scope to the users.

Ullauri: Nice! though `ast.literal_eval` seems to only work for addition and subtraction.

Ullauri: Ended up using [this for parsing](https://stackoverflow.com/a/9558001/5431797) =) (a5819e9)

Phyks: Nice!

Phyks:
My only remaining concern so far is that the `<input id="amount">` is of `type="number"` and then operator signs are not valid in it in modern browsers. There is already an opened issue about using comma or dots for decimal separation, I'm not sure keeping it as `number` is a good idea. What do you think?

Ullauri:
Hmm correct it me if I'm wrong but the amount input field was of `type="number"` because [it was using a DecimalField](https://github.com/spiral-project/ihatemoney/blob/master/ihatemoney/forms.py#L37).  I now [changed it to a StringField](https://github.com/spiral-project/ihatemoney/pull/413/files#diff-ce673044d2d4dc61f111be45a5f61559R48) making it `type="text"` (I inspected the input field on chrome to make sure).

Phyks:
Sorry for the noise :/ that's totally true and I simply missed it during the review of the code (I could not test the code running).

The code looks good to me and this would solve #234 and #235! Really nice! :)

ihatemoney/utils.py
```diff
@@ -1,5 +1,7 @@
 import base64
 import re
+import ast
+import operator
```
almet: Adding `from __future__ import division` would make it possible to return floats from integer divisions (in other words, without it `3/2` returns `1`, and with it it returns `1.5`. I believe we should also add a test for this.

ihatemoney/tests/tests.py
```diff
@@ -1349,6 +1349,69 @@ def test_bills(self):
...
+        del got["creation_date"]
+        self.assertDictEqual(expected, got)
+
+        erroneous_amounts = [
+            "lambda ",  # letters
+            "(20 + 2",  # invalid expression
+            "20/0",  # invalid calc
+            "9999**99999999999999999",  # exponents
+            "2*2*2*2*2*2*2*2*2*2*2*2*2*2*2*2*2*2*2*2*2*2*2*2*2*2",  # greater than 50 chars,
+        ]
```
almet: I see a few missing tests here:
- what happens on integer divisions which should return a decimal value (e.g `3/2`)?
- Some tests with the precedence of operators (multiplication takes precedence on addition for instance)

ihatemoney/forms.py
```diff
@@ -44,6 +45,25 @@ def process_formdata(self, value):
...
+    A class to deal with math ops (+, -, *, /)
+    in StringField
+    """
+
+    def process_formdata(self, valuelist):
+        if valuelist:
+            error_msg = "Not a valid amount or expression"
+            value = str(valuelist[0]).replace(",", ".")
+
+            if not match(r'^[ 0-9\.\+\-\*/\(\)]{0,50}$', value) or "**" in value:
```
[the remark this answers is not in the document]
Ullauri: The 50 char limit was more of a guesstimate as to what the limit should be, a pretty low one in hindsight =). We should def enforce a limit to avoid the `**` issue. I think bumping it up to 200 should be more than what most people would need.

ihatemoney/tests/tests.py
Ullauri: Re: "Some tests with the precedence of operators (multiplication takes precedence on addition for instance)"

I use `((100 + 200.25) * 2 - 100) / 2` [here](https://github.com/spiral-project/ihatemoney/pull/413/files#diff-5aeddb67c56d5aa64202d6e5dd32b188R1366).

almet: I've seen this one, but you would need to remove the parenthesis to test that precedence is correctly taken into account.

ihatemoney/forms.py
```diff
@@ -44,6 +45,25 @@ def process_formdata(self, value):
...
 
+class CalculatorStringField(StringField):
+    """
+    A class to deal with math ops (+, -, *, /)
+    in StringField
+    """
+
+    def process_formdata(self, valuelist):
+        if valuelist:
+            error_msg = "Not a valid amount or expression"
```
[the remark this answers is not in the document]
Ullauri: Agreed.

ihatemoney/utils.py
Ullauri: Good Call! Added import as well as test.

ihatemoney/tests/tests.py
Ullauri: Added test [above.](https://github.com/spiral-project/ihatemoney/pull/413/files#diff-5aeddb67c56d5aa64202d6e5dd32b188R1364)

almet:
Thanks for your changes. The only missing bits are the translations. If you are able to do them (in french only for now), please do, instructions on how to proceed are [here](https://ihatemoney.readthedocs.io/en/latest/contributing.html#as-a-translator). Otherwise, let us know and we'll proceed!

almet:
I've just merged this. A big thanks to you for your work and quick responses to our concerns!

PR #440, opened 2019-07-23
bump bootstrap to latest stable

eMerzh:
Let's use the latest version of bootstrap instead of an alpha release :)


almet:
It seems good to me :-) Thanks for taking the time to do this :+1:

PR #441, opened 2019-07-24
Improve Responsive

eMerzh:
Hi,
Here are a few small steps to improve the responsive situation.
I've start by removing some unused css properties, then adding some missing btn classes here and there.
Then adding responsive classes all around especially within the hamburger.

It's not perfect but better IMHO. I would have liked to rework more the nav bar but didn't want to change too much. I've only choose to put the dashboard inside the options instead of a plain nav item.

Let me know if there are things that does not work for you :)

before
before

after
after

Note: i've also added a prettier for the css file in a distinct commit, for better readability. I you want that i split the PR, let me know

### DIFF
--- a/ihatemoney/static/css/main.css
+++ b/ihatemoney/static/css/main.css
@@ -5,277 +5,326 @@
 /* General */
 
 body {
-    /* For fixed navbar */
-    padding-top: 3.5rem;
-    padding-bottom: 2rem;
+  /* For fixed navbar */
+  padding-top: 3.5rem;
+  padding-bottom: 2rem;
 }
 
 /* Navbar */
 
 .navbar h1 {
-    font-size: 1rem;
-    margin: 0;
-    padding: 0;
+  font-size: 1rem;
+  margin: 0;
+  padding: 0;
 }
 .navbar .secondary-nav {
-    text-align: right;
-    flex-direction: row-reverse;
+  text-align: right;
+  flex-direction: row-reverse;
 }
 .secondary-nav .nav-link {
-    padding: .5rem 1rem;
+  padding: 0.5rem 1rem;
 }
-.navbar-brand{ font-family: 'Lobster', arial, serif; }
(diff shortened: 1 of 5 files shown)


almet:
Thanks! It's worth noting that I've already started working on improving a bit the navigation and the top bar. Related pull requests are at #368 and #369. I should probably have merged them at some point but it's not done. With the work you've done, it seems to me that these other pull requests would be made obsolete. Do you confirm?

Also, if you want to enhance how everything is displayed, go ahead, it's a welcome addition, so don't hesitate to propose bigger changes !

Thanks for taking the time to work on this :-)

eMerzh:
Hi, 
thanks for your quick reply! 
hum yes some of the changes collide, but some other things are missing... like other button / table changes ...

Not sure how to proceed here? Should i revert the "navbar" things and only leave the rest?

almet:
Since I'm the author of the two other pull requests, I will just rework them with regard to the work you've done once it's merged. Let me know if you want to add something here, and otherwise I will just merge it :-)

eMerzh:
Nope nothing for now... You can proceed...
If I have more ideas, I'll double check your PRs...

ihatemoney/forms.py
```diff
@@ -51,6 +52,8 @@ class CalculatorStringField(StringField):
     in StringField
     """
 
+    widget = NumberInput()
```
almet: Can you explicit why this is needed? :-)

eMerzh: It's needed to use input type=number instead of text.... This help display the corresponding keyboard on mobile... 


But you mean in the code?

almet: The thing is that numbers and ` + - * /` operators are actually accepted, and by using the NumberInput I believe it would not be possible to enter operations in this field, isn't it?

almet:
Thanks, we're good to go :-)

PR #442, opened 2019-07-26
Improve already project exists message

### LINKED ISSUE #346: UX : error message for "project's name already taken"

eMerzh:
let's improve this message, 

extracted from #373, can fix #346

not sure if the rest of the pr is still valid or relevant... but at least let's move on with this :)


ihatemoney/translations/fr/LC_MESSAGES/messages.po
```diff
@@ -16,14 +16,12 @@ msgstr ""
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-"Generated-By: Babel 2.6.0\n"
+"Generated-By: Babel 2.7.0\n"
 
 msgid ""
-"Not a valid amount or expression.Only numbers and + - * / operators are "
+"Not a valid amount or expression.Only numbers and + - * / operatorsare "
```
almet: This misses a space :-)

eMerzh: 🤔 it's what pybabel did ....
i've updated the source and re extract  :)

almet:
Thanks!

PR #443, opened 2019-07-26
Remove "Added on" column from bill list.

### LINKED ISSUE #418: Items don't seem to be ordered by date
Bills doesn't seem to be listed by date!

eMerzh:
Let's use "Date" (When?) column as primary time info, and simplify the bill list!

so, i demoted the "Added on" column to a tooltip, and sort the bill list by date.
i had to add popper lib for the tooltip... but shouldn't be a big deal.

let me know what you think
(i hesitate to remove the added on col completely.... not sure how other people are using it 🤷‍♂️)

fix #418


almet:
I like the idea of having this in a tooltip :-)

ihatemoney/models.py
```diff
@@ -175,8 +175,8 @@ def get_bills(self):
             .filter(Bill.payer_id == Person.id)\
             .filter(Person.project_id == Project.id)\
             .filter(Project.id == self.id)\
-            .order_by(Bill.creation_date.desc())\
             .order_by(Bill.date.desc())\
+            .order_by(Bill.creation_date.desc())\
```
almet: Why moving this line?

eMerzh: to sort by date first... like in #418 .... 
feels a bit weird to me that creation date is the sort col too.

almet: Interesting. I wonder if we shouldn't just remove this order_by? Maybe @Phyks has an idea on the matter?

eMerzh: As we only store dates we might want to have another sort column....
But indeed, id could serve this purpose.... Maybe the creation date was there to be a bit more explicit or smth

almet:
It's good for me!

Issue #352, opened 2018-07-05
Project name not visible enough


Nartagnan:
Name of the project not really visible. Could be problematic when having more than 1 project.

almet:
Then that seems to be a style issue. The name of the project appears on the top right. What kind of changes do you suggest to be more visible?

Nartagnan:
I suggest, instead of :
LOGO    |    Bills / Settle / Statistics    |    Lang    |    Project's name
Having
LOGO    |    Project's name (in bold)    |    Bills / Settle / Statistics    |    Lang    |    Settings icon

Must be tested with the smartphone rendering.

almet:
That sounds great! Let's do it !

PR #445, opened 2019-07-29
Navbar fiddeling again

eMerzh:
Hey here :)

seeing some open issues, like #344 or #352 i was fiddeling with another nav rework... i should probably though more about the first iteration...

image

image

Done : 
- [x] enable new langs (NL)
- [x] language switcher is recognisable
- [x] make project name visible (and display the possibility to switch)
- [x] add some nice svg icon
- [x] make the project switcher available on homepage if already logged
- [X] Add link to mobile app and doc in footer
- [x] move dashboard link to footer...


almet:
Yes! That looks better than the current version, thanks.

About your open questions:

> dashboard links
you're proposing that they go away completely : why not. As an admin, I'm feeling it's okay to have to enter the URL myself, but... it's also nice to have a way to discover that this feature exists;

> move the disconnect always visible
I think it's a good choice.

> shouldn't we use shorter/icon only label for disconnect / options
I don't have a strong opinion on this, but I feel more comfortable having the full text.

> maybe add icons on other menu/dropdown
Why not! If that makes the tool more easily usable, I'm +1 for it. We need to make sure that the icons makes the purpose clearer (that's not always the case from my experience).

> is the svg incude is the best way to go (vs fontawesome/glyphicons/... smth like that)

Whatever works, I don't have any opinion on this matter.

eMerzh:
thanks @almet for the feedback!

about the dashboard link... i'm thinking of a way to keep it... 
in the footer 🤔but there is already the open source message...
in the header🤔 isn't already much crowded?
in the "login" page 🤔maybe but where...

>disconnect
ok i'll keep it :)
>short label
agree i'll try to keep them textual... 
>icons
i've added a + for the new project in dropdown... for the method, if we keep the nr of svg low then my method might be ok... if not we'll have to rethink it.

i'll push a few changes soon :)

eMerzh:
what do you think about adding a footer like this


image


bottom links may vary

almet:
Ah, that's nice! I like it :-)

Good call to have a way to connect to a project from the login page if already connected. I wonder if we should also put a link just next to the login form in this case?

The only concert I have is about having the github logo in here! (I'm thinking about moving away from github for this project). If we can find a version control icon instead that would be great.

eMerzh:
👍 about the github logo! ... 
replaced it 
image

i'll plan to see if i can do smth about the login / page a bit later... not sure if we need to duplicate the info there...  🤔 i'll think about it

almet:
Okay. We might want to postpone this, it was just an idea. Let me know when you consider this pull request not a "Work in progress"  anymore; I'll then review the code. Thanks!

eMerzh:
If you are ok with the dashboard link, 

the only thing i'm wondering is we need to merge the project selector and the disconnect part...
but currently i'm more in favor of letting it like this...

if you feel the same, then i think we can move forward...

almet:
After some more thoughts, I think that makes sense to merge the disconnect link and the project selector. The main reason not to is to keep the disconnect link always visible, but I'm not sure it should be. What's the rationale you followed to put the disconnect button always visible?

eMerzh:
the thing is that disconnect might be related to project you are connected to, but not currently on, 
and also admin space (which you can also not be in).
it was a way to clearly display that you are still connected.

I'll try to see if i have a good idea to merge the two while making sure that the current project is visible and make clear that your are connected.

eMerzh:
@almet so, i've merge the menu and put it to the right... visible when connected (to project, to admin dash) on every page.

disconnect is there under... less visible but as the menu is visible, it may be clear that you need some actions or smth to disconnect. not as clear as we can do, but in the meantime the menu is now less crowded...
(btw, i've added the link to the dashboard too if you are connected to it)

wdyt?

image

almet:
Just to let you know: I'm off the hook until end of august. If that helps you going forward, you might attempt to ping other members of the project (@JocelynDelalande, @aavenel, @0livd) to get feedback. Cheers!

eMerzh:
ok @almet have great vacations!
others, if you want to comment ;-) ?

(just pushed a fix ...not caused by this but still... for flash message placement)

other than that i'm ready

eMerzh:
@almet & friends, 
i've played also with the homepage a bit... not sure if ready yet, but here is a sneak peak..
image

should i top it in this MR or create a new one?

Glandos:
Personal opinion: I like the gradient, but I think that button should be the same color. Sure, they have a different action, but they all are buttons that engage an action at the same level: enter a project (either new or existing).

eMerzh:
I understand your point....The rationale behind that was to clearly differentiate the two actions... My feeling was that it was hard to perceive the distinction between the two and sometimes often feels like a single form.

Any other idea? Or anyone feeling the same?

almet:
That looks good to me. I feel a bit weird about having two colors but the rationale seems okay to me, so I'm +0 for it!

almet:
We might want to differentiate the whole « create a new project » input ?

eMerzh:
ok i'll put it back to primary ... the intent is not to confuse people... maybe having clear boxes is already a good enough differentiator ;)

(pushed)

almet:
And it's in master! Thanks for your work, again :-) I feel that a new release is coming.

eMerzh:
awesome 👍 
i might have another small thing in the pipe...  if nothing comes within a week, it means that it might take a bit more time :p

PR #446, opened 2019-07-29
[docker] Add fixed release vs Nightly mode

bnjbvr:
See commit message. Alternatively, the non-Nightly version could just download what's on Pypi; I don't have a clear criteria of choice here.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,7 @@
-FROM python:3.6-alpine
+FROM python:3.7-alpine
 
-RUN apk update && apk add gcc libc-dev libffi-dev openssl-dev &&\
-    mkdir /ihatemoney &&\
-    mkdir -p /etc/ihatemoney &&\
-    pip install --no-cache-dir gunicorn pymysql
-
-COPY . /ihatemoney
-ARG INSTALL_FROM_PYPI="False"
-RUN if [ "$INSTALL_FROM_PYPI" = True ]; then\
-    pip install --no-cache-dir ihatemoney ; else\
-    pip install --no-cache-dir -e /ihatemoney ; \
-    fi
-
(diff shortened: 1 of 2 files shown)


almet:
This looks good to me, and having an easy way to install the latest master from github seems a nice idea.

almet:
I would prefer to rely on PyPI rather than on github tags for the download. There is not reason that the packages downloaded from github would be pip-installable.

Defaulting to `pip install ihatemoney` instead seems better to me.

bnjbvr:
I've got local changes to do this, but I am now wondering: this means that the image won't be able to use local code, which could be an issue for development (maybe? some people use Docker for development purposes too). So I'm wondering if I should introduce a third mode (then I'd use an env variable to control this: 1. code gets copied into the Docker image, or 2. pip install or, 3. git clone/update from this repository). Any opinions about this?

almet:
So, I'm not using docker myself for development, so I'm not familiar with how things should be done.

That being said, requirements for dev aren't the same as for "normal" install itself. One should install `dev-requirements.txt` in addition, also requirements are potentially pinned in the `setup.py` but not in the `requirements.txt` files.

Long story short : if we want to also support the "dev" mode, then I think we should run `pip install -e . && pip install -r requirements.txt && pip install -r dev-requirements.txt`.

+1 for having this choice in the docker image if that helps people.

bnjbvr:
Ok, in this case it might even be simpler to propose another Docker image for development purposes, since the current image wasn't thought for development anyways. So I'd prefer this keep the dev feature out of scope for the current PR, if that's fine with you too.

almet:
Okay, that's fine for me, thanks for the dockerfile update!

PR #450, opened 2019-08-19
Move Export to file instead of form

eMerzh:
Form are the worst, especially for elements with a few (less than 4-5 ) values.
It break discoverability and should be avoided IMHO.
So here is a proposal to change it to a simple link... 
i've added a small description too, because i was not sure what the "settle" means.

Before:
image

After:
image


almet:
This is soo much better this way! It really makes a difference to have someone with UX skills around.

I'm thinking about making a blog post with the before / after changes to summup what changed. What do you think?

Alexis Métaireau

-------- Message d'origine --------

eMerzh:
(tests fixed ...)

yeah no problem for the blog post, seems like a good idea... ~people~ I like screenshots in changelogs :)

Glandos:
> Form are the worst, especially for elements with a few (less than 4-5 ) values.
> It break discoverability and should be avoided IMHO.

Plus, you can now share the download link. It maybe seldomly used, but it's an interesting feature.

almet:
That's just perfect, big up!

PR #451, opened 2019-08-20
Rework invite people to project

eMerzh:
Hey, another small improvement here (at least i think 😉 )

Summary:
- small ie fixes for icons
- move send invite links to right side with participants
- change some labels
- add some icons
- split send invite explanation into distinct methods
- tiny improvements on send email form

Before:
image

After:
image

Before:
image

After:
image

please check the wording and if you like the split ... i was feeling that the previous text was a bit long-ish and messy (here you can send invite, or password, but put emails down there , or you can use the link, here is the form )

PR #453, opened 2019-08-26
fix: add table layout for sharing options

eMerzh:
follow up of #451

didn't pushed it completely 🤕 
here is the end of it ...

sorry again


almet:
Cheers!

Issue #412, opened 2018-12-28
Expose invitation link creation via the API


Natim:
Is it possible to get an invitation link for a project using the API?
Can we make it possible ?

almet:
I think that currently it's not possible. But it shouldn't be too complicated to do via the API!

almet:
This means different things:
1. Create an endpoint which can generate a new auth token
2. Accept a new authentication scheme, with the auth token

Natim:
As a first step, we could also generate a base64 basic token with an embedded payload containing the user/password in the basic format.

almet:
It's now possible to authenticate with a token to the API, I believe what's missing is the way to create this token from the API, and this should be pretty easy to do.

JoseAntonioTorre:
I want to help here! Could I?

almet:
Sure thing!

JoseAntonioTorre:
I am trying to understand which is the problem that this issue tries to solve.
I have been reading related issues and I notice that there is a share link that can be used to access to a project. Which is the use case that we are trying to support?

almet:
Currently, you can send a share link to other people so they can login to the project without knowing its password.

This mechanism is currently onluy supported trough the web interface, and not trough the API. This issue aims at supporting this mechanism: generating the invitation link, and then using it to authenticate on the API.

Hope it's clearer, let me know if not.

indatwood:
This was fixed with #504

PR #454, opened 2019-08-26
api: add bearer token support

eMerzh:
In sharing options, we have a link with a token.
It might be a good idea to have, as API auth method, this token.
so apps can catch those links and use it directly

I think this is probably not enough because it doesn't let the app discover the project id ...
not sure if we need to include some endpoint that does : token => project id
or include project id in clear text in the url...

### DIFF
--- a/ihatemoney/api.py
+++ b/ihatemoney/api.py
@@ -26,12 +26,27 @@ def need_auth(f):
         auth = request.authorization
         project_id = kwargs.get("project_id")
 
+        # Use Basic Auth
         if auth and project_id and auth.username == project_id:
             project = Project.query.get(auth.username)
             if project and check_password_hash(project.password, auth.password):
                 # The whole project object will be passed instead of project_id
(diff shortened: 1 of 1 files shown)


almet:
Related to #412, where I suggested to create an endpoint which can generate a new auth token and then accept a new authentication scheme, with the auth token.

eMerzh:
@almet yes can probably relate, but here it was more about using this token than getting one.... Any prefs about the project id?

almet:
Sorry for the delay answering you here! As currently tokens are only generated per project, it makes sense to have a way to find back the project they're related to, for sure.

Including the projectid in the URL seems the simplest way to handle this, so I would go for this, I think!

almet:
@eMerzh hey! Hope you're doing well! Should we move forward on this?

eMerzh:
mmh yes 👍
for me this can still be useful. If you feel the same and have nothing more to fix, let's move it on ;-)

PR #456, opened 2019-08-27
use zopflipng to the max to recompress png

Glandos:
Small reductions, to improve load time.


almet:
Thanks, can you indicate how you did that? Thanks!

Glandos:
- Install zopflipng (On Debian `apt install zopfli`
- Run it like the last example in the help page: `zopflipng --iterations=500 --filters=01234mepb --lossy_8bit --lossy_transparent --prefix`
- Since `--prefix` add `zopfli_` as output image prefix, check that all images are OK, and rename.

For the record, running with all switches did a little little better than without any switches (~1‰). So it's only useful for production.

almet:
Thanks ! We probably want to add these instructions to the docs somewhere :-)

PR #457, opened 2019-08-31
Add placeholder to bill list

eMerzh:
Let's add a placeholder when there is no bills and tweak the message when there is also no members

image


ihatemoney/templates/list_bills.html
```diff
@@ -106,31 +108,50 @@ <h3 class="modal-title">{{ _('Add a bill') }}</h3>
...
     {% else %}
-    <div class="py-3"></div><p>{{ _("Nothing to list yet. You probably want to") }} <a href="{{ url_for(".add_bill") }}" data-toggle="modal" data-target="#bill-form">{{ _("add a bill") }}</a> ?</p></div>
+        <div class="py-3 d-flex justify-content-center empty-bill">
+        <div class="card d-inline-flex p-2">
+            <div class="card-body text-center text-muted">
+                <i class="icon icon-white hand-holding-heart">{{ static_include("images/hand-holding-heart.svg") | safe }}</i>
+                <h3>{{ _('No bills')}}</h3>
+                <p>
+                    {{ _("Nothing to list yet.")}} <br />
+                    {{ _(" You probably want to") }}
```
almet: The space at the beginning of the sentence is here on purpose?

ihatemoney/tests/tests.py
```diff
@@ -178,7 +178,10 @@ def test_invite(self):
         self.client.get("/exit")
         # Test that we got a valid token
         resp = self.client.get(url, follow_redirects=True)
-        self.assertIn('You probably want to <a href="/raclette/add"', resp.data.decode('utf-8'))
+        self.assertIn(
+            'You probably want to <a href="/raclette/members/add"',
+            resp.data.decode('utf-8')
+        )
```
almet: Can you add a test checking that if members exists, the URL is proposing to add a new bill?

ihatemoney/static/images/hand-holding-heart.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 576 512"><path d="M275.3 250.5c7 7.4 18.4 7.4 25.5 0l108.9-114.2c31.6-33.2 29.8-88.2-5.6-118.8-30.8-26.7-76.7-21.9-104.9 7.7L288 36.9l-11.1-11.6C248.7-4.4 202.8-9.2 172 17.5c-35.3 30.6-37.2 85.6-5.6 118.8l108.9 114.2zm290 77.6c-11.8-10.7-30.2-10-42.6 0L430.3 402c-11.3 9.1-25.4 14-40 14H272c-8.8 0-16-7.2-16-16s7.2-16 16-16h78.3c15.9 0 30.7-10.9 33.3-26.6 3.3-20-12.1-37.4-31.6-37.4H192c-27 0-53.1 9.3-74.1 26.3L71.4 384H16c-8.8 0-16 7.2-16 16v96c0 8.8 7.2 16 16 16h356.8c14.5 0 28.6-4.9 40-14L564 377c15.2-12.1 16.4-35.3 1.3-48.9z"/></svg>
```
almet: Love this one :-)

ihatemoney/templates/list_bills.html
eMerzh: Oups no...

ihatemoney/tests/tests.py
eMerzh: of course... added :)

almet: I think tests should be more separate from each other. Here we're in the function testing that the invitation is working by email, so we probably don't want to change this one, and create two new test functions in order to test the new behavior. What do you think?

ihatemoney/tests/tests.py
```diff
@@ -298,6 +301,12 @@ def test_membership(self):
         result = self.client.get("/raclette/")
         self.assertIn("fred", result.data.decode('utf-8'))
 
+        # Empty bill list should now propose to add bills
+        self.assertIn(
+            'You probably want to <a href="/raclette/add"',
+            result.data.decode('utf-8')
```
almet: For this test it seems to be an acceptable place, but for the other, I would prefer to put it somewhere else :-)

eMerzh: move them to a specific test function.
tell me if it's what you meant ;-)

almet:
Perfect!

PR #459, opened 2019-09-02
Zopfli in makefile

### LINKED ISSUE #458: Add png recompression howto in the docs
- Install zopflipng (On Debian `apt install zopfli`)
- Run it like the last example in the help page: `zopflipng --iterations=500 --filters=01234mepb --lossy_8bit --lossy_transparent --prefix`
- Since` --prefix add zopfli_` as output image prefix, check that all images are OK, and rename.

Glandos:
Fix #458 
The target is open for more compressions, if it's ever required some day…

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ DEV_STAMP = $(VENV)/.dev_env_installed.stamp
 DOC_STAMP = $(VENV)/.doc_env_installed.stamp
 INSTALL_STAMP = $(VENV)/.install.stamp
 TEMPDIR := $(shell mktemp -d)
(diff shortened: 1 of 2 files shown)


eMerzh:
i would say that my secret plan is to move pngs to svg, 
but anyway, this is nice and handy 👍

Glandos:
You're right, but they have different usage. For example, a background image like gradient.png is better as lossless compressed image than an SVG, since it doesn't need to scale.

Don't forget that SVG is harder to draw on the client side :)

almet:
Perfect, thanks !

PR #460, opened 2019-09-06
Add more human "for who?" description.

almet:
The limit for displaying "Everyone except …"
instead of the ower's list is completely empirical.

Supersedes #335

### DIFF
--- a/ihatemoney/templates/list_bills.html
+++ b/ihatemoney/templates/list_bills.html
@@ -123,7 +123,13 @@
                 </td>
                 <td>{{ bill.payer }}</td>
                 <td>{{ bill.what }}</td>
-                <td>{{ bill.owers|join(', ', 'name') }} </td>
+                <td>{% if bill.owers|length == g.project.members|length -%}
+                {{ _("Everyone") }}
+                {%- elif bill.owers|length > g.project.members|length / 2 + 1 -%}
+                {{ _("Everyone but %(excluded)s", excluded=g.project.members|reject('in', bill.owers)|join(', ', 'name')) }}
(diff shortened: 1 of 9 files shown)


JocelynDelalande:
Thanks for reviving this topic :-).

.po/.mo edits are missing, no ?

ihatemoney/templates/list_bills.html
```diff
@@ -123,7 +123,13 @@ <h3 class="modal-title">{{ _('Add a bill') }}</h3>
                 </td>
                 <td>{{ bill.payer }}</td>
                 <td>{{ bill.what }}</td>
-                <td>{{ bill.owers|join(', ', 'name') }} </td>
+                <td>{% if bill.owers|length == g.project.members|length -%}
+                {{ _("Everyone") }}
+                {%- elif bill.owers|length > g.project.members|length / 2 + 1 -%}
+                {{ _("Everyone excepted %(excluded)s", excluded=g.project.members|reject('in', bill.owers)|join(', ', 'name')) }}
```
eMerzh: isn't it `except`? not sure about it

Glandos: I guess so. But I would go for "Everyone but […]" because it is shorter to read, and it is the goal of this change.

almet: « Everyone but » seems better to me :-)

almet:
> .po/.mo edits are missing, no ?

I'm actually not sure how we should handle these? I believe it's part of the releasing job to run build-translations from time to time and then the translators will update on the weblate interface, no?

PR #461, opened 2019-09-07
remove compatibility of py34

eMerzh:
Since newest dependencies (flask) have drop python3.4 and it's a kind of old release (in debian it's Old Old stable...)

so i think it's fair ... 
(it unbreak the tests)


almet:
You're right, there is no point in maintaining this version of python. Merging.

Issue #455, opened 2019-08-27
Password recovery doesn't display any validation message


almet:

LeoMouyna:
Can you please describe a bit this issue ? Which page is involved ? What are your user actions ? And what is the expected behavior ?  
Thanks :)

almet:
Hi, when you reset a password, a mail is sent, but no validation message is shown to the user. When I did it last time, I clicked multiple times, and… received multiple emails ! So everything seems to be working, *but* no message is shown to the user when the mail is sent.

LeoMouyna:
Seems like I got a message on ihatemoney.org.
![validationMessage](https://user-images.githubusercontent.com/17725363/64620089-8857c080-d3e3-11e9-9203-dbad1cd81a4a.png)

almet:
Thanks for taking the time to look into this. Strange! Maybe this version of IHM fixes it, it was definitely the case with the previous release.

That being said, I believe the UX here could be better : rather than having a "flash message", we could redirect the user to another page where they're displayed a text "A link to reset your password has been sent to you, please check your emails", and it's not possible to click again on the "submit" button.

Do you want to work on this?

LeoMouyna:
Why not, but not sure if I will be able to start it before Thursday. So feel free to do it if you want it sooner :smile:

almet:
Oh, Thursday seems perfect. We don't have any need to rush. If you want to do it, please do it with your own agenda! :-)

PR #464, opened 2019-09-12
fix: display validation message on password reminder

LeoMouyna:
Create a new route with a new generic page that display a title and a message on a <p> tag. This route will be triggered after succesfully submit password reminder form

See issue #455

### DIFF
--- a/ihatemoney/templates/display_message.html
+++ b/ihatemoney/templates/display_message.html
@@ -1,0 +1,8 @@
+{% extends "layout.html" %}
+
+{% block content %}
+<h2>{{ _(title) }}</h2>
+<p>
+    {{message}}
+</p>
+{% endblock %}
--- a/ihatemoney/web.py
+++ b/ihatemoney/web.py
@@ -271,16 +271,21 @@ def remind_password():
         if form.validate():
             # get the project
             project = Project.query.get(form.id.data)
-
             # send a link to reset the password
             password_reminder = "password_reminder.%s.j2" % get_locale().language
             current_app.mail.send(Message(
                 "password recovery",
                 body=render_template(password_reminder, project=project),
                 recipients=[project.contact_email]))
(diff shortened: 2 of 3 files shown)


ihatemoney/web.py
```diff
@@ -271,18 +271,23 @@ def remind_password():
...
                 recipients=[project.contact_email]))
-            flash(_("A link to reset your password has been sent to your email."))
+            return redirect(url_for(".password_reminder_sent"))
 
     return render_template("password_reminder.html", form=form)
 
 
+@main.route("/password-reminder-sent", methods=["GET"])
+def password_reminder_sent():
+    message = "A link to reset your password has been sent to you, please check your emails"
```
almet: nit : can you add a dot at the end of the sentence? Thanks!

ihatemoney/templates/display_message.html
```diff
@@ -0,0 +1,8 @@
+{% extends "layout.html" %}
+
+{% block content %}
+<h2>{{ _(title) }}</h2>
+<p>
+    {{message}}
+</p>
```
almet: I believe it would be nice to have a link here to go back to the homepage, and that means that having a generic page to display messages here would not be enough.

I see different ways to go from here : either we extend this new `display_message` template and add a section with a specific text after the message, or we do a `password-reminder-sent.html` template. I believe the latter is the KISS option.

LeoMouyna:
I'll handle those changes  tonight :smile:

almet:
Thanks for the update. I think I wasn't clear in my last answer : I believe it would be better to have only one template, with the text displayed, rather than going with the inheritance like you did.

Sorry, I probably led you to the « wrong path » by mentioning it. Let me know if you want to rework this, otherwise I can do it. Cheers!

LeoMouyna:
Those are easy changes I can do it :wink:

almet:
thank you :-)

Issue #429, opened 2019-03-18
Allow to attach files (as proof of expense)


Norde:
Could the possibility of attaching a file to an expense (like proof / invoice) be feasible?

almet:
It's technically feasible, but I don't think this is the scope we want for this software. What's your usecase here?

Norde:
Just being able to attach proof to an expense (especially for large expenses).

It's very convenient for roommate (anyone can therefore check the accounts and trust reign).
It also makes rent receipts easily accessible to everyone.

I think it's a logical function for shared budget management software but it may not be the desired goal.
You may prefer to keep an application as simple / light as possible?

almet:
Yep, one of the goals of IHM is to stay as simple as possible. Handling these pieces also means that the server contains a lot more valuable data than before. I'm +0 on this one. If anyone else thinks this is needed, don't hesitate to comment here :-)

0livd:
I agree with @almet. Another idea might be to add a new `External link` icon in the `Actions` menu, which can link to whatever this bill is related to, e.g. a file stored on nextcloud/dropbox. What do you think ?

almet:
If that solves the problem for @Norde, why not!

Norde:
This would be a very good solution that would not weigh down HMI!

almet:
@Norde are you willing to work on this?

Norde:
My knowledge of python is nil but I will take a look :)

I think that the opening of a simple field "url" is within my capabilities.
For integration with nextcloud I make no promises.

PR #467, opened 2019-09-15
feat: Optional link field in bill form.

LeoMouyna:
An optional field has been added to the bill form to add a link to a real bill. A new action button allow user to see this bill. Breaking change with Bill model update for database, a migration is needed.

See issue #429.


ihatemoney/forms.py
```diff
@@ -139,6 +139,7 @@ class BillForm(FlaskForm):
     what = StringField(_("What?"), validators=[DataRequired()])
     payer = SelectField(_("Payer"), validators=[DataRequired()], coerce=int)
     amount = CalculatorStringField(_("Amount paid"), validators=[DataRequired()])
+    document = URLField(_("Document"), validators=[Optional()], description="A link to your bill.")
```
almet: The description here isn't clear enough. What do you think about "A link to an external document, related to this bill"

almet: Also, if this is a link, I feel that the name here isn't right, because it isn't explicit about what is stored (it's not a document, it's an URL, or a link). What about ``external_link`` or ``document_url` rather than "document"?

LeoMouyna: I don't mind just let me know what you think it's the best for your project :smile:.
I'll change the description and the field name for external_link.
Just a remark, I don't know how to show that it's an optional field, it looks like other that are required and how to display the html5 field check error. If you have an hint tell me :loudspeaker:

almet: let's go for ``external_link`` then. You can put ``(Optional)`` in plain text, I think?

ihatemoney/models.py
```diff
@@ -358,12 +358,13 @@ def delete(self, project, id):
...
     creation_date = db.Column(db.Date, default=datetime.now)
     what = db.Column(db.UnicodeText)
+    document = db.Column(db.UnicodeText)
 
     archive = db.Column(db.Integer, db.ForeignKey("archive.id"))
 
     @property
     def _to_serialize(self):
-        return {
+        dict = {
```
eMerzh: nit: not sure i would use dict as name... as a reserved word....

ihatemoney/models.py
```diff
@@ -372,6 +373,9 @@ def _to_serialize(self):
             "creation_date": self.creation_date,
             "what": self.what,
         }
+        if self.document:
```
eMerzh: nit: not sure why we wouldn't return a "document"/url/... if there is none... like couldn't return a "null"/None?

ihatemoney/templates/list_bills.html
```diff
@@ -128,6 +128,9 @@ <h3 class="modal-title">{{ _('Add a bill') }}</h3>
                 <td class="bill-actions">
                     <a class="edit" href="{{ url_for(".edit_bill", bill_id=bill.id) }}" title="{{ _("edit") }}">{{ _('edit') }}</a>
                     <a class="delete" href="{{ url_for(".delete_bill", bill_id=bill.id) }}" title="{{ _("delete") }}">{{ _('delete') }}</a>
+                    {% if bill.document %}
+                    <a class="see" href="{{ bill.document }}" target="_blank" title="{{ _("see") }}">{{ _('see') }} </a>
```
eMerzh: i would add `ref="noopener"` and maybe noreferrer ... but the latest is more questionable...

eMerzh:
If you could change the help text style to match latest bootsrap...
https://getbootstrap.com/docs/4.1/components/forms/#help-text

ihatemoney/templates/list_bills.html
```diff
@@ -128,6 +128,9 @@ <h3 class="modal-title">{{ _('Add a bill') }}</h3>
                 <td class="bill-actions">
                     <a class="edit" href="{{ url_for(".edit_bill", bill_id=bill.id) }}" title="{{ _("edit") }}">{{ _('edit') }}</a>
                     <a class="delete" href="{{ url_for(".delete_bill", bill_id=bill.id) }}" title="{{ _("delete") }}">{{ _('delete') }}</a>
+                    {% if bill.document %}
```
eMerzh: i know i kinda match the lines 👆but title exactly like the next note seems useless...
https://inclusive-components.design/tooltips-toggletips/
i would use some documentation in place of this...

almet:
@LeoMouyna Hi! Hope you're doing well. Did you got our latest reviews?

LeoMouyna:
Hey @almet !
Sorry I'm a bit busy those weeks, let's say I'll do it during the week-end or Tuesday for the latest ;)

almet:
Great, thanks!

LeoMouyna:
Sorry I'm in trouble with database migration scripts. I don't know how to downgrade correctly the changes made in my last commit. I'll create an other PR with all remarks.
Here is the error that is displayed and I don't know flask and sqlachemy enough to handle it correctly.

> sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) near "DROP": syntax error
[SQL: ALTER TABLE bill DROP COLUMN document]

And if you want all the stacktrace: https://pastebin.com/nkEZnZ6D

ihatemoney/models.py
LeoMouyna: I'm used to this behavior but if you think that it's useful to have a field returning None then I can add it like other fields.

ihatemoney/templates/list_bills.html
LeoMouyna: Title is useful on mouse over events to display a small text box. I don't want to refactor this for now but according to your link we can use tooltips in future.

LeoMouyna:
> If you could change the help text style to match latest bootstrap... it could be really nice!
> https://getbootstrap.com/docs/4.1/components/forms/#help-text
@eMerzh, I did some changes but I didn't put an aria-describedby on input tags. If you know how to add it feel free to add this to my new PR: https://github.com/spiral-project/ihatemoney/pull/479

PR #469, opened 2019-09-16
Move translations in template and translate link

eMerzh:
Let's allow the link to be translated, 
and also make sure that the string is extracted in setting it inline... no need for a var here.


almet:
cc @LeoMouyna.

almet:
Thanks for doing these reviews @eMerzh. It makes me realize I'm too lazy when doing them myself. Need to get back on track on the matter.

PR #472, opened 2019-09-23
Update from Weblate

weblate:


DiegoCaraballo:
@almet It was 100% translated into Latin American Spanish

Issue #353, opened 2018-07-05
Possibility to had the same name with juste blank space


Nartagnan:
No visible difference when adding "Bob" and "Bob ".

Glandos:
Project name or members should be stripped before validation

almet:
Yep, that seems to be an easy thing to do. If anyone wants to go for it, don't hesitate!

PR #478, opened 2019-09-27
I delete initial and final blanks of members that are added in form.py

DiegoCaraballo:
Modify the validate_name method of forms.py so that when entering a new member remove the blanks at the beginning and end of the name.

See issue #353


almet:
Hi, thanks for the pull request. However, I've already started solving this, and it's about to be ready to merge. You can have a look at how I did it in https://github.com/spiral-project/ihatemoney/pull/367, which I believe is a better approach than the one you took here.

Also, note for later : you need to add tests when you create a pull request ;-)

PR #479, opened 2019-09-28
feat: Optional field 'external link' in bill form.

LeoMouyna:
An optional field has been added to the bill form to add a link to a real bill. A new action button allow user to see this bill. Breaking change with Bill model update for database, a migration is needed.

See issue #429.


ihatemoney/forms.py
```diff
@@ -146,6 +146,8 @@ class BillForm(FlaskForm):
     what = StringField(_("What?"), validators=[DataRequired()])
     payer = SelectField(_("Payer"), validators=[DataRequired()], coerce=int)
     amount = CalculatorStringField(_("Amount paid"), validators=[DataRequired()])
+    external_link = URLField(_("External link"), validators=[Optional(
+    )], description="A link to an external document, related to this bill")
```
almet: The description should probably be translatable too.

almet:
Thanks! We're almost there :-)

ihatemoney/forms.py
LeoMouyna: I can easily add a `_("A link to an external document, related to this bill")` but I don't know how to generate correctly the translation even after a `make update-translation` with some changes in messages.po. Am I missing something ?

almet: `make update-translations` will update the strings, and then the translation process is done via Weblate (http://hosted.weblate.org/). This is currently undocumented, so you didn't have any way to know, sorry.

ihatemoney/templates/list_bills.html
```diff
@@ -134,6 +134,9 @@ <h3 class="modal-title">{{ _('Add a bill') }}</h3>
                 <td class="bill-actions">
                     <a class="edit" href="{{ url_for(".edit_bill", bill_id=bill.id) }}" title="{{ _("edit") }}">{{ _('edit') }}</a>
                     <a class="delete" href="{{ url_for(".delete_bill", bill_id=bill.id) }}" title="{{ _("delete") }}">{{ _('delete') }}</a>
+                    {% if bill.external_link %}
+                    <a class="see" href="{{ bill.external_link }}" ref="noopener" target="_blank" title="{{ _("see") }}">{{ _('see') }} </a>
```
almet: Should we replace "See" by "See attached document"?

LeoMouyna: I have no preference tell me what's the best in your opinion :)

almet: I think [explicit is better than implicit](https://www.python.org/dev/peps/pep-0020/) :-)

almet:
Hmm, actually, for the "see attached document", I believe it would be better to stay with "see". It's less explicit, but due to where this is placed, we probably want to keep it short.

So… Let's merge!

PR #480, opened 2019-10-01
Paginate the list of bills

zorun:
This is a work-in-progress proposal to paginate the list of bills. The main motivation is that I have 1095 bills, and the list of bills is starting to be slow to display!

Before investing time in design etc, I would like to first gather some feedback on the idea itself.

By default, the first page is displayed (most recent bills). Older bills can be displayed with the `page` URL parameter:

```
$URL/project/?page=42
```

Documentation for this feature in flask-sqlalchemy is here:

https://flask-sqlalchemy.palletsprojects.com/en/2.x/api/#flask_sqlalchemy.BaseQuery.paginate
https://flask-sqlalchemy.palletsprojects.com/en/2.x/api/#flask_sqlalchemy.Pagination

Some things to improve:

- decide on the number of bills to display on each page (20 items by default, which is probably fine)
- decide what happens when fetching an out-of-range page. Currently it just displays an empty table, but we can also return a 404.
- maybe add a link to the previous and next page above the table?
- better style for the list of pages
- implement the same thing in the API! (would be useful for MoneyBuster, it's also really slow to synchronise with 1000 bills)

Here is a screenshot, with only 2 items per page for testing:

screenshot demo

### DIFF
--- a/ihatemoney/static/css/main.css
+++ b/ihatemoney/static/css/main.css
@@ -272,6 +272,15 @@ footer .footer-left {
   margin-top: 30px;
 }
 
+#previous-page {
+  margin-top: 30px;
+}
+
+#next-page {
+  margin-top: 30px;
+  padding-left: 2em;
+}
+
 /* Avoid text color flickering when it loose focus as the modal appears */
 .btn-primary[data-toggle="modal"] {
   color: #fff;
--- a/ihatemoney/templates/list_bills.html
+++ b/ihatemoney/templates/list_bills.html
@@ -97,13 +97,23 @@
         </div>
     </div>
 
-    {% if bills.count() > 0 %}
+{% if bills.pages > 1 %}
+    <span id="previous-page" class="float-left">
+      <a class="btn btn-outline-secondary float-left {% if bills.page == 1 %}disabled{% endif %}" href="{{ url_for('main.list_bills', page=bills.prev_num) }}">&laquo; {{ _("Newer bills") }}</a>
+    </span>
+
+    <span id="next-page" class="float-left">
+      <a class="btn btn-outline-secondary float-left {% if bills.page == bills.pages %}disabled{% endif %}" href="{{ url_for('main.list_bills', page=bills.next_num) }}">{{ _("Older bills") }} &raquo;</a>
+    </span>
+{% endif %}
+
+    {% if bills.total > 0 %}
         <div class="clearfix"></div>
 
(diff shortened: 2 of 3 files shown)


almet:
Thanks for starting working on this :-)

On the things to improve :

> decide on the number of bills to display on each page (20 items by default, which is probably fine)
I would go for 100 bills displayed by default.

> decide what happens when fetching an out-of-range page. Currently it just displays an empty table, but we can also return a 404.

404 seems to be the sane thing to do.

> maybe add a link to the previous and next page above the table?

I wonder if we should try to use autoscroll here, in order to load the bills as they're needed?

> better style for the list of pages

Yes! (But I'm bad at this :()

> implement the same thing in the API! (would be useful for MoneyBuster, it's also really slow to synchronise with 1000 bills)

I wonder if paginating would make it faster? I believe it would just add some overhead here, don't you think? But, how to enhance the API is probably an issue by itself. We could probably enhance how we synchronize, but that's a bunch of changes.

I've been working (with @Natim, who's a contributor here) on [the Kinto project](https://kinto-storage.org), which is basically an JSON HTTP API, made to be efficient. We might want to borrow some principles we've implemented there :
- (Pagination)[http://docs.kinto-storage.org/en/stable/api/1.x/pagination.html]
- How (Timestamps)(http://docs.kinto-storage.org/en/stable/api/1.x/timestamps.html) are used to ease the syncing (and avoid downloading again stuff that's already been retrieved).

almet:
Hey @zorun, hope everything's going well. Are you okay with this, or do you need any help?

almet:
Hey. This seems stalled. I believe that until we have some news from @zorun we should close the pull request (it's been more than 3 months)

zorun:
@almet yeah sorry, didn't have time to work on this.

I just rebased and pushed, with a few changes:

- 100 bills per page by default
- 404 page when requesting an out-of-range page
- add a "Page:" string in front of the page numbers

This should be ready for merging, unless we want to have a configuration option to choose the number of bills per page.

almet:
Neat! You can run black on web.py to fix the travisCI error.

I believe that if there is no pagination, the section is still shown, can we change this?

zorun:
Right, it should be better now.

I have also improved the rendering using Bootstrap classes:

ihm-paginate

For the screenshot I displayed just 1 item per page, this is of course not the case for the final commit (100 items per page).

@almet does it sound good? I am not a designer but bootstrap makes a lot of things easier :)

almet:
yeah, it's great, thanks!

PR #482, opened 2019-10-06
Thank you, black.

almet:


almet:
Used black to enhance a bit the codebase. Not sure when to merge this, since it'll probably create conflicts with the to-be-merged PRs.

eMerzh:
shouldn't we add black in the dev dependancies or smth ?

almet:
Yes, I believe that would be a nice addition :-) I'll cook up something, maybe providing a pre commit hook in the docs?

eMerzh:
i personnaly don't like pre commit hooks but yeah why not add it in the doc 

(i prefer format on save).

btw we might want to loosen flake8 on formating issues (and let black do the trick)

almet:
I prefer format on save too ;-)

About removing the flake8 CI integration, we could probably replace it with "black --check"? This way, individuals might prefer format-on-save or something else.

eMerzh:
Yep sound good for black check.
I wasn't talking about removing ALL flake8, but maybe rules about "formatting" that should be managed by black (the thing actually failing the CI test for example).

so black to check and fix formatting issues (quotes, line length, ...) flake8 for the other things like variable names, arguments checks, ...

almet:
Replaced by #488 and #489.

PR #488, opened 2019-10-11
Some general cleaning

almet:


almet:
I've separated #482 in two distinct pull request, to ease the flow. I think this can be merged straight ahead.

PR #489, opened 2019-10-11
Add black in the tox matrix.

almet:


almet:
So, we're finally there. Can you have a quick look @eMerzh?

eMerzh:
awesome! looks pretty good from here...
i would also add a note in readme and or contributing (in doc) or smth similar ...to point potential contributor to this :)

eMerzh:
thanks @almet for the note ;)

PR #492, opened 2019-10-12
German translation

freetim:
Hi,

I added the german translation. I saw the thing with weblate to late and hope it's not a problem. In the future I would use weblate if prefered but I don't want to do things twice.

Btw. if I would use weblate is there a way to see the translations on the website before adding them finaly?

Grettings Tim


almet:
Thanks, I see no problem with this approach so far.

> Btw. if I would use weblate is there a way to see the translations on the website before adding them finaly?

I don't understand what you mean?

freetim:
My question was about the weblate-website. During my local developing I update the local site so I can see the translated website how it would look. But if I translate on the weblate-website I'm asking me if there is a way to see this translation in the ihatemony-website before starting the pull request.

almet:
I don't think there is a way to get this directly, but you probably can see your changes by installing IHM locally and updating the translation files, with an export from weblate.

freetim:
Then developing localy with poedit is much easier.

almet:
If this is better for you, please do it like that :-) Weblate will get the changes as soon as they're merged into the master branch.

PR #494, opened 2019-10-14
Added install of ihatemoney module to documentation

JoseAntonioTorre:
I have tried to follow instructions in https://ihatemoney.readthedocs.io/en/latest/contributing.html in order to run the server.
It has been impossible to configure it. Finally I have discovered that the errors was that ihatemoney module should be installed before run "python run.py".
I have updated the documentation in order to reflect it.

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -27,7 +27,7 @@ Alternatively, you can also use the `requirements.txt` file to install the
 dependencies yourself. That would be::
 
      pip install -r requirements.txt
-     pip install .
+     pip install -e .
 
 And then run the application::
 
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -27,6 +27,7 @@ Alternatively, you can also use the `requirements.txt` file to install the
 dependencies yourself. That would be::
 
      pip install -r requirements.txt
+     pip install .
 
 And then run the application::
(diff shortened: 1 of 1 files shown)


docs/contributing.rst
```diff
@@ -27,6 +27,7 @@ Alternatively, you can also use the `requirements.txt` file to install the
 dependencies yourself. That would be::
 
      pip install -r requirements.txt
+     pip install .
```
almet: Then, the best way to do this would be to do `pip install -e .`, so that it's editable in place.

JoseAntonioTorre: You are right!
I have modify it in order to reflect your advice.

almet:
Thanks! I've also discovered recently that it's possible to run the flask app in another way, with reload and everything, by doing this;
```
export FLASK_ENV=development
export FLASK_MAIL_DEBUG=1
export FLASK_DEBUG=1
export FLASK_APP=ihatemoney/run.py
flask run
```

So we might want to create a pull request to reflect this at some point. :-)

JoseAntonioTorre:
> ```
> flask run
> ```

Interesting I think that that way was introduced in Flask 1.0. As soon as I start developing in ihatemoney I will update the documentation.

Issue #133, opened 2016-01-31
Split emails on new line, coma, and semi-column and trim them.


Natim:
In the invitation form, it's possible to enter a list of emails. Currently, they are only coma-separated, but they could also be split with new lines and semi-columns.

Also, they could be trimmed.

almet:
That would be a welcome addition!

JoseAntonioTorre:
Could you describe this issue?

almet:
@JoseAntonioTorre I've just updated the description, that should make things a little easier to grasp.

JoseAntonioTorre:
Something like this could be enough? 
![Screenshot_20191015_004113](https://user-images.githubusercontent.com/1927763/66787601-99b15280-eee4-11e9-92cc-d52b207f607d.png)

JoseAntonioTorre:
If that is enough I can send the PR.

Natim:
I guess it can help. Can we copy paste a list of email comma separated as well as tab or new line separated?

JoseAntonioTorre:
Now, I have configured it to accept only commas. In my opinion, having multiples ways to insert the same information is not a good idea in terms of usability. I can extend it to allow entering with ';'.
The PR consists of an input that actuates as normal and when you enter a ',' it will envelope it in a tag.

Natim:
The goal is not to have multiple ways to enter the thing but to handle copy paste from different source.

JoseAntonioTorre:
Ok, I have read the title and I don't see it there.
So besides the functionality that I have already developed, you want to have the option to copy text from a comma-separated list or any of the format you mention before and make it compatible with this system right?

Natim:
We don't need to support all of them, I guess handling coma and newlines would be enough.

JoseAntonioTorre:
Ok, this afternoon I will work on it and if I have time I will send the PR :)

PR #495, opened 2019-10-15
Added support to split invitation emails #133

JoseAntonioTorre:
Support for adding emails as tags.
Now you can paste an email list of comma-separated, tab-separated, semicolon-separated
and the input will render each email as a tag.
Also, you can enter emails in the input and use any of those characters in order to create an email tag. Finally, if you enter an email and click enter it will create a tag.
I have added support to not allow duplicates.
image


ihatemoney/static/css/tagsinput.css
```diff
@@ -0,0 +1,16 @@
+.tagsinput,.tagsinput *{box-sizing:border-box}
```
almet: Can you put these instructions on multiple lines, rather than having each selector on a distinct line?

JoseAntonioTorre: I'm used to be like you and my mind has changed (I don't know if in the correct direction). I think that the functionality that this library (plugin of jquery) implements it is to easy to develop and in fact, I have read all the code in order to make sure that I can implement the behaviour that you told me. The code is straightforward and I think that the dependency should not be a problem (the code can be maintained by ourself).

This afternoon I am going to correct the css and notify you when ready.

JoseAntonioTorre:
@almet Ready!

almet:
Thank you :-)

Issue #246, opened 2017-06-29
Control public project creation via a flag when using the API.


0livd:
Even when disabled, public project creation is still possible via the API.

almet:
It's just a matter of adding a conditional check in the correct `ProjectHandler` that lives at https://github.com/spiral-project/ihatemoney/blob/master/ihatemoney/api.py#L32

Flagging as a good issue to get started.

JoseAntonioTorre:
I would be interested in working on this issue.
What do you mean @almet with checking a ProjectHandler?

almet:
@JoseAntonioTorre the correct link is now https://github.com/spiral-project/ihatemoney/blob/master/ihatemoney/api.py#L77

You just need to add a conditional here. The flag is `ALLOW_PUBLIC_PROJECT_CREATION`, and you can access its state with `current_app.config.get(flag)`.

PR #496, opened 2019-10-18
Fixed public project creation flag in API creation #246

JoseAntonioTorre:
Solved #246 via config check.

### DIFF
--- a/ihatemoney/tests/tests.py
+++ b/ihatemoney/tests/tests.py
@@ -298,7 +298,7 @@ class BudgetTestCase(IhatemoneyTestCase):
             # session is not updated
             self.assertNotIn("raclette", session)
 
-            # project is created
+            # project is not created
             self.assertEqual(len(models.Project.query.all()), 0)
 
     def test_project_creation_with_public_permissions(self):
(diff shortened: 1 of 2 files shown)


almet:
That seems good to me. Can you add a test for this?

JoseAntonioTorre:
> That seems good to me. Can you add a test for this?

Yes, I will try it!

JoseAntonioTorre:
Ready!

ihatemoney/tests/tests.py
```diff
@@ -281,6 +281,46 @@ def test_project_creation(self):
...
+                    "id": "raclette",
+                    "password": "party",
+                    "contact_email": "raclette@notmyidea.org",
+                },
+            )
+
+            # session is not updated
+            self.assertNotIn("raclette", session)
+
+            # project is created
```
almet: I believe the comment should state « not created ».

JoseAntonioTorre: uops, you are right!, I change it and upload the commit.

JoseAntonioTorre: ready

almet:
Thanks!

PR #499, opened 2019-10-18
Change the color of the add a member button.

almet:


almet:
Before
before

After
after

almet:
@eMerzh any input on this? I believe you changed this at some point?

eMerzh:
go for it 👍 
that was not really intentional :)

PR #500, opened 2019-10-19
Added support to generate token of project

JoseAntonioTorre:
Now with endpoint /api/token/<project_id> you will get a {"token": "tokenofproject"} that you can use in GET "/api/project" with header "Authenthication: token <token>".

Hope it helps!

Issue: #412

### DIFF
--- a/ihatemoney/api.py
+++ b/ihatemoney/api.py
@@ -186,8 +186,21 @@ class BillHandler(Resource):
         return "OK", 200
 
 
+class TokenHandler(Resource):
+    method_decorators = [need_auth]
+
+    def get(self, project):
+        print(project)
+        if not project:
+            return "Not Found", 404
+        token = project.generate_token()
+        print(token)
+        return {"token": token}, 200
+
+
 restful_api.add_resource(ProjectsHandler, "/projects")
 restful_api.add_resource(ProjectHandler, "/projects/<string:project_id>")
+restful_api.add_resource(TokenHandler, "/token/<string:project_id>")
 restful_api.add_resource(MembersHandler, "/projects/<string:project_id>/members")
 restful_api.add_resource(
     ProjectStatsHandler, "/projects/<string:project_id>/statistics"
--- a/ihatemoney/api.py
+++ b/ihatemoney/api.py
@@ -1,5 +1,5 @@
 # coding: utf8
-from flask import Blueprint, request
+from flask import Blueprint, request, current_app
 from flask_restful import Resource, Api, abort
 from flask_cors import CORS
 from wtforms.fields.core import BooleanField
@@ -55,7 +55,7 @@ def need_auth(f):
 class ProjectsHandler(Resource):
     def post(self):
         form = ProjectForm(meta={"csrf": False})
-        if form.validate():
+        if form.validate() and current_app.config.get("ALLOW_PUBLIC_PROJECT_CREATION"):
             project = form.save()
             db.session.add(project)
             db.session.commit()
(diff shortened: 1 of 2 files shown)


JoseAntonioTorre:
> Now with endpoint /api/token/<project_id> you will get a {"token": "tokenofproject"} that you can use in GET "/api/project" with header "Authenthication: token ".
> 
> Hope it helps!
> 
> Issue: #412

@almet I have branched from branch fix-control-public-creation instead of master. How can I fix it?

ihatemoney/api.py
```diff
@@ -186,8 +186,21 @@ def delete(self, project, bill_id):
         return "OK", 200
 
 
+class TokenHandler(Resource):
+    method_decorators = [need_auth]
+
+    def get(self, project):
+        print(project)
```
almet: It seems that this print shouldn't be here :-)

ihatemoney/api.py
```diff
@@ -186,8 +186,21 @@ def delete(self, project, bill_id):
...
+        if not project:
+            return "Not Found", 404
+        token = project.generate_token()
+        print(token)
```
almet: Extraneous print statement.

ihatemoney/api.py
```diff
@@ -186,8 +186,21 @@ def delete(self, project, bill_id):
...
+        return {"token": token}, 200
+
+
 restful_api.add_resource(ProjectsHandler, "/projects")
 restful_api.add_resource(ProjectHandler, "/projects/<string:project_id>")
+restful_api.add_resource(TokenHandler, "/token/<string:project_id>")
```
almet: Should we put this at `/projects/:id/token` instead?

ihatemoney/tests/tests.py
```diff
@@ -1317,6 +1357,29 @@ def test_project(self):
...
+
+        decoded_resp = json.loads(resp.data.decode("utf-8"))
+
+        # Access with token
+        resp = self.client.get(
+            "/api/token/raclette",
+            headers={"Authorization": "Basic %s" % decoded_resp["token"]},
+        )
+
+        self.assertEqual(200, resp.status_code)
```
almet: It seems to me that we should also test that the token can be used to login.

ihatemoney/api.py
JoseAntonioTorre: You are right!

ihatemoney/api.py
JoseAntonioTorre: you are right, only for testing

JoseAntonioTorre:
> Here is some feedback:
> 
> * It seems that you also have code from your last pull request into this very one.
> * Some stuff I put in inline comments.
> 
> Thanks!

Thanks for your feedback it is very useful :) 
I will try to change it and I will come back.

I don't know how to modify your fist advice "it seems that you also have code from your last pull request into this very one". I created the branch from a feature-request branch instead of master one. How can I change it?

almet:
> I don't know how to modify your fist advice "it seems that you also have code from your last pull request into this very one". I created the branch from a feature-request branch instead of master one. How can I change it?

I believe the simplest way is to re-create a branch from master (`git checkout master && git pull origin master && git checkout -b mynewbranch`) and then cherry-picking the commits you want from this branch (using `git cherry-pick`).

ihatemoney/tests/tests.py
almet: It should have been, but my review was too weak :(

PR #507, opened 2019-11-01
Feature/#133 pay from settle page with form

LeoMouyna:


LeoMouyna:
@almet I gonna need some help for this build failure :ambulance:.
I have no idea how to make it pass.

almet:
Fixing it should be as easy as running ```black --target-version=py34``` :-)

ihatemoney/forms.py
```diff
@@ -31,14 +31,14 @@ def strip_filter(string):
         return string
 
 
-def get_billform_for(project, set_default=True, **kwargs):
+def get_billform_for(project, set_default=True, bill_type=BillTypes.loan.value, **kwargs):
```
almet: You probably just don't need to add bill_type here, let it go trough the kwargs, and set up a default value in the Form's ``__init__`` method.

ihatemoney/forms.py
```diff
@@ -174,13 +174,19 @@ class BillForm(FlaskForm):
     submit = SubmitField(_("Submit"))
     submit2 = SubmitField(_("Submit and add a new one"))
 
+    def __init__(self, bill_type=BillTypes.loan.value, *args, **kwargs):
```
almet: Especially as it's defined here already.

ihatemoney/forms.py
```diff
@@ -31,14 +31,14 @@ def strip_filter(string):
...
-def get_billform_for(project, set_default=True, **kwargs):
+def get_billform_for(project, set_default=True, bill_type=BillTypes.loan.value, **kwargs):
     """Return an instance of BillForm configured for a particular project.
 
     :set_default: if set to True, on GET methods (usually when we want to
                   display the default form, it will call set_default on it.
 
     """
-    form = BillForm(**kwargs)
+    form = BillForm(bill_type=bill_type, **kwargs)
```
almet: So, remove it from here.

ihatemoney/forms.py
```diff
@@ -174,13 +174,19 @@ class BillForm(FlaskForm):
     submit = SubmitField(_("Submit"))
     submit2 = SubmitField(_("Submit and add a new one"))
 
+    def __init__(self, bill_type=BillTypes.loan.value, *args, **kwargs):
+        super(BillForm, self).__init__(*args, **kwargs)
+        self.bill_type = bill_type
```
almet: To ease the logic, we might want to use strings for the types, and select the correct type here. It will allow to pass a string rather than a python class to the form.

ihatemoney/forms.py
```diff
@@ -174,13 +174,19 @@ class BillForm(FlaskForm):
...
+        self.title = ''
```
almet: Should we set a default title here?

ihatemoney/forms.py
```diff
@@ -191,6 +197,8 @@ def fill(self, bill):
         self.external_link.data = bill.external_link
         self.date.data = bill.date
         self.payed_for.data = [int(ower.id) for ower in bill.owers]
+        if bill.type:
```
almet: What's the case where `bill.type` is set to `None`? (why do you need the `if` statement?)

ihatemoney/models.py
```diff
@@ -12,6 +12,7 @@
     BadSignature,
     SignatureExpired,
 )
+from ihatemoney.utils import BillTypes
```
almet: It seems better to put it in models rather than in utils.

ihatemoney/models.py
```diff
@@ -80,13 +81,17 @@ def members_stats(self):
             {
                 "member": member,
                 "paid": sum(
-                    [bill.amount for bill in self.get_member_bills(member.id).all()]
+                    [
+                        bill.amount
+                        for bill in self.get_member_bills(member.id).all()
+                        if bill.type != BillTypes.settlement.value
```
almet: I think I don't see the value of having a type for this. We probably could use strings instead (or constants)

ihatemoney/forms.py
```diff
@@ -199,6 +207,12 @@ def validate_amount(self, field):
         if field.data == 0:
             raise ValidationError(_("Bills can't be null"))
 
+    def is_loan(self):
```
almet: I believe these should be in the model instead.

almet: (Or maybe I've missed why they're important in the form?)

ihatemoney/templates/forms.html
```diff
@@ -93,8 +93,11 @@
     {{ form.hidden_tag() }}
     {{ input(form.date, class="form-control", inline=True) }}
     {{ input(form.what, inline=True) }}
+    {% if form.is_loan() %}
```
almet: with the change, it would be `if form.type == "loan"`

ihatemoney/tests/tests.py
```diff
@@ -959,6 +973,8 @@ def test_statistics(self):
             response.data.decode("utf-8"),
         )
 
+        # TODO: Add a refund !
```
almet: Why's this todo here?

ihatemoney/tests/tests.py
```diff
@@ -674,6 +675,19 @@ def test_manage_bills(self):
         balance = models.Project.query.get("raclette").balance
         self.assertEqual(set(balance.values()), set([19.0, -19.0]))
 
+        # TODO: Then test refund for balance.
```
almet: Please remove the TODO's

ihatemoney/tests/tests.py
```diff
@@ -1843,6 +1862,7 @@ def test_weighted_bills(self):
             "date": "2011-08-10",
             "id": 1,
             "external_link": "",
+            "type": "loan"
         }
         got = json.loads(req.data.decode("utf-8"))
         self.assertEqual(
```
almet: It seems we're missing tests for the `settle` type.

ihatemoney/utils.py
```diff
@@ -10,6 +10,7 @@
 from babel import Locale
 from werkzeug.routing import HTTPException, RoutingException
 from datetime import datetime, timedelta
+from enum import Enum
```
almet: No need for this :-)

ihatemoney/utils.py
```diff
@@ -234,3 +235,8 @@ def _eval(node):
         raise ValueError("Error evaluating expression: {}".format(expr))
 
     return result
+
+
+class BillTypes(Enum):
+    loan = "loan"
+    settlement = "settlement"
```
almet: No need for this :-)

ihatemoney/web.py
```diff
@@ -640,7 +642,60 @@ def change_lang(lang):
...
+    bill.what = _("Refund")
+    bill.date = datetime.now()
+    bill.owers = [settlement_bill['receiver']]
+
+    form.fill(bill)
+    receiver_name = [user
+                     for (id, user)
+                     in form.payed_for.choices
+                     if id == form.payed_for.data[0]
+                     ][0]
```
almet: This looks quite ugly to me. Isn't there a better way to write this? (in a more readable way)

ihatemoney/web.py
```diff
@@ -640,7 +642,60 @@ def change_lang(lang):
...
+
+    return render_template(
+        "settle_bills.html",
+        bills=bills,
+        bill_forms=bill_forms,
+        current_view="settle_bill"
+    )
+
+
+def get_payment_form(settlement_bill, **kwargs):
```
almet: naming this `bulid_payment_form` would help readability.

almet: You're passing kwargs but not using them.

ihatemoney/web.py
```diff
@@ -640,7 +642,60 @@ def change_lang(lang):
...
+    form = get_billform_for(g.project, set_default=False, bill_type=BillTypes.settlement.value)
+    bill = Bill()
```
almet: You probably can pass all the parameters to the `Bill` init directly.

almet: Or, what about having :+1: 
1. a`from_settlement` class method for the `Bill`, which creates the bill;
2. a function which takes this bill to fill the form.

ihatemoney/web.py
```diff
@@ -640,7 +642,60 @@ def change_lang(lang):
...
+def add_payment(ower, receiver):
+    receiver = Person.query.get(receiver, g.project)
+    if not receiver:
+        raise NotFound("Person with id: {} not found".format(receiver))
+
+    form = get_billform_for(g.project, bill_type=BillTypes.settlement.value)
+    if request.method == "POST":
+        form.payer.data = ower
+        form.payed_for.data = [receiver.id]
+        print(form.payed_for.__dict__)
```
almet: Please, remove this print.

ihatemoney/web.py
```diff
@@ -640,7 +642,60 @@ def change_lang(lang):
...
+        if form.validate():
+
+            bill = Bill()
+            db.session.add(form.save(bill, g.project))
+            print(bill.type)
```
almet: Ditto.

ihatemoney/web.py
```diff
@@ -640,7 +642,60 @@ def change_lang(lang):
...
+            db.session.commit()
+
+            flash(_("The payment has been added"))
```
almet: We probably can be a bit more precise here?

ihatemoney/forms.py
LeoMouyna: It can happens if you have some old bills stored in database. I wanted to avoid compatibility issues according to eneiluj.
> This parameter could be optional in the API (default to "loan") to avoid compatibility problems with old clients versions.

That's why I do a check here. Do you think we should add to the migration script a default value to `loan` for all previously saved bills ?

ihatemoney/templates/forms.html
LeoMouyna: Personally I prefer use a function that contains the check logic but if you think it's a better way to handle this I can change it.
Why do I prefer a function ? What if we change the string in the future ? If you use a function you only change BillTypes.loan value one time, if not you have to check where all `"loan"` occurrences are ...

ihatemoney/tests/tests.py
LeoMouyna: Wow seems like I didn't look at my PR correctly before opening this PR :disappointed:

ihatemoney/forms.py
almet: Migrations ensure the database is valid for a specific schema, so we don't need to do defensive programing. Adding a default value of "loan" here seems the best way to handle this.

We probably want to do it on the model directly :-)

ihatemoney/templates/forms.html
almet: It depends. Here the check is so simple that I greatly prefer to check using strings: it's less indirection, and if we ever change the string in the future, it'll still be really easy to change.

« Premature optimization is the root of all evil ».

ihatemoney/web.py
LeoMouyna: Is it better like this ?
```
receiver_id = settlement_bill['receiver']
users = form.payed_for.choices
receiver_name = [name for (id, name) in users if id == receiver_id][0]
```

almet: If you want to access the users by IDs, then one simple way would be to use a dict (easier to read) :
```python
receiver_id = settlement_bill['receiver']
users = dict(form.payed_for.choices)
receiver_name = users[receiver_id]
```

LeoMouyna:
Hi ! @almet 

Sorry for the delay on this PR but I didn't have time to code on my free time recently.

I hope it will fit all your requirements :wink:

ihatemoney/migrations/versions/52903c4b2459_.py
```diff
@@ -0,0 +1,26 @@
...
+revision = "52903c4b2459"
+down_revision = "6c6fb2b7f229"
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column("bill", sa.Column("type", sa.UnicodeText(), nullable=True))
```
Natim: I don't think we want a nullable there but more a loan by default.

LeoMouyna:
Sorry @almet I pinged you for a review but I don't know who I'm suppose to ping for a review.

almet:
@LeoMouyna It seems that some of the review you've had in the past are still requiring some changes on the code. Are you okay doing them?

ihatemoney/forms.py
almet: This still applies.

ihatemoney/models.py
```diff
@@ -382,13 +394,27 @@ def pay_each(self):
         else:
             return 0
 
+    def is_loan(self):
+        return self.type == BillTypes.loan
```
almet: `== "loan"`

almet:
@LeoMouyna let me know if you need help on this :-)

ihatemoney/models.py
LeoMouyna: I don't get why you don't want to use a variable instead of using a raw string :thinking:

ihatemoney/migrations/versions/52903c4b2459_.py
LeoMouyna: When I look at all the other migration scripts every single new column is nullable except "id". I guess a default value in the model definition should be enough.

ihatemoney/models.py
almet: As I said elsewhere, it's to ease the logic. That's the more "pythonic" way of doing things. It's more idiomatic. It's less useless code floating around :-)

ihatemoney/forms.py
```diff
@@ -182,13 +182,22 @@ class BillForm(FlaskForm):
     submit = SubmitField(_("Submit"))
     submit2 = SubmitField(_("Submit and add a new one"))
 
+    def __init__(self, bill_type=BillTypes.loan, *args, **kwargs):
+        super(BillForm, self).__init__(*args, **kwargs)
+        if bill_type in [BillTypes.loan, BillTypes.settlement]:
+            self.bill_type = bill_type
+        else:
+            self.bill_type = BillTypes.loan
```
almet: You don't need all this logic here. Just passing the default value ("loan" if we use a string) and we'll be good to go.

ihatemoney/models.py
LeoMouyna: But that means if you want to change `"loan"` by `"bill"` for example you will  have to update each `"loan"` occurrence.
Doesn't seem a bad practice to you ?
If it's more "pythonic" I can remove the class `BillTypes` and replace `BillTypes.loan` by `"loan"` and `BillType.settlement` by `"settlement"`.
It doesn't fit the way I'm used to code but if it makes the code more readable why not. 🤷‍♂

almet: If we were to change "loan" by "bill" for instance, then we would just have to find all occurrences and update them, really no big deal.

I believe it would make the code more readable. I can understand why this would be the way to do things in Java for instance, but in python, it's easier to do it this way I believe :-) Let's do it !

LeoMouyna:
Hi there !
As suggested I changed class attributes call by string directly.
I also add API endpoints: `/api/projects/project_id/payments` to generate or list refunds. (refunds are still considered as Bill).
Hope everything is there !

indatwood:
Hi @LeoMouyna, I'm new here so excuse me if it's not the right way you people are working usually! I've checked your pull request, and here is some feedback about it:

- It works, and that's great :-)
- The dollar button you used could be replaced with something easier to understand (like a text !), at first I didn't understood what it was made for (and I figured it out with the tooltip)
- Would it be possible to visually indicate that refunds are refunds in the bills list? (maybe with a star or something, and an indication that when there are stars, it means it's refunds?)

That's it for the user experience.

indatwood:
Hi @LeoMouyna. This is open for quite some time now, so it would be cool if we could get it to a mergeable state. Do you still want to work on this or should we close the issue? Thanks!

LeoMouyna:
I think we can close it 😃. I won't work on it and the devs are probably out dated now.

Glandos:
Too bad, but this will stay as history for future works.

Issue #512, opened 2019-11-25
Add ability to record expenses in different currencies


ArthurG:
Consider the use case when a group of friends travels to a different country. They will likely have some expenses in their home currency (airplane, uber to airport, etc) as well as some expenses in the foreign currency (food, entertainment). 

This would be a really useful feature to add!

skyrl69:
+1 to this feature. Last year we used 9 currencies on our august 
holidays, this year the same.
Having an automatic [xe.com] conversion is one thing for debt balance, 
but keeping both amounts is useful to input & check bills.

Nice to follow you for the last two months, guys ^

Charles+++

Le 25/11/2019 à 11:22, Arthur Guo a écrit :

almet:
Hi, thanks for taking the time to open an issue for this.

I've been thinking a bit and… I'm not sure we want this. It seems at a first glance out of scope for IHM, and will probably create more problems than it would solve : how should we handle the conversion rate ? In which currency should we compute the balances?

One simple way to solve this would be to have a default currency per project, and then have a tool to do the conversion *when we enter the new bill*, so that it converts from one currency to another one. Then everything is stored in one currency.

We could detect the currency quite easily using their text reprentation (EUR, USD, etc), which would make this an advanced use case.

This approach seems a bit overcomplicated for the scope of this very small project : it will tie IHM to another external lib / API, and it means we'll need to maintain it in the long term.

For the record, here are a few libs that might help : 
- https://pypi.org/project/CurrencyConverter/ seems to do the conversion from a database (not sure it works well from day to day changes, in which case we would need to use an API of some sort)
- here is a link to an API : https://www.exchangerate-api.com/docs/python-currency-api
- https://pypi.org/project/currencies/ seems to do the display properly (but I'm not even sure we need this)

If somebody want to take it, I'll be happy to review the code for this. Also, feedback from other users / dev is welcome :-)

ndkoch:
If nobody else has taken this up, I am just beginning to contribute and would like to try and take this on.

almet:
@ndkoch feel free to hack something up! As I said, I'm not sure how this should be done and depending on how it's done, it'll potentially not be merged if it makes things too complicated.

Not to discourage you, so if you feel like taking it, go ahead :-)

ArthurG:
Another way you could do this is to :

- Record expenses in any currency
- At settle time, the user can:
1) See how the settlement would work if they were to pay in the original payment currency
[...]

PR #516, opened 2019-12-09
Close issue #512, Adding the ability to record multiple different expenses

ndkoch:
Added new database schema to include a default currency for every project. Also added two new schema to the bill model to account for the currency of that bill. 

Currently using the ExchangeRate-API recommended by @almet to handle the currency exchange rates.


ihatemoney/web.py
```diff
@@ -277,6 +277,8 @@ def create_project():
             project = form.save()
             db.session.add(project)
             db.session.commit()
+            print(project.name)
+            print(project.default_currency)
```
almet: please remove these.

ihatemoney/forms.py
```diff
@@ -181,6 +199,10 @@ def save(self, bill, project):
         bill.external_link = self.external_link.data
         bill.date = self.date.data
         bill.owers = [Person.query.get(ower, project) for ower in self.payed_for.data]
+        bill.original_currency = self.original_currency.data
+        bill.original_amount = self.currency_helper.exchange_currency(
```
almet: This seems weird to me: it's named `original_amount` but it actually stored the converted amount, right?

almet: If so, we might want to rephrase how this is named. What about `converted_amount`?

requirements.txt
```diff
@@ -25,3 +25,4 @@ pytz==2019.2
 SQLAlchemy==1.3.8
 Werkzeug==0.16.0
 WTForms==2.2.1
+requests==2.22.0
```
almet: you also want to add requests in the `setup.cfg`

almet:
Thanks for you work on this. I've checked-out the branch locally and it seems to work, well done!

I think we're missing a few things here and there in order to have this merged : 

- I believe this is for advanced usage, and as such, should probably be hidden behind a flag. Maybe ask the creator after the project is created, if they want to support multiple currencies, and store the answer.
- If we do this, we might want to have a « NO CURRENCY » currency by default.

Then, it adds some new corner cases : what happens if somebody want to turn their no-currency project in a multiple-currency one? They would need to select the default currency that was used until now somehow, but we don't want to ask them upfront (especially if we don't need it).

I've also added a few comments in line.

ihatemoney/utils.py
```diff
@@ -12,6 +12,31 @@
...
 import csv
+import requests
+
+
+class CurrencyConverter(object):
+    api_url = "https://api.exchangerate-api.com/v4/latest/USD"
+    response = []
+
+    def __init__(self):
+        self.response = requests.get(self.api_url).json()
```
almet: Implementing the conversion this way make it problematic when the service is not available. What happens if we can't call the service? Currently it means that the server will probably return an HTTP 500 error to the user.

One way to solve this would be to have a sort of "state" attached to the records, telling if the conversion has been done or not. If it hasn't been done, then we could try to do it later on (maybe on a subsequent operation). In the meantime, no balance will be computed, and an error message should probably be shown to the end user.

ihatemoney/utils.py
```diff
@@ -12,6 +12,31 @@
...
+
+    def get_currencies(self):
+        currencies = []
+        for rate in self.response["rates"]:
+            currencies.append((rate, rate))
+        return currencies
+
+    def exchange_currency(self, amount, currency1, currency2):
```
almet: please rename to `source_currency` and `destination_currency`.

ihatemoney/utils.py
```diff
@@ -12,6 +12,31 @@
...
+        if currency1 == currency2:
+            return amount
+
+        conversion_rate1 = self.response["rates"][currency1]
+        conversion_rate2 = self.response["rates"][currency2]
+        new_amount = (amount / conversion_rate1) * conversion_rate2
```
almet: ditto : `source_rate`, `destination_rate`

ihatemoney/utils.py
```diff
@@ -12,6 +12,31 @@
 from datetime import datetime, timedelta
 
 import csv
+import requests
+
+
+class CurrencyConverter(object):
+    api_url = "https://api.exchangerate-api.com/v4/latest/USD"
```
almet: Why are we using `USD` as a base here?

ihatemoney/utils.py
almet: should we name this `convert_currency` instead?

ihatemoney/utils.py
```diff
@@ -12,6 +12,31 @@
...
+        for rate in self.response["rates"]:
+            currencies.append((rate, rate))
+        return currencies
+
+    def exchange_currency(self, amount, currency1, currency2):
+        if currency1 == currency2:
+            return amount
+
+        conversion_rate1 = self.response["rates"][currency1]
+        conversion_rate2 = self.response["rates"][currency2]
```
almet: the logic seems weird to me : should we issue a call to the API with the correct currency instead, to just get the currency rate and apply it?

ihatemoney/web.py
```diff
@@ -299,7 +301,7 @@ def create_project():
             )
             try:
                 current_app.mail.send(msg)
-            except SMTPRecipientsRefused:
+            except (SMTPRecipientsRefused, ConnectionRefusedError):
```
almet: Thanks for this.

almet:
Here are some other points I have in mind : 

- Please, add tests.
- How should the API reflect these changes? This could be reflected in the tests also.

almet:
Hi, it's been two months since we last interacted on this pull request. Are you still willing to work on it, or should we close it? Thanks!

ndkoch:
Hello yes, sorry I hve stopped working on it you can go ahead and close it

Glandos:
Sorry to here that. Feel free to reopen it!

almet:
Hey @ndkoch, thanks for taking the time to answer us here. I hope you are doing well. If we did anything wrong on our side, please let us know so we can change that in the future.

If it's due to a lack on time or a lack on interest on your side, then don't worry, it's okay for us :-) Bye!

Issue #417, opened 2019-01-24
Add a way to import previously exported json data


almet:

jeffspurlock:
Hello almet, I'm looking to contribute to my first open source project. It looks like you need something that can parse through the json data from a previous export... is that data that was exported from ihatemoney, or is this json data coming from elsewhere?

almet:
Hi @jeffspurlock, nice that you chosen our project to contribute :-)

It's only data exported from ihatemoney :-)

NVanvyve:
Hi, I'm also interested in this problem.
I looked at the project a little bit and wondered where to integrate the import of json.
Do we want to import the data and create a new project, do we add it to an existing but empty project, or do we import the json and add the data to a project that already contains " bills "?
Basically if I have to create an "import" button, do I have to do it on the homepage or in the project's option menu? :)

almet:
The simplest of all of them: import to an existing project, that should be enough to get started.

waqarmuhammad1:
Hi, if this issue still open i would like to contribute in it.

almet:
@jeffspurlock @waqarmuhammad1 @NVanvyve one of you will need to take it over, I suggest @jeffspurlock does it, if that's okay for you, as you were the first one to propose. What do you think?

almet:
Haha, so in the end, anybody wants to work on this?

NVanvyve:
I'm still ready to work on it, but I thought @jeffspurlock had priority. If nothing moves, I start working

almet:
Awesome, don't hesitate to go for it :-)

NVanvyve:
It took me a little while to finally succeed in importing a JSON and identifying the data to be imported.
I extract the members who are not yet in the project, as well as their weight, I also have a list of the bills that are not in the project.
Now I just have to add his information to the database and that's where I'm blocking it. Is there a proper way to write data to it?
I'm not sure how to use (or create a variant of) the add_member function (web.py) or use the Person class

almet:
That should be quite easy to do with the models directly, so the `Person` class. Something like :

```
from ihatemoney.models import Person, db, Project, Bill
Person(name="Noé", project=g.project, weight=1)

# Do the same for bills, etc, and then save everything with : 
db.session.commit()
```

Hope that helps !

NVanvyve:
It works great for adding Person but impossible to add bills...
I may miss something but I don't see and I've been going around in circles for hours

```python
bill = Bill(
[...]

PR #518, opened 2019-12-17
Import previously exported json data

NVanvyve:
In order to solve the issue #417, I added a tab to load a JSON, as well as a way to add the data found there, in an existing (or empty) project.
I realize that this was not the best way to manage the front end. But I couldn't add the form in the options section. So I did it like that to focus on the rest.
I had to add some messages, I have already noted their translation into French in `fr/LC_MESSAGES/messages.po`


Natim:
This is really appalling, do you mind adding a test for the upload_form?

NVanvyve:
I can do that. 
You mean, test if the file is a correct json?
Because I don't see how else to test the form, especially since I didn't find a test for the other types of forms

Natim:
I meant adding a test to make sure that importing a given json will create the project with the expected members and bills.

NVanvyve:
I added a test as requested and also modified some of what I had done before. I realized that it was possible to search for a user that had been deactivated. At first, I considered that they were two different people

ihatemoney/templates/forms.html
```diff
@@ -89,7 +89,6 @@
     {% include "display_errors.html" %}
     {{ form.hidden_tag() }}
     {{ form.file }}
-    {{ form.csrf_token() }}
```
almet: ?

NVanvyve: I had a little trouble when I had to add a new form because I don't master HTML very well. That's why I created a new tab "Upload", because despite my attempts, I couldn't put it in "Settings".
But then, when I succeeded in creating this form, I had problems reading the JSON. When I searched on the internet, I was advised to add this.
It's afterwards that I realized that it didn't bring anything, so I deleted it. Especially since it was not present anywhere else.

NVanvyve:
Is there anything else I can do?

almet:
Thank you, we just need to have someone review your changes and validate them. Let me summon @Glandos and @Natim on this :-)

ihatemoney/tests/tests.py
```diff
@@ -1190,6 +1190,62 @@ def test_export(self):
...
+                "date": "2016-12-31",
+                "what": "fromage a raclette",
+                "amount": 10.0,
+                "payer_name": "alexis",
+                "payer_weight": 2.0,
+                "owers": ["alexis", "fred", "tata", "pepe"],
+            },
+        ]
+
+        with open("json_test.json", "w+") as file:
```
Natim: I am pretty sure you can use a StringIO there:

```suggestion
    from io import StringIO
        file = StringIO(json.dumps(json_to_import))
```

ihatemoney/tests/tests.py
```diff
@@ -1190,6 +1190,62 @@ def test_export(self):
...
+            json.dump(json_to_import, file)
+
+        self.client.post("/raclette/upload_json", data={"file": "json_test.json"})
```
Natim:
```suggestion
        self.client.post("/raclette/upload_json", data={"file": file})
```

Natim: If this doesn't work, please use [mkstemp](https://docs.python.org/2/library/tempfile.html#tempfile.mkstemp) to make sure to create the file in a temporary folder.

ihatemoney/tests/tests.py
```diff
@@ -1190,6 +1190,62 @@ def test_export(self):
...
+
+        os.remove("json_test.json")
+
+        project = models.Project.query.get("raclette")
+        bills = project.get_pretty_bills()
+
+        for i in json_to_import:
+            for j in bills:
+                if j["what"] == i["what"]:
```
Natim: This doesn't assert that all bills have been loaded properly and that none are missing.

ihatemoney/utils.py
```diff
@@ -234,3 +234,29 @@ def _eval(node):
...
+def same_bill(bill1, bill2):
+    attr = ["what", "payer_name", "payer_weight", "amount", "date", "owers"]
+    for a in attr:
+        if bill1[a] != bill2[a]:
+            return False
+    return True
+
+
+def parse_date(string_date):
+    date = string_date.split("-")
```
Natim: please use [dateutil parser](https://dateutil.readthedocs.io/en/stable/parser.html) for this.

ihatemoney/web.py
```diff
@@ -391,6 +396,83 @@ def edit_project():
...
 
+@main.route("/<project_id>/upload_json", methods=["GET", "POST"])
+def upload_json():
+    form = UploadForm()
+    pid = g.project.id
+    if form.validate_on_submit():
+        filename = pid + "_uploaded_bills.json"
+        form.file.data.save(filename)
+        try:
+            import_project(filename)
```
Natim: There we have the same issue where we create the file locally rather than in memory.

ihatemoney/web.py
```diff
@@ -391,6 +396,83 @@ def edit_project():
...
+        finally:
+            os.remove(filename)
+        return redirect(url_for("main.list_bills"))
+
+    return render_template("upload_json.html", form=form)
+
+
+def import_project(file):
+    # From json : export list of members
+    json_file = json.load(open(file))
```
Natim: if we use a file descriptor rather than a filename here we would fix the previous issue.

ihatemoney/web.py
Glandos: By the way, since we already have the data available, it should be simpler to just parse them as JSON here, handles exceptions and give the JSON dict to `import_project`. Not touching the disk seems a good idea.

ihatemoney/translations/fr/LC_MESSAGES/messages.po
```diff
@@ -539,6 +539,16 @@ msgstr "A dépensé"
 msgid "Balance"
 msgstr "Solde"
 
+msgid "Import"
+msgstr "Importer"
+
+msgid "Project successfully uploaded"
+msgstr "Le projet a correctement été importé"
```
Glandos: Hmmm, mon intuition me guiderait vers « Le projet a été correctement importé ». Je trouve que la formulation actuelle est bancale, mais je n'ai pas de justification…

NVanvyve:
I tried this but `self.assertEqual(len(bills), 3)` fail.
Obviously I can't give a file as an argument.
I think I had the same problem before but like @Natim said I wasn't looking to see if I had the right number of bills.
``` python
        tmpdir = tempfile.mkdtemp()
        predictable_filename = 'file.json'

        # Ensure the file is read/write by the creator only
        saved_umask = os.umask(0o077)

        path = os.path.join(tmpdir, predictable_filename)
        with open(path, "w") as tmp:
            json.dump(json_to_import,tmp)

        self.client.post("/raclette/upload_json", data={"file": path})

        os.remove(path)
        os.umask(saved_umask)
        os.rmdir(tmpdir)

        project = models.Project.query.get("raclette")

        bills = project.get_pretty_bills()

        # Check if all bills has been add
        self.assertEqual(len(bills), 3)
```

ihatemoney/translations/fr/LC_MESSAGES/messages.po
Natim:
```suggestion
msgstr "Le projet a été correctement importé"
```

ihatemoney/web.py
```diff
@@ -391,6 +396,79 @@ def edit_project():
...
+        except ValueError:
+            flash(_("Invalid JSON"), category="error")
+        return redirect(url_for("main.list_bills"))
+
+    return render_template("upload_json.html", form=form)
+
+
+def import_project(file, project):
+    # From json : export list of members
+    json_file = json.loads(file)
```
Glandos: I tried to change this with the following:
```py
json_file = json.load(file)
```
The goal was to avoid double loading from memory, since the main use case is reading data from `upload_json`.
There are other changes then.

ihatemoney/web.py
```diff
@@ -391,6 +396,79 @@ def edit_project():
     )
 
 
+@main.route("/<project_id>/upload_json", methods=["GET", "POST"])
+def upload_json():
+    form = UploadForm()
+    if form.validate_on_submit():
+        file = form.file.data.stream.read()
+        try:
+            import_project(file, g.project)
```
Glandos: Replace this with a simpler:
```py
import_project(form.file.data.stream, g.project)
```
And remove the previous `file` declaration

ihatemoney/tests/tests.py
```diff
@@ -1190,6 +1190,156 @@ def test_export(self):
...
+                "amount": 10.0,
+                "payer_name": "alexis",
+                "payer_weight": 2.0,
+                "owers": ["alexis", "fred", "tata", "pepe"],
+            },
+        ]
+
+        from ihatemoney.web import import_project
+
+        import_project(json.dumps(json_to_import), project)
```
Glandos: This is a test, so we don't care losing some time:
```py
        file = io.StringIO()
        json.dump(json_to_import, file)
        file.seek(0)
        import_project(file, project)
```

ihatemoney/tests/tests.py
```diff
@@ -1190,6 +1190,156 @@ def test_export(self):
...
```
Glandos: Same here as in `test_import_new_project`:
```py
        file = io.StringIO()
        json.dump(json_to_import, file)
        file.seek(0)
        import_project(file, project)
```

ihatemoney/tests/tests.py
```diff
@@ -1190,6 +1190,215 @@ def test_export(self):
...
+    def test_import_wrong_json(self):
+        self.post_project("raclette")
+        self.login("raclette")
+
+        project = models.Project.query.get("raclette")
+
+        json_1 = [
+            {  # wrong keys
+                "checked": False,
+                "dimensions": {"width": 5, "height": 10,},
```
Glandos: I don't know why the last comma in dict fails only in Py3.6 tests… But can you remove it?

And thanks for that new test! :clap:

NVanvyve: Thank you. I was just looking for where the mistake could have come from.

Natim: > why the last comma in dict fails only in Py3.6 tests

It is because flake8 tests are only ran with Python 3.6

PR #520, opened 2019-12-22
Statistics view #323

### LINKED ISSUE #323: Remove sidebar on statistics view
image

It seems to me totally redundant with the central table. What do you think ?

jyleong:
Updates:

Remove header and table column of 'Balance' in statistics view

Edited tests to reflect these changes

Fixes #323

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,7 @@ This document describes changes between each past release.
 
 - Add support for espanol latino america (es_419)
 - Use the external debts lib to solve settlements (#476)
+- Remove balance column in statistics view (#323)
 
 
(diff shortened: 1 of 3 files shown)


Glandos:
It seems to achieve the goal mentioned in #323.

However, I am still not sure it will increase readability in this overview.

jyleong:
What if we moved the heading over to the side bar?
like:

Screen Shot 2019-12-22 at 2 08 45 PM

almet:
@jyleong where are we with this? Should we merge it? Let us know!

jyleong:
Yes we should thanks!

Glandos:
I don't feel 100% sure about this, but I have nothing better in mind.

PR #526, opened 2020-01-10
Monthly expenses

Dutchy-:
When used as a monthly expenses tracker (without the requirement for settling) this tool lacks an overview. This is a WIP initial attempt to add a monthly overview of all expenses.

This is my first version:
image

I do not believe it is ready to merge yet, but I wanted to get some feedback on the location of the new table, the design, should it also contain running/yearly totals, etc. 

Last but not least, is this a feature you even want? :) 
I'm looking forward to your feedback.

Edit: I know I still need to use the translation functions for the strings

### DIFF
--- a/ihatemoney/models.py
+++ b/ihatemoney/models.py
@@ -95,6 +95,18 @@ class Project(db.Model):
         ]
 
     @property
+    def monthly_stats(self):
+        """Compute expenses by month
+
+        :return: a dict of years mapping to a dict of months mapping to the amount
+        :rtype dict:
+        """
+        monthly = defaultdict(lambda: defaultdict(float))
+        for bill in self.get_bills().all():
+            monthly[bill.date.year][bill.date.month] += bill.amount
+        return monthly
+
+    @property
     def uses_weights(self):
         return len([i for i in self.members if i.weight != 1]) > 0
 
--- a/ihatemoney/static/css/main.css
+++ b/ihatemoney/static/css/main.css
@@ -311,8 +311,9 @@ footer .footer-left {
   background: url("../images/see.png") no-repeat right;
 }
 
-#bill_table {
+#bill_table, #monthly_stats {
   margin-top: 30px;
+  margin-bottom: 30px;
 }
 
 .project-actions > .delete,
@@ -486,4 +487,10 @@ footer .icon svg {
 }
 .icon.icon-white {
   fill: white;
+}
+
+/* align the first column */
+#monthly_stats tr *:first-child {
+  text-align: right;
+  width: 200px;
 }
--- a/ihatemoney/web.py
+++ b/ihatemoney/web.py
@@ -55,6 +55,8 @@ from ihatemoney.utils import (
     get_members,
     same_bill,
 )
+from datetime import datetime
+from dateutil.relativedelta import relativedelta
 
 main = Blueprint("main", __name__)
 
@@ -737,9 +739,12 @@ def settle_bill():
 @main.route("/<project_id>/statistics")
 def statistics():
     """Compute what each member has paid and spent and display it"""
+    today = datetime.now()
     return render_template(
         "statistics.html",
         members_stats=g.project.members_stats,
+        monthly_stats=g.project.monthly_stats,
+        months=[today - relativedelta(months=i) for i in range(12)],
         current_view="statistics",
     )
 
--- a/ihatemoney/templates/statistics.html
+++ b/ihatemoney/templates/statistics.html
@@ -17,6 +17,7 @@
 
 
 {% block content %}
+    <h2>{{ _("Balance") }}</h2>
     <table id="bill_table" class="split_bills table table-striped">
         <thead><tr><th>{{ _("Who?") }}</th><th>{{ _("Paid") }}</th><th>{{ _("Spent") }}</th><th>{{ _("Balance") }}</th></tr></thead>
     <tbody>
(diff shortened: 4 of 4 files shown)


Glandos:
To me, monthly expenses are not a goal.

However, more statistics is a nice feature :) Your example is very basic, but remember that some people would like to gather them differently, e.g. quarter (3 months), yearly, or even daily if you are using it on holidays.
So for example, the second column is useless. The period should go in the first (e.g. 2019-01 in your first work), and a control to change the gathering period. As a start, a simple select should be fine to me.

Dutchy-:
I've updated it to look like this
image
I've changed the display so it can be used for different period aggregations later.

I do not have time at this moment to implement a select box for different period aggregations, would you consider merging it as is, so different periods can be added later?

Glandos:
It's simple, nice, clear, and extendable. To me, it's OK to merge this.
@Natim Can you confirm? It's a small review :)

By the way, just a small question: is it OK to have a fixed width in pixels for the 1st column? I know the whole page isn't completely responsive, but it is not a good reason :)

ihatemoney/models.py
```diff
@@ -94,6 +94,17 @@ def members_stats(self):
...
         ]
 
+    @property
+    def monthly_stats(self):
+        """Compute expenses by month
+        """
+        monthly = defaultdict(lambda: defaultdict(int))
+
+        for member in self.active_members:
+            for bill in self.get_member_bills(member.id).all():
```
Natim: Is it possible to rewrite this so that we do a unique SQL request rather than one per members?

almet:
Hi, thanks for the work on this. If we are to have a « statistics » view like this one, I would rather have a graph rather than have a table like is proposed here, what do you think?

Glandos:
Both table and chart are valid to me, with different use cases. Sometimes, you want a quick look at raw numbers, or just copy/paste them. Sometimes, you need to compare months with each other, and charts are better.

However, charting usually requires external JS libraries, and it usually increases the page loading.

ihatemoney/models.py
Dutchy-: Probably; I will change it.

Dutchy-:
> By the way, just a small question: is it OK to have a fixed width in pixels for the 1st column? I know the whole page isn't completely responsive, but it is not a good reason :)

I think it's ok but I haven't checked on a mobile device or anything. Since I have to change something anyway I will doublecheck.

> Hi, thanks for the work on this. If we are to have a « statistics » view like this one, I would rather have a graph rather than have a table like is proposed here, what do you think?

In my use case I'd rather have a table. An additional chart does not hurt but I'd say that can stay for future work :)

ihatemoney/models.py
Dutchy-: I've changed this - as it turned out, a query for bills by active members was simply wrong for the historical statistics. I now just use `get_bills()`.

Dutchy-:
I've updated the query.
Here's a screenshot of a mobile view:

image

Let me know if you have any further feedback. It's all done for me.

ihatemoney/models.py
```diff
@@ -94,6 +94,18 @@ def members_stats(self):
...
         ]
 
+    @property
+    def monthly_stats(self):
+        """Compute expenses by month
+
+        :return: a dict of years mapping to a dict of months mapping to the amount
+        :rtype dict:
+        """
+        monthly = defaultdict(lambda: defaultdict(int))
```
almet: amount seems to be [a float](https://github.com/spiral-project/ihatemoney/blob/1a657ae67a77b11e1126a07cb470b563b29788b1/ihatemoney/models.py#L367) rather than an int.

ihatemoney/static/css/main.css
```diff
@@ -311,8 +311,9 @@ footer .footer-left {
   background: url("../images/see.png") no-repeat right;
 }
 
-#bill_table {
+#bill_table, #month_table {
```
almet: I believe we can improve a bit the naming of this table. what about "monthly_expenses_table" ?

almet: or `monthly_stats` as you named it like this elsewhere (I find the naming even better than my proposal)

ihatemoney/web.py
```diff
@@ -737,9 +739,12 @@ def settle_bill():
 @main.route("/<project_id>/statistics")
 def statistics():
     """Compute what each member has paid and spent and display it"""
+    today = datetime.now()
     return render_template(
         "statistics.html",
         members_stats=g.project.members_stats,
+        monthly_stats=g.project.monthly_stats,
+        months=[today - relativedelta(months=i) for i in range(12)],
```
almet: that's smart ;-)

ihatemoney/models.py
Glandos: It doesn't really matter, since it is only used for setting the initial value, which is 0.
And, to be exact, it should be a `Decimal`, but that will be for the future :)

Dutchy-: You are totally correct, it's better not to confuse types. Changed it to a float :)

ihatemoney/static/css/main.css
Dutchy-: Changed.

Dutchy-:
> Sorry for the delay on all this, it should be easier for contributors to push code around here. Thanks again for you contribution.

I don't feel like I've had to wait much :) There were longer pauses between my own changes. You guys are doing great!

almet:
This looks good to me! Thanks :-)

PR #530, opened 2020-02-08
Very Minor Changes

DavidRThrashJr:
* Made the forms template more readable
* Removed FIXME within edit_bill function, no bill will be returned if the bill_id is not associated with that particular project
* Resolved FIXME within the pay_each function, now using a sql query for sum rather than for loop

### DIFF
--- a/ihatemoney/web.py
+++ b/ihatemoney/web.py
@@ -699,6 +699,7 @@ def delete_bill(bill_id):
 
 @main.route("/<project_id>/edit/<int:bill_id>", methods=["GET", "POST"])
 def edit_bill(bill_id):
+    # FIXME: Test this bill belongs to this project !
     bill = Bill.query.get(g.project, bill_id)
     if not bill:
         raise NotFound()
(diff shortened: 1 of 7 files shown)


ihatemoney/web.py
```diff
@@ -699,7 +699,6 @@ def delete_bill(bill_id):
 
 @main.route("/<project_id>/edit/<int:bill_id>", methods=["GET", "POST"])
 def edit_bill(bill_id):
-    # FIXME: Test this bill belongs to this project !
```
Glandos: How did you test this?
Looking at the code, it seems correct, but I don't see a test for this either :)

Glandos:
Hi @DavidRThrashJr and thanks for your support to this project :+1: 

We would like to avoid mixing Pull Requests with different goals. In this case, I see 3:
- Formatting template
- Use an optimized SQL query
- Multiple API version

All of them are useful, and appreciated, but could you open 3 different requests? It is much easier to discuss and improve them separately.

DavidRThrashJr:
will open 3 different oull requests

PR #532, opened 2020-02-09
Update models: Bill.pay_each()

DavidRThrashJr:
SQL query is faster for summing up weights

### DIFF
--- a/ihatemoney/models.py
+++ b/ihatemoney/models.py
@@ -18,6 +18,9 @@ db = SQLAlchemy()
 
 
 class Project(db.Model):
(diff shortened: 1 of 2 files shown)


Glandos:
It looks good to me. @Natim or @almet, can you have a quick look?

almet:
That looks good to me, thanks! We currently don't have any unit test for this, but I believe that would actually be needed.

Are you willing to create a unit test testing this scenario? Thanks!

Glandos:
The test looks good to me. Thanks for it!

almet:
Thanks :-)

PR #536, opened 2020-02-23
Misc small updates

am97:


Glandos:
I merged this because it concerns only typo.

Except for nightly, but it was undocumented before, so I hope that no-one was using `NIGHTLY=1` in a critical environment.

PR #537, opened 2020-02-25
Feature/#112 dashboard sorting

### LINKED ISSUE #112: Add a way to sort projects in the dashboard
Currently, when you access /dashboard you have some stats but that's not possible to sort them so they're pretty useless.

bmatticus:
Added DataTables for sorting, pagination and filtering of bill_table on dashboard. fix #112


almet:
Hi, and thanks for this.

I believe this feature need also to have some changes on the python part, as we're now supporting pagination, so if we sort, we need to send a new query to the backend.

bmatticus:
Ah, I can disable the pagination in the DataTables overlay, I didn't notice any pagination from python when I looked. I'll disable that pagination for the moment, I can look further at the backend pagination to see if there needs to be other updates.

almet:
Yeah, it's been added quite recently in https://github.com/spiral-project/ihatemoney/commit/9378694034d6e9040548b9e65ea65769fb8272b7

Neat, let us know if you find any issue down the road ;-)

bmatticus:
Dug a little bit here, this is specifically for the project view itself, at 100 bills it'll paginate the view. I can also apply this sorting for this view using DataTables if that is desired, should be pretty easy to add it. I could look into adding the SQLAlchemy pagination to the dashboard as well, if it makes sense. Maybe plumb in a way to adjust the items per page as well.

bmatticus:
I'll add a little here as well, sorry for the spam.

There could be a couple of approaches for the sorting here. For the dashboard specifically the projects are pulled in boundless. In this case there is a load on the DB to pull all rows and potentially the browser/client for memory to store the data. The plus side here is less complexity in passing data back and forth with Flask/SQLAlchemy to drive sorting, and it allows the adjustment of number of rows visible and searching inside that dataset.

The update on the list_bills template is leveraging SQLAlchemy to do pagination which provides the advantage of less load on the DB, potentially but it still has to restrict rows/etc, and less memory footprint on the client. However, there is complexity in driving the data to drive the pages and if we want to sort we need to re-run queries against the DB instead of allowing the client to fetch from memory and perform inside of JS.

I thought this was for the dashboard specifically, but maybe I need some further clarification on the target(s) for this issue. 

I think I'll cancel/close this PR out for now.

almet:
> I thought this was for the dashboard specifically, but maybe I need some further clarification on the target(s) for this issue.

You're actually correct : I was mistaken, if this is for the dashboard only, then your pull request seems just fine. I though it was for the bills themselves.

PR #538, opened 2020-02-25
Bmatticus/feature/112 dashboard sorting

bmatticus:
Added DataTables to Dashboard for sorting, pagination and filtering. Also made project name a link into the list_bills API to allow users to drill in directly to project to do other things from the dashboard. Feature #112


almet:
That looks good to me. Any other feedback?

Glandos:
Very nice improvement.

I have one remark: is it possible to load CSS and JS only from dashboard? Right now, it will be loaded from every page, even if it will only be used for administration purposes.
It currently add ~95KB (uncompressed) on the main page, for a total of 529KB, which is something to be considered (at least to me :yum: )

bmatticus:
> Very nice improvement.
> 
> I have one remark: is it possible to load CSS and JS only from dashboard? Right now, it will be loaded from every page, even if it will only be used for administration purposes.
> It currently add ~95KB (uncompressed) on the main page, for a total of 529KB, which is something to be considered (at least to me 😋 )

Yep, I can move it over, I was going to put it in the dashboard directly but most JS was being loade din layout and I was trying to be consistent :)

bmatticus:
One other thing I was considering, should I remove the button/link on the project name and add a eye icon under actions to view the project? I think that may be more in line with the general feel.

Glandos:
I am surprised that it works… I thought `<link>` elements should be included in the `<meta>` tag… But it seems that the browser find it anyway. Do we have a better solution for now?

Glandos:
> One other thing I was considering, should I remove the button/link on the project name and add a eye icon under actions to view the project? I think that may be more in line with the general feel.

This is a good idea. The button seems a bit big in this table. Maybe keep a simple link on the project name, and add an eye button. The eye button will guide the user to the available action, and the link is just a shortcut for hurry people.

bmatticus:
> I am surprised that it works… I thought `<link>` elements should be included in the `<meta>` tag… But it seems that the browser find it anyway. Do we have a better solution for now?

Yea, it should be in the <head> tag at least, it works either way but it is not proper. I used an if in the template to load only if the target is the dashboard for now. Also put the eye icon in place.

Glandos:
All good :) Since @almet already approved this, let's enter the merge process.

Thank you for your work on this, it's really appreciated.

PR #539, opened 2020-02-26
Changed button label for the Member-edit-Form

jkopka:


LeoMouyna:
@jkopka you should not close your own pull request. You have to wait for a review and then someone gonna accept the PR and merge your changes into the master branch ;)

jkopka:
@LeoMouyna Ok, i've reopened it. Thanks.

Glandos:
Don't you think it will be better to trust the Python class `MemberForm` that defined the [`submit` button](https://github.com/spiral-project/ihatemoney/blob/eea50b9b2c6963e7b0f5cdb3fda3684c51958533/ihatemoney/forms.py#L225)?
So, instead of redefining a button here, and ignoring the one from `MemberForm`, just change the current `"Add"` at the current line with a new i18n key.
I have to admit that I needed to test things to understand this :)

By the way, it may be good to use another key than `"edit"`, because it will use a lowercase, and button action should be capitalized.

If you need help on this, please ask :)

LeoMouyna:
@Glandos I really appreciate your suggestion but that means we need to update several forms template to match with your suggestion. It's probably out of this issue scope 🙊 .

Glandos:
Maybe I missed something, but the `submit` class field of `MemberForm` class is only used by the macro `edit_member` in `forms.html`. This macro is only referenced in `edit_member.html`, called by the route `/<project_id>/members/<member_id>/edit`.
So with this PR, the field `submit` could be safely removed, if I understood things correctly.

Glandos:
@jkopka Is there any reason for closing this?

jkopka:
@Glandos Sorry, i'm new to this. Daniel Atwood wrote "If the issue is fixed, then you should close it".

What's the next steps here? How will the branch get merged with the master?

Glandos:
The UI of GitHub doesn't make it clear but you have 2 different things:
- Pull Request aka PR
- Issue

The UI is the same for the conversation, but usually, you open an issue, then a pull request that fix the issue, and close both in the end when the pull request is merged. Sometime, the "open the issue" step is skipped, and you just have to deal with the pull request. Here, we are in the pull request, that is currently not merged. I've written a comment, but @LeoMouyna seems to disagree, and I was currently waiting for his answer on my answer.
Sometimes, it takes time for project like this to go forward :smiley: 

But I'll take a decision this week anyway!

jkopka:
Thank you for your explanation. That helped me. I've closed the issue.

Natim:
I am wondering if we should take the opportunity of this pull-request to plug a service like lingohub for this project.

eMerzh:
@Natim i think weblate has integrations already ...
we should address this in a distinct issue

indatwood:
Hi. I'm not sure what we should do about this?

Glandos:
I'm closing this, it will be too hard to merge, given the conflicts.

PR #541, opened 2020-03-10
Feature/currencies

dark0dave:
For issue: #512  and update from https://github.com/spiral-project/ihatemoney/pull/516


ihatemoney/utils.py
```diff
@@ -14,6 +14,32 @@
...
+    def __init__(self):
+        self.response = requests.get(self.api_url).json()
+
+    def get_currencies(self):
+        currencies = []
+        for rate in self.response["rates"]:
+            currencies.append((rate, rate))
+        return currencies
+
+    def exchange_currency(self, amount, source_currency, dest_currency):
```
almet: Thanks for taking this back to life :-) It would be great to cache the rates on disk, rather than doing the request all the time. It would also allow to save us in case the service is down. Let's say we'll cache the rates once a day per currency, if that's okay for you!

dark0dave: Hey, sounds great, I'll find a python package for it!

dark0dave: @almet please have a look now. I added a on disk cache.

almet: That looks great :-) I'm just a bit reluctant to have all this in utils.py now. What about creating a `currency_convertor.py` file and putting all the logic in it, as well as instantiating the convertor object in the setup of the application, passing it the cache as an argument, for instance ?

dark0dave: @almet happy to change that. Give me some more time.

dark0dave: @almet please have another look.

ihatemoney/tests/tests.py
```diff
@@ -408,6 +415,8 @@ def test_membership(self):
                 "payer": fred_id,
                 "payed_for": [fred_id],
                 "amount": "25",
+                "converted_amount": "25",
+                "original_currency": "USD",
```
almet: I don't understand why, when creating a new bill, we need to define both amount and converted_amount?

almet:
Hey, it looks great! I'm sorry to add yet another thing before we can merge but : 

- i don't see any tests implemented to check that things are working as they should ;
- I would like to be sure that the current API doesn't break, and that it's still possible to sync with for instance MoneyBuster ;
- Do we really need the client to specify the converted amount when creating a new bill ? I don't understand why this is the case.
- I know that we would like to get this packaged for Debian, and I wonder what the policy is when it comes to the use of external services such as the one we're using here. Do you (or anybody) have any idea if that would be okay? cc @indatwood

ihatemoney/tests/tests.py
dark0dave: I fix it so it defaults to USD

dark0dave:
- Will add some tests for currency converter
- Not sure what you after here. Can you give me an old api request?
- I am going to fix this.
- Is this needed for this pr?

dark0dave:
> I would like to be sure that the current API doesn't break, and that it's still possible to sync with for instance MoneyBuster ;
I will make the api backwards compat

almet:
> Will add some tests for currency converter

Great, thanks!

> I will make the api backwards compat, basically assumes USD from the past.

Not sure we should assume USD from the past, maybe we can have an "undefined" currency, which will allow to have a backward compat?

About Debian packaging and using external services, I would rather fix this sooner rather than later, let's see if we can get any inputs from Dan :-)

ihatemoney/currency_convertor.py
```diff
@@ -0,0 +1,41 @@
...
+    api_url = "https://api.exchangerate-api.com/v4/latest/USD"
+    _internal_cache = Cache()
+
+    def __init__(self, expire=86400):
+        self.expire = expire
+
+    def set_rate(self):
+        if "rates" not in self._internal_cache:
+            self._internal_cache.set(
+                "rates", requests.get(self.api_url).json()["rates"], expire=self.expire
```
Glandos: The `parse_float` argument from [`json.loads`](https://docs.python.org/3/library/json.html#json.load) is even given in the examples.

ihatemoney/currency_convertor.py
```diff
@@ -0,0 +1,41 @@
...
+        return [(rate, rate) for rate in self._internal_cache.get("rates")]
+
+    def exchange_currency(self, amount, source_currency, dest_currency):
+        if source_currency == dest_currency:
+            return amount
+
+        self.set_rate()
+        source_rate = self._internal_cache.get("rates")[source_currency]
+        dest_rate = self._internal_cache.get("rates")[dest_currency]
+        new_amount = (amount / source_rate) * dest_rate
```
Glandos: Of course, since for now the amount is in float, we will output the result in float, but we need to start somewhere.

indatwood:
Hey, thanks for poking me here. I'm not sure about the Debian policy for these cases, but I did some research, and nothing seems to forbid the usage of a specific API.

That being said, I would rather rely on an open source API if possible, just because it will be easier to maintain in case the service closes its doors. https://exchangeratesapi.io/ (not the same one you're using, despite the close name) seems to be a good candidate, but I don't think that changing is necessary : if it works, it works :-)

ihatemoney/currency_convertor.py
indatwood: While looking for python packages that might already be doing this, I stumbled upon https://github.com/limist/py-moneyed which seems to provide tooling for not doing weird operations between different currencies. Might be worth looking at?

dark0dave: I am happy to add that, perhaps as a separate pull request. As this one is quite large, and I would also have to change the previous models. Which seems like a separate feature, where we upgrade all the amounts to decimal. As that is the only format the models would support natively.

dark0dave:
@indatwood changed to  https://exchangeratesapi.io/ , seems to be a very similar api call. And I am happier using an open source api.

ihatemoney/forms.py
```diff
@@ -171,6 +184,13 @@ class BillForm(FlaskForm):
     what = StringField(_("What?"), validators=[DataRequired()])
     payer = SelectField(_("Payer"), validators=[DataRequired()], coerce=int)
     amount = CalculatorStringField(_("Amount paid"), validators=[DataRequired()])
+    currency_helper = CurrencyConverter()
+    original_currency = SelectField(
+        _("Currency"),
+        choices=currency_helper.get_currencies(),
+        validators=[Optional()],
+        default="USD",
```
Glandos: … and the default value should be the project one here :)

ihatemoney/forms.py
```diff
@@ -89,6 +94,12 @@ class EditProjectForm(FlaskForm):
     name = StringField(_("Project name"), validators=[DataRequired()])
     password = StringField(_("Private code"), validators=[DataRequired()])
     contact_email = StringField(_("Email"), validators=[DataRequired(), Email()])
+    currency_helper = CurrencyConverter()
+    default_currency = SelectField(
+        _("Default Currency"),
+        choices=currency_helper.get_currencies(),
```
Glandos: This should be sorted (see below)

Glandos:
I have also another request: it should be nice to use currency symbol if available, displayed as:
 - € (EUR)
 - $ (USD)
 - £ (GBP)

And sorted according to the 3-letter code. Of course, this could be done in another enhancement request, to avoid the endless run for the perfection :wink:

ihatemoney/forms.py
```diff
@@ -171,6 +184,13 @@ class BillForm(FlaskForm):
     what = StringField(_("What?"), validators=[DataRequired()])
     payer = SelectField(_("Payer"), validators=[DataRequired()], coerce=int)
     amount = CalculatorStringField(_("Amount paid"), validators=[DataRequired()])
+    currency_helper = CurrencyConverter()
+    original_currency = SelectField(
+        _("Currency"),
+        choices=currency_helper.get_currencies(),
```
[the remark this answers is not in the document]
dark0dave: ok, will sort lexicographically.

ihatemoney/forms.py
dark0dave: done, but in a slightly different way.

dark0dave:
> I had some UX issues. The majority of this PR is really nice, but selecting a currency should be easier.

These should be fixed. Please have another look.

I agree, lets land this then add the symbols.

ihatemoney/forms.py
dark0dave: done.

Glandos:
With already existing project, `None` is written everywhere. Maybe it should be replaced with an empty string for the currency, and the `Amount in …` should be written without conversion.
image

dark0dave:
@Glandos, I updated the migrations so now its not none:
image

dark0dave:
@Glandos I also updated, the change currency in the project tab. When some one changes the default currency all bills are now updated, with new converted amounts. This prevents users from doing this manually.

almet:
Nice :-) "Amout in" should be just "amount" in case there is no specified default currency, right?

dark0dave:
@almet done. Now just displays amount. If there are no more issues can we land this?

ihatemoney/tests/tests.py
```diff
@@ -2224,5 +2252,24 @@ def test_bill_pay_each(self):
...
+    converter._internal_cache.set("rates", {"USD": 1, "EUR": 0.81149})
+
+    def test_only_one_instance(self):
+        one = id(CurrencyConverter())
+        two = id(CurrencyConverter())
+        self.assertEqual(one, two)
+
+    def test_get_currencies(self):
+        self.assertCountEqual(
+            self.converter.get_currencies(), [("USD", "USD"), ("EUR", "EUR")]
```
almet: I don't understand what get_currencies is supposed to return? (and why it returns this!)

almet:
I've just a question about `get_currencies` but otherwise I think we're good from the code point of view.

Still need to try it on my machine and then let's land this :-)

ihatemoney/tests/tests.py
dark0dave: @almet , get_currencies returns an array tuple pair of the currencies. It is used for currency selection in form.py and whenever a user selects a currency. It is the map from the choice to the value:

Ie I am shown "USD" and the application registers my currency as  "USD".

image
image

almet: Alright! Thanks for the explanation. That's why the currencies appear two times. I find it kinda weird to tie the logic of the forms to the CurrencyConverter.

Due to the name, I was expecting `get_currencies` to return a list of currencies (maybe with their representation with the dollar sign etc) !

Maybe should we not tie the two together, and do the adaptation for the forms in the forms module instead (`choices = [(c, c) for c in currency_helper.get_currencies()]`) ? What do you think?

dark0dave: I mean, that's kinda violating DRY, but alright. At this point I just want to land this. So I am happy to change that.

dark0dave: Done. Please let me know there is anything else.

dark0dave:
@almet have you had a chance to try this? Its been a while.

indatwood:
Seems @almet is not as available as he used to. I'll try to help here and check this out locally!

Some feedback:

When testing it locally, `diskcache` was missing from `requirements.txt` so I had to install it manually;

The tests are now dependent on an external service, and I'm not sure what to think about it. In other projects we used to mock the calls to the external API to avoid this in unit tests, and also had functional tests to ensure that everything was working properly. Maybe it's time to add functional tests to this project? Not sure.

Migrations didn't seem to be run automatically for me. I got a `sqlite3.OperationalError: no such table: project` error. It might just be me and the way I ran things, but I expected the migrations to be run automatically for me. I couldn't find how to run them [in the docs](https://ihatemoney.readthedocs.io/en/latest/installation.html) so it was a show stopper for me :/

indatwood:
> Migrations didn't seem to be run automatically for me. I got a sqlite3.OperationalError: no such table: project error. It might just be me and the way I ran things, but I expected the migrations to be run automatically for me. I couldn't find how to run them in the docs so it was a show stopper for me :/

Found my way around it, sorry for the noise (seems it was due to the fact I needed to reset my database).

Here's the rest of the feedback!

- I tried on the test project first, and I wasn't expecting multiple currencies to show by default. I think we shouldn't display them by default, and only display them if a "use mulitple currencies" switch is present in the project settings, as it will let things be simpler (otherwise I think there is too many optional inputs);

- Default currency was by default to "USD" (for the test project) which is weird for me, because I leave in the euro zone. I don't think we should decide for the user, and having a default "no currency" is better than defaulting to USD.

Then I moved to creating a new project.

- Again, I find it weird to select the default currency when I create the project, as it's now an extra step. I can understand why it's useful to some people, but I've been using this project for quite some time and for me the question of the currency never was one (and I used it in different countries). So I believe we should really make this optional. What about a checkbox here "I want to use multiple currencies", which would then allow to select which default currency you want to use?

I created the project and then created a few members, and added a new bill : 

- When entering the new bill, I'm being asked which currency it should be in. What about a link "I want to enter an amount in a different currency" here, which would make things simpler for the general user ?

I then entered the amount in USD, it worked well !

- In the list of bills, `150.00 USD (69.54 EUR each)` appeared, which I think is weird. We should choose between EUR and USD here to be consistent.

That's it for the user experience feedback :-) My recommendation boils down to hiding all this behind a user choice! Good job otherwise, of course!

dark0dave:
Hey I appreciate the feedback, I will do my best to answer your questions.

> When entering the new bill, I'm being asked which currency it should be in. What about a link "I want to enter an amount in a different currency" here, which would make things simpler for the general user ?

I think this is fine. It starts on the default currency for that project anyways, so I am not sure that this much of an improvement.

> Default currency was by default to "USD" (for the test project) which is weird for me, because I leave in the euro zone. I don't think we should decide for the user, and having a default "no currency" is better than defaulting to USD.

I have changed to blank, it will cause issues further down the road. We either have some kind of sane default or we can't settle very easily. I happen to live in the euro zone too, but I still just picked USD. I could have equally pick EURO, but USD is more commonly used around the [world](https://en.wikipedia.org/wiki/Template:Most_traded_currencies). 

> In the list of bills, 150.00 USD (69.54 EUR each) appeared, which I think is weird. We should choose between EUR and USD here to be consistent.

^^ This issues has been fixed so its now:
150.00 USD (69.54 USD each)
or 
150.00 EUR (69.54 EUR each)
depending on the the currency of the bill

>Again, I find it weird to select the default currency when I create the project, as it's now an extra step. I can understand why it's useful to some people, but I've been using this project for quite some time and for me the question of the currency never was one (and I used it in different countries). So I believe we should really make this optional. What about a checkbox here "I want to use multiple currencies", which would then allow to select which default currency you want to use?

I am sure you can appreciate this is a lot of work to implement, not to mention code, where I am constantly checking if currencies are enabled. I don't wish to implement a situation where I am effectively supporting two very different tools.

indatwood:
I believe it all boils down to not having a more complex UX than it is today for people who use the tool now (without the multiple currencies), that's why I'm proposing to have all this behind a kind of flag

I can understand why this is frustrating for you: you've been working on this quite a lot, and had different interlocutors, asking for different things all along the way. Sorry to add to this, but the focus on this project had been to have a "Keep It Simple, Stupid" tool, and that's why I like it, hence my feedback!

From the user stand point, I really think it could be better to hide all this, rather than asking the user to make choices that they don't understand / find valuable. This is especially true for the operations the users are doing when they first use the software (creating a project) and the one they use the most (adding a bill).

I would be happy to have feedback from an user experience specialist (I believe @QuentinRoy is?) about this, and maybe prove me wrong :-)

From the code perspective, I believe it's not too much work (but still, it's work!) to hide all of this. Just a quick note to also tell that having a "No currency" currency might help to keep with the default behaviour, and ease while developing this !

Also, I'm just a reviewer here, and my opinion is just mine, and doesn't reflect the opinion of the project owners (@Glandos and @Natim I think), so they would be the one to have the final word here!

dark0dave:
@indatwood are you able to merge this pr?

indatwood:
> @indatwood are you able to merge this pr?

I'm not :-) I'm just a reviewer, trying to push things in the right direction :-)

QuentinRoy:
I am not convinced we need a default currency from the start either. I have not looked at the UI, but if its main use is when users settle, wouldn't it make more sense to select it then? At this point, you may even be able to infer a reasonable default by looking at the most used currency in the project.
A project settings could do it, but it also adds complexity.. I do not know if it is the best idea. And how do you enable this setting? Do we add a project-wide setting page, or a checkbox at project creation (in that case, it seems almost as fast to select a currency). Essentially, the functionality we are talking about adds a select widget at bill creation, and another when settling in (and at this point we already have a fairly good idea of what makes more sense), isn't it? It does not sound too bad, but again I haven't had a look at the UI for a long time.. Ideally, we'd do some user testing and compare alternatives, but we don't have that kind of resources and user testing is... compromised at the moment.
It would be a good idea to be able to find a reasonable default also at bill creation in order to avoid issues similar to what @indatwood was talking about. I think we should be able to find one from the locale settings of the browser, isn't it?

Glandos:
@dark0dave as a start let me be clear: we want to merge this, because it fits a user need, and you are not alone.

But I agree with @indatwood the UI/UX should be as simple as possible for first time users. And displaying multiple columns with or without currency makes is more complex.

Your discussion is constructive on this, let me propose a compromise:
1. Add a currency "No currency" in the dropdown list, selected by default. It will serves as a setting. When selected, all currency conversion in the project are disabled, and the UI keeps being simple.
2. If later on, a default currency is selected, then all bills without currency should be converted to this currency.
2. If a project currency is switched to "no currency", do nothing. New bills will have no currency, but the UI will be simpler. We can add a confirmation if the project already have bills with mixed currencies, but I consider it optional for now.
3. Keep the dropdown as-is on the home page for now. We could have an "Advanced" section, hidden by default, but I don't think it's useful for now.

I guess that adding a "no currency" is a little more work, but 2. seems the hardest part of these "new" requirements.

I hope you didn't lose faith, but I'll say it again: I would like to see this feature merged, and I have the power to do it. I just want to be sure that the UI that made this software popular isn't lost.

QuentinRoy:
Wouldn’t it make the interaction more complicated?

QuentinRoy:
This “no currency” just looks like an extra currency with some special behaviors.

Glandos:
[**@Quentin Roy**](https://github.com/QuentinRoy) commented on [7 avr. 2020 à 19:16 UTC+2](https://github.com/spiral-project/ihatemoney/pull/541#issuecomment-610513437 "2020-04-07T17:16:17Z - Replied by Github Reply Comments"):
> This “no currency” just looks like an extra currency with some special behaviors.

Yes, exactly. But I have the feeling that a switch "Multiple currency" on-off and a dropdown menu is more complicated than a single dropdown with a special currency. I can be wrong…

QuentinRoy:
Oh I see, that makes sense. I am not bought on the idea of having two different mode of operations though. Modes are not cheap in a UX point of view. Concretely, what is the main difference between these two modes? Is this about one dropdown at bill creation, and an other in the settling page? The most critical is probably the bill creation page since it is much more frequent, and you'd need to be able to create a bill super quick..
Please disregard my feedbacks if I am slowing you down. I apologize, I don't really have the time to do much more than participating a bit in the discussions at the moment :-/.

Glandos:
There is also the added columns in the bill list:

Screenshot_2020-04-08 Account manager - test

This new column only make sense when there is really two different currencies in the bill list.

QuentinRoy:
Do you think we need this extra column? Maybe the conversion is only needed when settling? Oh the balance is a problem I suppose if you do not define a “default” currency.. An alternative to solve this issue could be an option to choose the balance currency (maybe an overlay panel could show the detail when we hover on a balance entry). We could also show all currencies in the balance (e.g. 2-3 balances), but that may quickly become too much.
Splitwise Support multiple currencies, but don’t even make currency conversions. When settling.. I am wondering if they did user testing before going for that...

dark0dave:
@QuentinRoy, this pr changes it so the project has a default currency, which all bills are settled in. In this way there aren't multiple currencies to settle on. Hopeful easing the strain on the users. If we need to add a feature to settle in multi currencies it can be done **later**.

dark0dave:
> @dark0dave as a start let me be clear: we want to merge this, because it fits a user need, and you are not alone.
> 
> But I agree with @indatwood the UI/UX should be as simple as possible for first time users. And displaying multiple columns with or without currency makes is more complex.
> 
> Your discussion is constructive on this, let me propose a compromise:
> 
>     1. Add a currency "No currency" in the dropdown list, selected by default. It will serves as a setting. When selected, all currency conversion in the project are disabled, and the UI keeps being simple.
> 
>     2. If later on, a default currency is selected, then all bills without currency should be converted to this currency.
> 
>     3. If a project currency is switched to "no currency", do nothing. New bills will have no currency, but the UI will be simpler. We can add a confirmation if the project already have bills with mixed currencies, but I consider it optional for now.
> 
>     4. Keep the dropdown as-is on the home page for now. We could have an "Advanced" section, hidden by default, but I don't think it's useful for now.
> 
> 
> I guess that adding a "no currency" is a little more work, but 2. seems the hardest part of these "new" requirements.
> 
> I hope you didn't lose faith, but I'll say it again: I would like to see this feature merged, and I have the power to do it. I just want to be sure that the UI that made this software popular isn't lost.

1. Ok, default will now be "No currency" instead of "".
2. That is how it currently works.
3. I can add this.
4. Good.

dark0dave:
@Glandos I have made the required changes.

dark0dave:
I have also added some features, which prevent users who have currencies from adding bills without currencies.

dark0dave:
I have also rebased. so its ready to be merged.

ihatemoney/currency_convertor.py
```diff
@@ -0,0 +1,48 @@
...
+    def set_rate(self):
+        if "rates" not in self._internal_cache:
+            rates = requests.get(self.api_url).json()["rates"]
+            rates[self.default] = 1.0
+            self._internal_cache.set("rates", rates, expire=self.expire)
+
+    def get_currencies(self):
+        self.set_rate()
+        rates = [rate for rate in self._internal_cache.get("rates")]
+        rates.sort()
```
Glandos, line 32:
```suggestion
        rates.sort(key=lambda rate: "" if rate == self.default else rate)
```
This is done to force the 'No currency' to be the first on the list.

Glandos:
I have one small bug:
1. Changing the default currency of a project seems to have an effect, but the dropdown of the settings page always shows "No currency".

ihatemoney/currency_convertor.py
dark0dave: ^^ thanks! Added!

dark0dave:
> I have one small bug:
> 
>     1. Changing the default currency of a project seems to have an effect, but the dropdown of the settings page always shows "No currency".

Yeah its just the default will fix.

dark0dave:
@Glandos 

> I have one small bug:
> 
>     1. Changing the default currency of a project seems to have an effect, but the dropdown of the settings page always shows "No currency".

This has been fixed.

dark0dave:
@Glandos I would appreciate it if you could take another look and let me know if there is anything else that needs to be changed.

indatwood:
Hi, thanks for your work @dark0dave :-)

Here is another review from my perspective. If you feel this is too much of a change, we might be able to merge as is, and I will propose the changes in a separate pull request.

- When on the home page, it asks to create a project and by default displays "Default currency : no currency". I find this strange. I would rather not have to worry about this at this stage and when the project is created, have a tooltip or something asking me if I want to use multiple currencies ;
- If the currency is set to "No currency", then I would expect that no currency is asked when I enter a new bill, but this isn't the case ;
- The "Default : XXX" could be done like it's done with the "external link", with a text under the input rather than with the name of the field ;
- Because I don't care about either "external link" or "currencies", I find these two fields should be made optional and activated / deactivated on a per project basis ;

That's it for me, apart from that it seems to work great :-)

Glandos:
[**@indatwood**](https://github.com/indatwood) commented on [17 avr. 2020 à 17:40 UTC+2](https://github.com/spiral-project/ihatemoney/pull/541#issuecomment-615316420 "2020-04-17T15:40:37Z - Replied by Github Reply Comments"):
> - When on the home page, it asks to create a project and by default displays "Default currency : no currency". I find this strange. I would rather not have to worry about this at this stage and when the project is created, have a tooltip or something asking me if I want to use multiple currencies ;

We may need to have a complete discussion on project creation UX for that. I would keep it as-is for now.

> - If the currency is set to "No currency", then I would expect that no currency is asked when I enter a new bill, but this isn't the case ;

I agree, and it should be quite easy to hide.

> - The "Default : XXX" could be done like it's done with the "external link", with a text under the input rather than with the name of the field ;

Also agree, even if it can be done later as UI enhancement request.

> - Because I don't care about either "external link" or "currencies", I find these two fields should be made optional and activated / deactivated on a per project basis ;

Can you explain this in detail? Don't you think it can be enhanced by another request, to make this merge and go forward?

indatwood:
>> Because I don't care about either "external link" or "currencies", I find these two fields should be made optional and activated / deactivated on a per project basis ;
>
> Can you explain this in detail? Don't you think it can be enhanced by another request, to make this merge and go forward?

Sure thing, I'll try to explain myself better : The currency feature and the external link feature are not useful to me (and I believe to the majority of users), and I think having an option to activate them, in the settings of a project would make things better / easier for me (less inputs, more clarity).

I'm okay to have a discussion about project creation UI, and I'm okay to discuss this in another issue. Generally speaking, I tend to avoid introducing (UI) regressions because we want to get things out faster. I believe it's better to fix things at their inception, rather than creating technical debt.

In my experience, the vast majority of the cases, the "small fixes that are out of the scope of this feature" are almost never implemented.

But that's just my point of view, and I'm new here, so I will be happy to accomodate to how things work around here.

Also, I think it's good to mention that I'm not doing this against you @dark0dave. You invested a lot of effort in this, and thank you for that. I don't want to be rude or anything, I just want to do what I think is best suited for this project.

if @dark0dave doesn't want to invest more time on this (you did already a lot), I would be happy to continue working on a branch until we reach a state when it's mergeable to master, without introducing things that should be fixed later on.

dark0dave:
>     * When on the home page, it asks to create a project and by default displays "Default currency : no currency". I find this strange. I would rather not have to worry about this at this stage and when the project is created, have a tooltip or something asking me if I want to use multiple currencies ;

I agree with @Glandos, this just ordering does not matter right now.

>     * If the currency is set to "No currency", then I would expect that no currency is asked when I enter a new bill, but this isn't the case ;

I can add this. 

>     * The "Default : XXX" could be done like it's done with the "external link", with a text under the input rather than with the name of the field ;

This is for later.

>     * Because I don't care about either "external link" or "currencies", I find these two fields should be made optional and activated / deactivated on a per project basis ;

^^ This seems like a completely seperate request. You are changing both the new and the old UI. Make you own pr, after this one has landed.

dark0dave:
I have added, the feature which prevents users from even seeing currency. I think this should be the last one before opening a new pr with additional changes. Can we merge please.

Glandos:
I think you need to rebase after merging the "history" feature. Can you do this? Afterwards, I think I'll merge it.

dark0dave:
@Glandos I am happy to do that :)

dark0dave:
I can rebase, but all the tests now fail :(. I assume the recent changes have not gone well?

dark0dave:
I had to change the cache, due to python 3.8

Glandos:
Thanks, I've resolved some easy conflicts, and now, there is one last thing: this history merge need to update the database migration. But fear not, I have done this, I just don't have push access to your repository, so here's the file. It should replace every other migration scripts in `migrations/versions/`:

https://gist.github.com/Glandos/4a81d0e0c1732edf6cbe41a3f632d100

And then: merge party!

dark0dave:
@Glandos thanks! Doing now!

Glandos:
I'm sorry, I'm trying to fixing tests. I still have to found why `HistoryTestCase::test_double_bill_double_person_edit_second` is failing. All others are easily fixable.

Glandos:
Got it! History was messing around, but I made some quick changes. It's not perfect, but usable.

zorun:
Good work!  I spotted a few issues (nothing major), expect some tickets ;)

PR #544, opened 2020-03-21
 Integrate illustrations as a showcase

Jojo144:
Continue #383.

Voilà ! C'est prêt je pense. N'hésite pas à ne pas prendre certains commits s'il ne te conviennent pas.


Jojo144:
@almet Yo. Auras-tu du temps à un moment pour jeter un œil ?

indatwood:
Hey, I believe @almet is trying to step down as a maintainer here. I'll have a look!

CONTRIBUTORS
```diff
@@ -37,3 +37,5 @@ Richard Coates
 THANOS SIOURDAKIS
 Toover
 Xavier Mehrenberger
+
+The manual drawings are from Coline, they are under CC BY 4.0.
```
indatwood, line 41: Hey, is it possible to have a full name here?

Jojo144: I made it more precise indeed.

Jojo144:
Cool! Thanks!

Glandos:
I understand that the integration of photoswipe make it super easy for the user to look at the showcase and go back, but now we load the photoswipe machinery on every page: more than 80Kio of CSS and Javascript. It's nearly 20% of the page weight. It seems too much to me.

What about having a real link to another page, and load resources as it was done with the dashboard?
And if we want it integrated as it is now, we can load this new page in a new iframe on top of the current page.

Jojo144:
Ah. I have to admit that you are right!
What would you think about a solution based on loading the JS file dynamically?
E.g. https://stackoverflow.com/a/593483

Glandos:
It would also fit I guess.

I don't like loading JS resources dynamically because it is hard to debug.
I don't like iframe, because I think it is really a bad design workaround.

Pick your choice from here :)

Note however that using an iframe will also avoid loading the `showcase.html` in every page, even if it's simple HTML. But maybe that can also be achieved with dynamic JS.

Jojo144:
Hey!
What do you think of that?
Tested on Firefox and Chromium.

Glandos:
Cool stuff. Its hacky, but I don't think we can get a smaller hack, so this is good for me.

Do you think you can integrate the latest photoswipe before merging? It seems to be a small release, but it won't have to be done later.

Glandos:
By the way, tests are failing because of the remove `hand-holding-heart.svg` that was replaced. It's a good time to change the name in tests :+1:

Jojo144:
Thanks for being so reactive!
I will do your two suggestions and rebase against master.

Jojo144:
Et voilà!
Ready for being merged for me.

Glandos:
Merci :)

PR #546, opened 2020-04-04
modified run.py to make session permanent

rosechar:
in order to maintain language preference across sessions


rosechar:
going to open a new one after running black, my bad

Issue #135, opened 2016-02-02
Add the language code in the URL or in a cookie so that we can keep it between session.


Natim:
It could be nice to be able to save the bookmarked in the right language.
For each new session it's back to english.

almet:
That's a good idea! Would you like to work on a fix?

anunay999:
I would like to work on this issue. Can you help me on this

almet:
Hi @anunay999 ! Thanks for proposing to work on this.

Currently, there is a way to select the language, which is in the top-right corner. Once you've selected the language and if you close and reopen your browser, the default choice is selected again (I believe it's the language of your browser).

There is one relatively simple way to solve this, by persisting the language information into a session cookie. Currently, I hate money (IHM) use sessions to store some stuff, among others the language, but it's not persisted, so if you close / reopen your browser the information is lost.

Here is where the session is written to:

https://github.com/spiral-project/ihatemoney/blob/94e83306c344c47d49a5dc440df6d10d9a007172/ihatemoney/web.py#L631-L634

After some research, the fix for this is rather simple: you need to [mark the session as permanent](https://flask.palletsprojects.com/en/1.1.x/api/#flask.session.permanent).

The place to configure the session is [here](https://github.com/spiral-project/ihatemoney/blob/94e83306c344c47d49a5dc440df6d10d9a007172/ihatemoney/run.py#L149).

So… let's do it :-)

rosechar:
Hi, is this currently being worked on? If not I would like to give it a try.

Glandos:
I don't think so. You can have a look at it :)

Glandos:
Maybe we can change it in another request, if it's annoying.
But another "good" change would be to select the language based on UA locale.

rosechar:
I have submitted a pull request with the functionality of making session permanent! 

could I ask you to elaborate on your last comment about selecting language based on UA locale? I am confused about what UA locale is

Glandos:
Sorry, I guess I made a mistake.

The browser sends a list of accepted language. And the software should select a sane default based on this list.

indatwood:
@Glandos I'm not sure what to do about this then : we merged but we shouldn't have? or should we close this?

Glandos:
The discussion was a bit confused, but we can close this.

PR #547, opened 2020-04-04
updated web.py to make session permanent

rosechar:
to maintain language preference

### DIFF
--- a/ihatemoney/run.py
+++ b/ihatemoney/run.py
@@ -147,11 +147,6 @@ def create_app(
     app.jinja_env.globals["locale_from_iso"] = locale_from_iso
     app.jinja_env.filters["minimal_round"] = minimal_round
 
-    # Set session to permanent to make language choice persist
-    @app.before_request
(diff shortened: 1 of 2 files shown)


Glandos:
Is it required to make the session permanent for each request?

Why not setting it permanent after successful authentication in [authenticate](https://github.com/spiral-project/ihatemoney/blob/master/ihatemoney/web.py#L192)?

rosechar:
@Glandos yes that is something to consider. In the issue request (#135) @almet mentioned that the place to configure the session is [here](https://github.com/spiral-project/ihatemoney/blob/94e83306c344c47d49a5dc440df6d10d9a007172/ihatemoney/run.py#L149) so I followed that advice. I could move it to where it is authenticated if that is preferred

Glandos:
I'm not an expert of Flask sessions, but what I found on Q&A makes me think it's better like this.
We can simply revert this in case of troubles :)

PR #549, opened 2020-04-05
fix(Home): correct responsive width of cards

eMerzh:
To avoid issues likes this, 

image

let's drop min width for small screens


indatwood:
Thanks! That looks good to me :-)

Glandos:
It's funny, I don't have this behavior with the current master, but anyway…

PR #550, opened 2020-04-05
merge settings and import to avoid clutter

eMerzh:
I've previously tried to merge settigns and export to avoid having 12 items on the menu.
Somehow an import menu sneaked in ;-)

here is an attempt to merge them, because they don't have a point to be separated.

also adding some spacing and help....
 should give smth like this:

Capture d’écran 2020-04-05 à 21 08 09-fullpage


Glandos:
Very nice!

But when I select a file, it isn't shown instead of `Choose file`. Is it normal?
Of course, the import action works well, but in the older page, the selected file name was displayed as a confirmation.

eMerzh:
oups sorry, forgot to add a neaded bit

Glandos:
Thanks!

PR #555, opened 2020-04-15
undo bill delete

rosechar:
here is my proposed solution for undo functionality. it adds a button to the flash dialog after a bill is deleted. however, when the dialog occurs, it makes the "Add new bill" button temporarily not clickable. I have been struggling with how to fix this for a few days, I am wondering if any of you have insight on this issue?? in any case, let me know what you think

### DIFF
--- a/ihatemoney/static/css/main.css
+++ b/ihatemoney/static/css/main.css
@@ -429,7 +429,6 @@ tr.payer_line .balance-name {
   position: absolute;
   top: 4.5rem;
   width: 100%;
-  pointer-events: none;
 }
 
 .light {
--- a/ihatemoney/web.py
+++ b/ihatemoney/web.py
@@ -705,13 +705,14 @@ def delete_bill(bill_id):
 
     url = url_for(".undo_delete_bill")
     alert = (
-            'The bill has been deleted <a class="alert-link" href="'
-            + url
-            + '" id="undo"> undo </a>'
+        'The bill has been deleted <a class="alert-link" href="'
+        + url
+        + '" id="undo"> undo </a>'
     )
     flash(Markup(alert))
 
     return redirect(url_for(".list_bills"))
+
 
 @main.route("/<project_id>/undo")
 def undo_delete_bill():
--- a/ihatemoney/web.py
+++ b/ihatemoney/web.py
@@ -27,6 +27,7 @@ from flask import (
     url_for,
     send_file,
     send_from_directory,
+    Markup,
 )
 from flask_babel import get_locale, gettext as _
 from flask_mail import Message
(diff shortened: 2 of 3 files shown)


indatwood:
Hi, and thanks for proposing this!

From the user perspective, I believe it's a good feature. I've checked out and tried locally but didn't manage to make it work. A undo link was shown, but I tried to click on it and it then disappeared, and I see no other way to undo the delete.

I then tried a second time and had the same issue. In general, I believe the "undo" text should be more explicit, maybe with a sentence "The bill has been deleted. You can undo the deletion by clicking here", otherwise it lacks expliciteness in my point of view.

The flash message disappears too quickly, and I'm even thinking we should let it on the screen, with the possibility to close it rather than making it disappear automatically for this specific case.

ihatemoney/static/css/main.css
```diff
@@ -429,7 +429,6 @@ tr.payer_line .balance-name {
   position: absolute;
   top: 4.5rem;
   width: 100%;
-  pointer-events: none;
```
indatwood, line 432: why ?

ihatemoney/web.py
```diff
@@ -698,10 +699,34 @@ def delete_bill(bill_id):
...
     db.session.commit()
-    flash(_("The bill has been deleted"))
 
+    url = url_for(".undo_delete_bill")
+    alert = (
+        'The bill has been deleted <a class="alert-link" href="'
+        + url
+        + '" id="undo"> undo </a>'
+    )
+    flash(Markup(alert))
```
indatwood, line 712: I find it weird to have this here, rather than in the templates. Maybe we should instead specify this in a template and display it if needed?

ihatemoney/static/css/main.css
rosechar: I removed this to prevent the situation you described above where clicking it does not trigger any action. if you try it in a incognito tab / safari it may work (I had the same issue while testing). in any case, I see what you mean about making the option more explicit. I will take a look at making these adjustments, thank you for the feedback

ihatemoney/web.py
rosechar: sure, I will try doing it via template

indatwood:
Awesome :-) Let me know if you need any help / review :-)

Glandos:
@Andrew-Dickinson I don't know about SQLAlchemy-continuum, but do you think it can undo things based on history? It could be useful for a general "undo" feature.

indatwood:
It seems to be the case, yes! see https://sqlalchemy-continuum.readthedocs.io/en/latest/revert.html

Andrew-Dickinson:
Yep! It's actually a [primary use case](https://sqlalchemy-continuum.readthedocs.io/en/latest/revert.html) for the library. There are definitely going to be corner cases for this application though (e.g. users who have project history disabled won't be able to revert changes)

Glandos:
So @rosechar what do you think of using Continuum's revert feature?
You can still use the session to store the last version of the form, and commit a revert from it. I don't know if `to_json` will still be useful.

It will be a very good use case for Continuum here.

rosechar:
@Glandos yes, this seems to be the optimal way to do it

indatwood:
Hi @rosechar do you still want to work on this? Let us know. Also, if you need some help don't hesitate, we might offer some to get this merged. Cheers!

almet:
Hi, I'm closing this since it has not got any activity since last april. Thanks for the work on this though.

PR #557, opened 2020-04-20
added tagging functionality and filtering by date, amount, payer, tag

### LINKED ISSUE #55: Add tags feature
Would be nice if we could append a tag to a bill, to sort them by categories (rent, food, party, etc).

Thanks for you hard work !

### LINKED ISSUE #430: Add the ability to search / filter expenses
In order to easily find a particular transaction (or a type of transaction).

The filter could take into account independently or cumulatively : 
Amount / Date / Payor / Recipient / Category ... (with range where it's relevant)

allisongu:
@nachow16 and I wrote this code to fix #430 and fix #55


ihatemoney/models.py
```diff
@@ -434,6 +434,7 @@ def delete(self, project, id):
     date = db.Column(db.Date, default=datetime.now)
     creation_date = db.Column(db.Date, default=datetime.now)
     what = db.Column(db.UnicodeText)
+    tag = db.Column(db.UnicodeText)
```
Glandos, line 437: Migration is missing for this new column

ihatemoney/forms.py
```diff
@@ -216,6 +217,9 @@ def save(self, bill, project):
         bill.payer_id = self.payer.data
         bill.amount = self.amount.data
         bill.what = self.what.data
+        tag = list(set(part[1:] for part in bill.what.split() if part.startswith('#')))
+        if tag:
+            bill.tag = tag[0]
```
Glandos, line 222: If I understand correctly, tags will be created like hashtags in tweets.

I think this could be mentioned somewhere in the bill creation form (tooltip?).
Don't you think they should be removed from the `what` column afterwards, to avoid duplication?

Glandos:
Hi!

Nice improvement. It seems simpler that what I expected, so good job!

There is however some issues for now, please have a look.

ihatemoney/forms.py
allisongu: We could definitely add a tooltip to the bill creation form! We weren't sure if that would clutter up the form too much.
As for removing it from the `what` column afterwards, we were thinking that people may also include the hashtag as part of their description like so: "#Rent for July" and in that case, we wouldn't want to remove it.

ihatemoney/models.py
allisongu: So we actually had a ton of issues with the migration. We tried running `make create-database-revision` like it says in the Contributing Guidelines on several clones of the repo with our changes, but the script would not work. It would write code to create a table with all the columns including tags. We would then fail tests because "project table already exists" and we'd be unable to run the server because somehow oppositely it wouldn't exist. Then we tried to write/edit the migration file to only add/drop the new tag column but then we'd fail the tests because "duplicate column name: tag" and the server/site wouldn't work either.  Not doing a migration was the only way we were able to get into the application and see that our changes were working... Not sure what the issue was; any ideas?

allisongu:
So I've run `make create-database-revision` and it created the upgrade and downgrade functions that are commented out in `321d3f042213_added_tag_column_to_bill_model`. When I run the tests, I fail every single one due to "table sqlite_sequence may not be dropped" and "table bill_version already exists". In the non-commented out code is my edit of the migration file that fails 6 tests due to "duplicate column name: tag". Really not sure how to fix this, does anyone know what I should do?

ihatemoney/forms.py
Glandos: OK, now I know that I don't use enough Twitter ;) I'm still using tags as optional fields like in blog post.

But you may be right, adding tags in another field is too complicated for this application's goal. So let's keep it like this for now.

zorun:
This sounds good! Being able to search is especially important now that we have paginated output (#480), because a simple browser-based text search won't show everything.

Is it possible to have a screenshot of how the search functionality would look like?

allisongu:
> This sounds good! Being able to search is especially important now that we have paginated output (#480), because a simple browser-based text search won't show everything.
> 
> Is it possible to have a screenshot of how the search functionality would look like?

We used a previous PR's code to start us off and this is how it ended up looking:
Screen Shot 2020-04-23 at 12 26 05 PM

zorun:
@allisongu great, thanks!

If you want to test what happens with pagination, you can change the number of items per page to a really low value in `ihatemoney/web.py`:

```
.paginate(per_page=100, error_out=True)
```
Although I see now that we also have javascript-based pagination and sorting (#538), I'll have to check how the two pagination methods interact...

zorun:
> So I've run `make create-database-revision` and it created the upgrade and downgrade functions that are commented out in `321d3f042213_added_tag_column_to_bill_model`. When I run the tests, I fail every single one due to "table sqlite_sequence may not be dropped" and "table bill_version already exists". In the non-commented out code is my edit of the migration file that fails 6 tests due to "duplicate column name: tag". Really not sure how to fix this, does anyone know what I should do?

To generate migrations, you need to make sure that you have an up-to-date database.
The simplest way is to start a new database from scratch:

- remove or rename any leftover database from previous testing (it should be at `/tmp/ihatemoney.db`)
- start from the latest ihatemoney code
- run `make serve` or `python run.py` as explained in the contribution guide
- stop the dev server
- apply your code changes
- create the migration with `create-database-revision`

It should pick up your changes. I tested it and it also included something strange and completely unrelated about `sqlite_sequence`:

```
INFO  [alembic.autogenerate.compare] Detected removed table 'sqlite_sequence'
INFO  [alembic.autogenerate.compare] Detected added column 'bill.tag'
INFO  [alembic.autogenerate.compare] Detected added column 'bill_version.tag'
```

If something like this happens (it might depend on the system you're using...), you would need to manually edit the migration file to only keep the bits related to the new field.

allisongu:
@zorun The instructions to get the database revision fixed worked! Thanks so much :) That might be a good thing to include in the Contributing Guidelines for other newbies too.

I tested out our search functionality with a low number of bills per page (1 haha) and it does not search through all bills currently, only the ones shown on the page.

zorun:
Great! The documentation update is a good idea: #569

For the issue with pagination, it needs some thinking. One way would be to do the searching/filtering on the server side. However, it means that:

- either we reload the entire page when the search/filter conditions have changed. This doesn't seem very user-friendly.
- either we need to load the list of bills asynchronously. This is probably a good idea in itself, but quite a large change.

Any other ideas?

zorun:
@allisongu what do you think of doing this in two steps?

The first step could be to just add tags, and provide simple server-side filtering on tags. Regarding the UI, maybe each tag in the bill description could be a link that filters down on this tag? Or also add a filtering bar like you did, but only for tags? Filtering would need to reload the page, which is fine as a first step I think.

The second step would then add more powerful searching (by date, payer, etc), and look for a better solution to the pagination problem & asynchronous loading.

This is just a suggestion and more ideas are welcome.

eMerzh:
i'll +1 @zorun , maybe first doing small steps.

Also, please keep in mind that we want to try to keep this page as light as possible, so adding a full blown search form in a big card might be a bit too much.

Glandos:
We are on the way to discover that filtering is complex, especially from the UI point of view.
But the tag part can be simple in both implementation and UI.

@allisongu, do you think you can open a new request with only the tag management in the database? As @zorun mentioned, you can also add a browse-by-tag feature, like this one: https://www.koreus.com/tag/chaton When you click on a new tag, it filters on both selected tags.
In my opinion, the tag list could be shown on demand, with a "Filter" button. In this list, we can later add a more complex filtering header.

indatwood:
Hi, here we are with this? It seems kinda stalled (last discussion and commit 6 months ago) but the feature is cool :-)

@allisongu, do you want to finish your work on this or should we let it go?

allisongu:
@indatwood Hey, unfortunately I don't have the time to contribute to this feature anymore so unless someone else wants to pick it up, i guess we should let it go

Issue #89, opened 2012-09-03
Add an « undo delete » action.


aavenel:
It's very easy to delete a bill by mistake (only one click).

I suggest to add an "undo" link in the flash message so the user have some seconds to undo his action.

almet:
Should we get this done?

almet:
I've just seen that github displays a « Are you sure you want to delete this » alert. That probably should be enough if we want, but having a way to « undo » the delete would work as well!

rosechar:
I am going to try taking this on using the approach outlined ( using session object to store the deleted record data)

almet:
Hey, I'm handing this to the new maintainers @glandos and @Natim. You're in good hands :-)

eMerzh:
My two cents here: UX wise, i think it's always better to do undo than confirm.

QuentinRoy:
Confirmation dialog are likely to be "chunked". You quickly get into the habit of instinctively clicking on it. It may prevent some "slips" but not "mistakes" (https://www.interaction-design.org/literature/book/the-glossary-of-human-computer-interaction/human-error-slips-and-mistakes).

indatwood:
@rosechar seems you have an answer here :-) Let me know if you need any help on this :-)

rosechar:
That certainly makes sense from a UI perspective. I will take the approach of an undo, thank you for your input

PR #565, opened 2020-04-22
Implemented a solution for 'undo delete'.

cjglover14:
I've created a quick fix for the undo delete button. It's a bit rough around the edges, but this was my first time making a pull request to an open source project, and I'm happy it works!

### DIFF
--- a/ihatemoney/messages.pot
+++ b/ihatemoney/messages.pot
@@ -204,6 +204,9 @@ msgstr ""
 msgid "The bill has been deleted"
 msgstr ""
 
+msgid "UNDO"
+msgstr ""
+
 msgid "The bill has been modified"
 msgstr ""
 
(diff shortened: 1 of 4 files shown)


Glandos:
This is another solution for undo, as already proposed in #555 trying to solve #89 
There are some good things, and some less good :)

First, there are some tests. This is clearly a very good thing!
Then, I like the API for `/undo`. This is the same as in #555. 

Here are some points to improve:
- The `list_bills_post_delete.html` seems to be a huge copy/paste of existing template. This will make future development harder and should be avoided.
- There are some change unrelated to the main goal of this request, like rewriting comments. While this is good, those changes should be done in another request.

As a conclusion, I think that using a "ghost" form is a bit overkill. Could you explain its advantage?
Otherwise, I would like to guide you to Continuum, which seems to do what we want. See https://github.com/spiral-project/ihatemoney/pull/555#issuecomment-617849639

indatwood:
Hi @cjglover14 ! Do you want to work more on this to improve it and be merged? If so, let us know; We might also offer some help if that's needed. Cheers!

almet:
Hey, I'm closing this as it has no activity since about one year. Thanks everybody involved!

PR #569, opened 2020-04-23
Improve documentation regarding database migrations

zorun:


Glandos:
Documentation is good. Always good, thanks!

PR #570, opened 2020-04-24
Preparing release 4.2

Natim:


zorun:
Thanks for preparing this!

Here is a proposal for the changelog:

```
5.0 (unreleased)
================

Breaking changes
----------------

- Drop support for Python 2 (#483)
- Drop support for Python 3.5 (#571)

The minimum supported version is now Python 3.6

Added
-----

- Add espanol latino america (es_419) translation
- Add German translation (#492)
- Allow to import previously exported json data (#518)
- Add new optional field "external link" in bill form (#429)
- Add new statistics showing monthly expenses (#526)
- Add pagination to the list of bills (#480)
- Add sorting, pagination, and searching to the admin dashboard (#538)
- Add Project History page that records all changes (#553)
- Add token-based authentication to the API (#504)

Changed
-------

- Use the external debts lib to solve settlements (#476)
- Remove balance column in statistics view (#323)
- Remove requirements files in favor of setup.cfg pinning (#558)
- Make language choice persistent (#547)

Fixed
-----

- Improve input of email addresses when inviting people to join a project (#133)
- Fix datepicker that was displayed twice on some browsers (#221)
```

zorun:
Maybe the release should be done after fixing https://github.com/spiral-project/ihatemoney/pull/553#issuecomment-618532880 and waiting for translations to be complete?

zorun:
We should check whether the new Currency features (#541 ) changes the API and import/export in a backwards-compatible way.  If not, we need to inform about it in "breaking changes".

Examples:

- export a project with ihatemoney 4 and import it in ihatemoney 5
- export a project with ihatemoney 5 and import it in ihatemoney 4
- check that MoneyBuster is still happy with the API

If necessary, we could add some code for compatibility in the upcoming 4.1.4 (#588).

Glandos:
[**@zorun**](https://github.com/zorun) commented on [30 avr. 2020 à 13:56 UTC+2](https://github.com/spiral-project/ihatemoney/pull/570#issuecomment-621786915 "2020-04-30T11:56:21Z - Replied by Github Reply Comments"):
> We should check whether the new Currency features (#541 ) changes the API and import/export in a backwards-compatible way.  If not, we need to inform about it in "breaking changes".

For now, it's impossible to break, since the export doesn't [export currencies](https://github.com/spiral-project/ihatemoney/blob/15ab04e636f1c267da3ef1c6bd4cdc68449b2e97/ihatemoney/models.py#L244). This seems a bug by the way, but it's a good time to think about it.

indatwood:
Hey @Natim ! This seems to be conflicting related to translation files, could you update it? And then we could release 5.0!

Glandos:
@Natim I think you can just keep version number, since changelog entries were integrated (and augmented), and `mo` files were also updated since then. The conflicts should be resolved.

docs/conf.py
```diff
@@ -3,9 +3,9 @@
 master_doc = "index"
 
 project = "I hate money"
-copyright = "2011, The 'I hate money' team"
+copyright = "2011-2020, The 'I hate money' team"
```
Glandos, line 6: It's 2021 now :wink:

Issue #475, opened 2019-09-24
Should we drop support for python 2.X ?


almet:
Hi,

I'm wondering if we should drop support for python2.X series, and eventually modernize a bit the codebase accordingly.

What do you think?

Glandos:
Since every major distribution have Python 3 in them (CentOS 8 is out!), Python 2 can be safely dropped.

To me, the minimal version should 3.5, but 3.6 is nice with f-strings.

almet:
I believe having 3.6 (or even 3.7 ?) as a target would be sane. IHM is not packaged yet on any distribution, so I believe it would be okay to ask people who want to install IHM to deploy a recent version of python? (not sure, but that's a wild guess)

zorun:
Dropping support for python2 is a good idea!

Regarding python3, this was discussed in #421, and I still believe that compatibility with python 3.5 is important.
The end-of-life for python 3.5 is still one year away (2020-09-13): https://devguide.python.org/
Same thing for Debian stretch, it is still supported until 2020 at least.

Issue #421, opened 2019-02-20
Drop support for Python < 3.6 and use f-strings


Natim:
![image](https://user-images.githubusercontent.com/229453/53079468-c26c7e00-34f6-11e9-8acd-7b31081c24f4.png)

zorun:
Debian stretch comes with python 3.5, so dropping support for it does not look like a great idea…

Glandos:
Indeed, but Buster will be delivered with 3.7.

Will a backport be in Stretch?

zorun:
I doubt very much that python 3.7 will be backported to stretch.  I expect that lots of servers will continue to use stretch for some years to come: it will receive security support from the Debian team until 2020, and then it will get LTS support until June 2022: https://wiki.debian.org/LTS

So, unless there is a very good reason to drop support for Python 3.5, please avoid doing this!  F-strings look like a nice feature but it's mostly code cosmetics, everything you can do with it can also be done with the "old-style" string formatting syntax.

Natim:
Thank you for your insights on this, I am closing for now.

Glandos:
@zorun You're right on a lot of aspect, and I agree that supporting 3.5 is a good idea.
But f-strings are not __only__ cosmetic: https://cito.github.io/blog/f-strings/

PR #571, opened 2020-04-24
Remove obsolete python code (<3.6).

Natim:


Glandos:
I personally love f-strings. They are cleaner, and on top of that, more performant at runtime.

But are we sure than we can expect all our user base to have python >= 3.6?

Natim:
> But are we sure than we can expect all our user base to have python >= 3.6?

Definitely yes. [Python 3.5 endoflife is September 2020](https://devguide.python.org/#status-of-python-branches) and people can still use a previous version in the meantime.

Glandos:
And Python3.6 is available in CentOS7. OK, let's go.

zorun:
As I already said in #421 and #475, I'm a bit sad about dropping support for Python 3.5 while both Python 3.5 and Debian stretch are still maintained (until September 2020 for Python 3.5, and until [June 2022](https://wiki.debian.org/LTS) for Debian stretch). But at this point, buster has been out for 9 months already so it's less annoying. I guess I'll just have to finally upgrade my server ;)
This will of course need to go in the upgrade notes for IHM 5.0.

@Natim btw, you seem to have missed a reference to python 3.5 in `docs/installation.rst`

JocelynDelalande:
Ouch, I missed the conversation about it.

A bit sad also, because as yunohost is still on stretch, and it will stop ihatemoney_ynh from receiving ihm upgrades for a while for little actual gain (except dev ease).

But go ahead, yunohost will upgrade to buster soon, and it seems that I am a few months late for such an objection :). My bad

PR #573, opened 2020-04-24
Reorganize "Contributing" documentation to be more accessible to new contributors

zorun:
The following changes were done:

- move "How to contribute" first, and "Set up a dev environment" as the second section.  This way, a new contributor has access to the most general information first, instead of the very specific "dev environment" documentation.

- reduce "How to contribute" / "As a developer" by moving the part about tests to a new "Contributing as a developer" section (see below).  This way, all types of contributions get roughly the same amount of text in the first "How to contribute" section.

- add a new "Contributing as a developer" section, which lists items that are useful when preparing a code contribution (running tests, formatting code, create database migration).  These items were moved either from "How to contribute" or from "Set up a dev environment".

I also added a commit to expand a bit on how to add tests.

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -126,9 +126,18 @@ Running tests
 Please, think about updating and running the tests before asking for a pull request
 as it will help us to maintain the code clean and running.
 
-To do so::
+To run the tests::
 
     make test
+
+Tests can be edited in ``ihatemoney/tests/tests.py``. If some test cases fail because
+of your changes, first check whether your code correctly handle these cases.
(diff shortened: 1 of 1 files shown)


eMerzh:
cool, nice work...

What do you think about doing a TLDR of that in the readme... it feels a bit dull to me ATM...
if you don't feel like it, i might take a shot later...

zorun:
Hmm, the README already links to the contribution guide, I'm not sure it's a good idea to duplicate information...

Natim:
Thank you !

PR #575, opened 2020-04-24
Fix translations

zorun:
Fix source strings for two translation issues (see commit messages for details)


zorun:
Oops, I forgot to update tests. Will fix.

zorun:
Tests should now be passing.

ihatemoney/translations/cs/LC_MESSAGES/messages.po
```diff
@@ -739,3 +739,9 @@ msgstr ""
 msgid "Period"
 msgstr ""
 
+#~ msgid "%(member)s had been added"
+#~ msgstr ""
+
+#~ msgid "Person"
+#~ msgstr ""
```
Natim, line 746: I guess we can remove them.

zorun: Ah yes, right! Thanks for the review

zorun:
@Natim thanks for merging!

I think it's useful to keep old strings (especially if they are translated in several languages) because Weblate can use them to suggest translations for new strings if they look similar.

Natim:
Yeah I think weblate is less clever than tool I am used to. Such as Pontoon or Lingohub in that matter.
Especially regarding conflict resolution and keeping info in a local database.

Natim:
I put them back in 64fbda8

zorun:
Thanks!

Issue #576, opened 2020-04-24
Improve translatability of the new "History Page" feature


zorun:
The new History Page (#553), while awesome, is a bit hard to translate as it stands currently.

The first problem is that the type of object is uppercased, because it's the first word in English. However, it might not be the first word in other sentences. For instance: `Participant Foo added` could be translated as `Ajout du participant Foo` or even `Ajout de Foo comme nouveau participant`. 

The second problem is that sentences are made of several small translatable strings such as `_("added")`, `_("and")`, etc.  Without context, this is hard to translate, and it also doesn't allow to reorder parts of the sentence in different languages (see example above).

Translatable strings should be built as a single string with variable interpolation, for instance:

```
Person %(name)s added
Bill %(bill_description)s removed
Bill %(bill_description)s: amount changed from %(old_amount)s to %(new_amount)s
Bill %(bill_description)s: date changed from %(old_date)s to %(new_date)s
```

I'm not sure how to best handle markup: it can be either added to the translatable string, or can be added at interpolation time (i.e. `%(bill_description)s` would be replaced by `<em>My bill description</em>`). I think the second solution is better.

@Andrew-Dickinson I had a look at the template and the macros. Unfortunately it looks like it would require a large amount of changes :/

Natim:
I did a first pass in #581 to be able to translate it in French. It could be sufficient for now. Let's iterate if we see other issues.

zorun:
As always the devil is in the details ;) With the current French translation:

```
Facture 42.0 for Test ajouté
Facture 42.0 for Test: Supprimé Machin du Liste des débiteurs 
Facture 42.0 for Test: Ajouté Norbert et Bidule au Liste des débiteurs 
Facture 42.0 for Test renommé en Nouveau nom
```
Some of the issues: no space before ":", wrong *accords* like "ajouté" instead of "ajoutée", inconsistent upper-case, "du liste" instead of "de la liste"...

Having whole strings to translate would easily solve all these issues for all languages.

The weird part `42.0 for <description>` is another issue, see #584

Andrew-Dickinson:
@zorun Should be mostly fixed now, can you take a look at #587 and let me know if you see anything I missed?

Glandos:
Closed via #587

PR #580, opened 2020-04-25
Fix translations.

Natim:
There are a couple of translations that seems broken to me.

### DIFF
--- a/ihatemoney/messages.pot
+++ b/ihatemoney/messages.pot
@@ -396,23 +396,23 @@ msgstr ""
 msgid "Balance"
 msgstr ""
 
+#, python-format
 msgid ""
-"This project has history disabled. New actions won't appear below. You "
-"can enable history on the"
-msgstr ""
-
-msgid "settings page"
+"\n"
+"            <i>This project has history disabled. New actions won't "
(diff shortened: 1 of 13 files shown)


zorun:
@Natim see #576, I think the whole page needs some overhauling to be able to translate it properly

Natim:
> I think the whole page needs some overhauling to be able to translate it properly

Yeah that what I noticed, I tried to look at it, do you see things I missed?

zorun:
It's an improvement! But having things `<a href="#" data-toggle="modal" data-keyboard="false" data-target="#confirm-erase">` in the source to translate looks like it could break lots of things and even introduce XSS :/

There remains lots of strings that are handled through macros (`simple_property_change`, `owers_changed`) and it seems difficult.

Let's discuss on #576 about the best way to improve it and involve the original authors?

Natim:
About HTML tags in the translation files, I believe that's the right way to do it.

I agree that weblate support for it is a bit weird. I like the way LingoHub handles it.

I think we can rely on review or tooling for the XSS matter.

For the macro I agree about involving original authors.

I translated it in French and it didn't seem too complex.

PR #583, opened 2020-04-25
Fix missing HTML tag

zorun:


Glandos:
It's funny to see how HTML is tolerant.

zorun:
I think browsers had to adapt to broken webpage :) Anyway, this was only the 404 page.

Thanks for merging!

PR #584, opened 2020-04-25
Fix string representation of bills

zorun:
Currently the string representation of a Bill is: `<amount> for <description>`

It is used in the History Page to describe changes that were applied to Bills, for instance:

    Bill "42.0 for Test" renamed to "Another Test"

This is inconsistent, not easy to read, and the "for" in the middle is not translatable.

To solve this issue, simply switch the string representation of a bill to its description.

### DIFF
--- a/ihatemoney/models.py
+++ b/ihatemoney/models.py
@@ -460,7 +460,7 @@ class Bill(db.Model):
             return 0
 
     def __str__(self):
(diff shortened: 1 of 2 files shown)


Glandos:
It's good for me.
A bill should be represented with an easy string, which is its description.
Anything else should go in the template, with i18n machinery.

PR #585, opened 2020-04-25
Add .idea/ folder to gitignore

Andrew-Dickinson:
Show some love for the folks using PyCharm!


eMerzh:
IMHO it should go into you own global gitignore ... 
but if it can helps beginers, why not

Natim:
Usually, this kind of files can be added to [the profile gitignore](https://stackoverflow.com/a/5724484/186202)

PR #586, opened 2020-04-25
Direct Alembic to ignore the sqlite_sequence table

Andrew-Dickinson:
I believe this fixes both of the database concerns raised in #553:

> Starting from this change, when I use alembic to create a new automatic migration (to test something for #557 ), it picks up something strange even if I don't touch the models:
> 
> ```
> $ python -m ihatemoney.manage db migrate -d ihatemoney/migrations -m test-after-history
> INFO  [alembic.autogenerate.compare] Detected removed table 'sqlite_sequence'
>   Generating /home/dev/tmp/ihatemoney/ihatemoney/migrations/versions/9c340fd31fff_test_after_history.py ...  done
> ```

and 

> Now I remember that Alembic was giving me a warning about not taking into account ALTER while applying the migration:
> 
> ```
> INFO  [alembic.runtime.migration] Running upgrade a67119aa3ee5 -> 6c6fb2b7f229, empty message
> INFO  [alembic.runtime.migration] Running upgrade 6c6fb2b7f229 -> 2dcb0c0048dc, autologger
> /dev/venv3-ihatemoney/lib/python3.8/site-packages/alembic/ddl/sqlite.py:40: UserWarning: Skipping unsupported ALTER for creation of implicit constraint
> Please refer to the batch mode feature which allows for SQLite migrations using a copy-and-move strategy.
> INFO  [alembic.runtime.migration] Running upgrade 2dcb0c0048dc -> cb038f79982e, sqlite_autoincrement
> ```

By directing Alembic to ignore the SQLite built-in table "sqlite_sequence", which according to [the documentation](https://sqlite.org/fileformat2.html#seqtab) is managed automatically by SQLite


zorun:
Thanks, it fixed the spurious migration detection `Detected removed table 'sqlite_sequence'` :+1: 

However the `Skipping unsupported ALTER` warning is still there when applying the `sqlite_autoincrement` migration. If you think that the warning can be ignored, it's fine, because the main issue was the spurious migration.

Andrew-Dickinson:
I did a bit more digging and discovered that the `Skipping unsupported ALTER` warning was actually coming from the enum type in the `autologger` migration. By wrapping it in a batch migration I believe  that I've fixed it

zorun:
Ah right, my bad! The warning is indeed gone :+1:

PR #587, opened 2020-04-25
Improve localization on the History page using string replacement

Andrew-Dickinson:
Goes a long way towards addressing #576

Just using string replacement instead of concatenation should fix most of the issues. 

There is a slight problem when it comes to the localization of lists of objects, I've implemented what should be a fairly general solution to this problem in `utils.localize_list()`. The translation strings look a little weird, so maybe we should add some information to the translation docs about what these strings do:
```python
one = _("{single_object}")
two = _("{dual_object_0} and {dual_object_1}")
start = _("{start_object}, {next_object}")
middle = _("{previous_object}, {next_object}")
end = _("{previous_object}, and {end_object}")
```


zorun:
This looks good, thanks! The template is actually more readable this way, so it's an improvement even without considering translations.

If I understand correctly, the translation strings for `localize_list()` are completely generic and only need to be translated once? If so, it's a really nice solution!

I spotted a small typo (`unkown` instead of `unknown`), and capitalizing after a ":" looks strange to me.

It would be nice to have separate strings for each object type, for instance something like:

| Current | Proposal |
|--|--|
| {object_type} {object_description} deactivated | Participant {participant_name} deactivated |
| {object_type} {object_description} renamed to {new_name} | Participant {old_name} renamed to {new_name} |
| {object_type} {object_description} renamed to {new_name} | Bill {old_description} renamed to {new_description} |
| {object_type} {object_description}: Added {owers_list_str} to owers list | Bill {bill_description}: Added {owers_list_str} to owers list |

It would mean more strings to translate, but would provide much more context to help the translator. But if it's really too complicated to implement, the current way could work.

Andrew-Dickinson:
Capitalization after ":" looked fine to me, but I suppose it is the same sentence. I believe I've implemented what you're asking for by removing object_type as a user facing property (that always was a little weird anyway, the `something == _("Project")` kinda bothered me)

Andrew-Dickinson:
> If I understand correctly, the translation strings for `localize_list()` are completely generic and only need to be translated once? If so, it's a really nice solution!

Yep! I believe this solution is general enough to handle a pretty diverse range of counting methods (e.g. rtl text), and it should work anywhere items need to be listed in written form.

ihatemoney/utils.py
```diff
@@ -278,3 +279,71 @@ def coerce(cls, item):
...
+    start = _("{start_object}, {next_object}")
+    middle = _("{previous_object}, {next_object}")
+    end = _("{previous_object}, and {end_object}")
+
+    item_wrapper = em_surround
+    # fmt: off
+    if not surround_with_em:
+        def item_wrapper(x):
+            return x
+    # fmt: on
```
Glandos, line 328: I know one-liner are sometimes cryptic, but here, I think it will simplify code.
```suggestion
    item_wrapper = em_surround if surround_with_em else lambda x: x
```

ihatemoney/utils.py
```diff
@@ -278,3 +279,71 @@ def coerce(cls, item):
...
+
+    if len(list) == 1:
+        return one.format(single_object=item_wrapper(list[0]))
```
Glandos, line 331:
```suggestion
        return _(item_wrapper(list[0]))
```

ihatemoney/utils.py
```diff
@@ -278,3 +279,71 @@ def coerce(cls, item):
...
+    elif len(list) == 2:
+        return two.format(
+            dual_object_0=item_wrapper(list[0]), dual_object_1=item_wrapper(list[1])
+        )
```
Glandos, line 335:
```suggestion
        return _(f"{item_wrapper(list[0])} and {item_wrapper(list[1])}")
```

ihatemoney/utils.py
```diff
@@ -278,3 +279,71 @@ def coerce(cls, item):
...
+    else:
+        output_str = end.format(
+            previous_object="{previous_object}", end_object=item_wrapper(list.pop())
+        )
```
Glandos, line 339:
```suggestion
        output_str = _(f"{{previous_object}}, and {item_wrapper(list.pop())}")
```

ihatemoney/utils.py
```diff
@@ -278,3 +279,71 @@ def coerce(cls, item):
...
+        while len(list) > 2:
+            temp = middle.format(
+                previous_object="{previous_object}",
+                next_object=item_wrapper(list.pop()),
+            )
```
Glandos, line 344:
```suggestion
            temp = _(f"{{previous_object}}, {item_wrapper(list.pop())}")
```

ihatemoney/utils.py
```diff
@@ -278,3 +279,71 @@ def coerce(cls, item):
...
+            output_str = output_str.format(previous_object=temp)
+
+        output_str = output_str.format(previous_object=item_wrapper(list.pop()))
+        output_str = start.format(start_object="{start_object}", next_object=output_str)
+        return output_str.format(start_object=item_wrapper(list.pop()))
```
Glandos, line 349:
```suggestion
        output_str = _(f"{item_wrapper(list.pop())}, {output_str}")
```

ihatemoney/utils.py
```diff
@@ -278,3 +279,71 @@ def coerce(cls, item):
...
+    :param list: The list of objects to localize by a call to str()
+    :param surround_with_em: Optionally surround each object with <em> tags
+    :return: A locally formatted list of objects
+    """
+    list = list[:]
+    one = _("{single_object}")
+    two = _("{dual_object_0} and {dual_object_1}")
+    start = _("{start_object}, {next_object}")
+    middle = _("{previous_object}, {next_object}")
+    end = _("{previous_object}, and {end_object}")
```
Glandos, line 321:
```suggestion
```

Natim:
Rather than using `_().format` I would use gettext substitution.
In templates, I believe the best way to do that is to use the `{% trans %}` tag.

ihatemoney/utils.py
zorun: Does it mean that the string to translate will be `{item_wrapper(list.pop())}, {output_str}`?

In general, I don't think having python code in translated string is a good idea...

Also, compared to the original proposal, this seems less general (would it work for right-to-left scripts?) and I find it harder to understand for a translator.

zorun:
@Andrew-Dickinson thanks, the new strings look good :+1: 

No opinion on the best format for translation strings... In general I find old-style variable substitution `Foo %(variable)s bar.` confusing because of the trailing `s`.  All other methods (`format()`-style variables, f-strings, `{% trans %}` tag) are better in that respect.

ihatemoney/utils.py
Andrew-Dickinson: I agree that the strings resulting from this change seem much harder to translate for a non-programmer, or even a programmer without looking at this code. I think we should try to keep the translation strings simple and in plain language.

Andrew-Dickinson:
@Natim It doesn't look like gettext substitutions support the `{}` syntax. This is pretty weird since the `%()s` syntax is deprecated in Python 3. Am I missing somthing:
```python
>>> _("Hi %(username)s", username="Andrew")
'Hi Andrew'
>>> _("Hi {0}", "Andrew")
  File "<input>", line 1, in <module>
TypeError: gettext() takes 1 positional argument but 2 were given
>>> _("Hi {username}", username="Andrew")
'Hi {username}'
```

ihatemoney/utils.py
Glandos: Yes, my bad, I misread the code, and I proposed something wrong here.

ihatemoney/utils.py
Glandos:
```suggestion
        two = _("{dual_object_0} and {dual_object_1}")
        return two.format(
            dual_object_0=item_wrapper(list[0]), dual_object_1=item_wrapper(list[1])
        )
```

ihatemoney/utils.py
```diff
@@ -278,3 +279,67 @@ def coerce(cls, item):
...
+    if len(list) == 1:
+        return str(item_wrapper(list[0]))
+    elif len(list) == 2:
+        return two.format(
+            dual_object_0=item_wrapper(list[0]), dual_object_1=item_wrapper(list[1])
+        )
+    else:
+        output_str = end.format(
+            previous_object="{previous_object}", end_object=item_wrapper(list.pop())
+        )
```
Glandos, line 335:
```suggestion
        start = _("{start_object}, {next_object}")
        middle = _("{previous_object}, {next_object}")
        end = _("{previous_object}, and {end_object}")
        output_str = end.format(
            previous_object="{previous_object}", end_object=item_wrapper(list.pop())
        )
```

ihatemoney/utils.py
```diff
@@ -278,3 +279,67 @@ def coerce(cls, item):
...
+
+    :param list: The list of objects to localize by a call to str()
+    :param surround_with_em: Optionally surround each object with <em> tags
+    :return: A locally formatted list of objects
+    """
+    list = list[:]
+    two = _("{dual_object_0} and {dual_object_1}")
+    start = _("{start_object}, {next_object}")
+    middle = _("{previous_object}, {next_object}")
+    end = _("{previous_object}, and {end_object}")
```
Glandos, line 320:
```suggestion
```

Glandos:
Hi @Andrew-Dickinson do you still have time to rebase this promising request?

indatwood:
So, I tried merging this with master, let's see if it's working okay!

almet:
Hey, I've merged this with the latest master. I believe we should merge now before it's too late. About the translations, I think sooner is better than later anyway :-)

zorun:
Looks good to me!

PR #590, opened 2020-04-26
Localize dates and times

zorun:
See individual commits for details.

I'm not sure about this one. On the one hand, having localized date and time that follows the selected language is nice. On the other hand, the timezone issue (see 812efad22a9d3a07699feec681fccce52acc5442) means that it would make more sense to format date and time in Javascript, but then we would have less control over the formatting and it won't follow the selected language.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -31,6 +31,7 @@ connection string. This will look like::
 
     SQLALCHEMY_DATABASE_URI = 'postgresql://myuser:mypass@localhost/dbname?client_encoding=utf8'
 
+.. _the SQLAlchemy documentation: http://docs.sqlalchemy.org/en/latest/core/engines.html#database-urls
 
 `SECRET_KEY`
 ------------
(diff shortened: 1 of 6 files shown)


indatwood:
The more I think about this, the more I think that we should handle this kind of things on the client side, because on the server side, we have no idea what the locale of the user should be…

Glandos:
[**@indatwood**](https://github.com/indatwood) commented on [30 avr. 2020 à 18:27 UTC+2](https://github.com/spiral-project/ihatemoney/pull/590#issuecomment-621961126 "2020-04-30T16:27:20Z - Replied by Github Reply Comments"):
> The more I think about this, the more I think that we should handle this kind of things on the client side, because on the server side, we have no idea what the locale of the user should be…

You mean the timezone? Because `from flask_babel import get_locale` does exactly that: returns the locale selected by the user, or a sane default based on UA preference.

indatwood:
Sorry, I missed you question @Glandos ! My mind isn't fresh anymore about the topic, but I kinda remember that on the server side we cannot be sure to output the correct timezone information because it's not what's sent by the User-Agent.

So, that's was the reason behind my comment at that time. But I believe that anyway, this patch provides some enhancements, and that we should merge it anyway. It will always be time to go back to this if we feel it's needed.

indatwood:
That being said, by reading the code it seems that the server will decide what the TZ should be, without having a look at `from flask_babel import get_locale` like you were mentioning @Glandos. I'm not sure what Babel does behind the scenes. So I believe we should check that the TZ used is dependent on the UA. Maybe we could add a test to prove this!

Also, it's not clear that it's what will happen from reading the docs, and I believe we should make it clearer. As the system administrator reading these docs, I'm not confident that the locale will be derived from the User-Agent.

I might add a big thanks for doing this :-)

Glandos:
Formatting date/time is always tricky on the web. Not really for technical reasons (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTimezoneOffset), but because it's difficult to know if the user prefers to display time in its own timezone, or from the server timezone. If it's on the user preference, we could have a user setting, or use JavaScript.

I don't really know what's best, but at least we will have one possibility here. So to me, we're good to go.

@zorun Have you been using this patch in production?

zorun:
I believe I already answered the question of which timezone is selected by default, in the documentation patch: it's the timezone of the server OS.  See also http://babel.pocoo.org/en/latest/dates.html#time-zone-support

Reading the patch again, I think it's OK to be merged even if it's not perfect:

- the timezone has a reasonable default (server OS) and can be configured if it's not right
- there is only a single place where we actually display a time: in the history
- I tested all the languages we have and Babel seems to adapt really well to each locale

Of course, it would be better to adapt to the user timezone, but it would be much more work. It does not seem worth the effort just to display a time in the history page.

Glandos:
Alright, let's stop trying to be perfect, we are not building a rocket full of uranium :)

indatwood:
Happy with this :-) Thanks for the work. I agree with you on this, let's try to ship more and worry less about being perfect ;-) Cheers!

PR #592, opened 2020-04-26
Fix crash when a localized email template is missing

zorun:
Currently we only have email templates for French and English.  Setting
the current locale to any other language generates an exception when
creating a project, because Flask can't find the template file.

Fix this by falling back to English if no language-specific template file
is found.


Glandos:
Thanks!

PR #594, opened 2020-04-26
Add support for different categories of "flash alerts"

zorun:
It is now possible to specify a category when calling `flash()` to change
the color of the alert box.  The default category is interpreted as
"success" for backwards compatibility.  For now, we only use the
additional "danger" category to show errors.

The available categories are the Bootstrap contextual classes listed here:

  https://getbootstrap.com/docs/4.0/components/alerts/


Glandos:
Thanks!

Issue #588, opened 2020-04-26
Prepare a bugfix-only 4.1.4 release?


zorun:
Currently, the latest release 4.1.3 is broken for new installs (#540).

The new 5.0.0 release (#570) might still take some time to happen (fix pending issues, wait for translators to pick up new strings, etc). In addition, it drops support for python 2 and python 3.5, so even in the long run some people may not be able to switch to 5.0.0 immediately.

Thus, I would like to propose a new "stable" 4.1.4 release, based on 4.1.3 instead of master, that fixes the following issues with 4.1.3 (checked items are already part of the [WIP branch](https://github.com/zorun/ihatemoney/tree/prepare_4_1_4)):

- [x] Fix failed installation because dependencies were not being pinned (#540, #545, #558)
- [X] Fix compatibility with werkzeug 1.0 (#488)
- [X] backend: Trim usernames to remove leading or trailing spaces. This avoids a situation where different names can be visually identical (#367)
- [X] backend: Fix API to forbid project creation when the `ALLOW_PUBLIC_PROJECT_CREATION` setting is set to false (#496)
- [X] backend: Fix crash when a localized email template is missing (#592)
- [X] backend: Fix language code parsing (#589)
- [x] backend: Improve error handling when sending emails (#595)
- [X] UI: Fix datepicker that was being displayed twice on some browsers (#221)
- [X] UI: Fix "Submit and add a new one" button that had no effect when adding a bill (#498)
- [X] UI: Prevent bill cancellation when cancelling autocomplete (#506)
- [X] UI: Fix responsive width of homepage on small screns (#549)
- [X] UI: Fix color of the "Add a member" button (#499)
- [X] UI: Fix missing HTML tag (#583)
- [X] UI: Fix a small typo in the french project-reminder email (#486)
- [X] UI: Fix typo on message displayed when adding a member (#575)
- [X] Add translations for German, Spanish, Norwegian, and Indonesian
- [x] Add translation for Russian
- [X] Update translations for all languages

Any other fix I may have missed?

What do you think on principle? @almet @JocelynDelalande @0livd  @eMerzh @Natim @Glandos  (I hope I didn't forget anybody)

eMerzh:
i  👍 that :) 
quickly checked and your summary cherry-picks are good for me.

zorun:
Cool :)

I have started preparing a WIP branch here: https://github.com/zorun/ihatemoney/tree/prepare_4_1_4

[...]

PR #595, opened 2020-04-26
Improve error handling when sending emails

zorun:
In one case, we were not catching a family of possible exceptions
(socket.error), and in the two other cases there was no error handling at
all.  Sending emails can easily fail if no email server is configured, so
it is really necessary to handle these errors instead of crashing with a
HTTP 500 error.

Refactor email sending code and add proper error handling.

Show alert messages that tell the user if an email was sent or if there
was an error.

When sending a password reminder email or inviting people by email, we
don't proceed to the next step in case of error, because sending emails is
the whole point of these actions.

### DIFF
--- a/ihatemoney/tests/tests.py
+++ b/ihatemoney/tests/tests.py
@@ -51,10 +51,10 @@ class BaseTestCase(TestCase):
             follow_redirects=True,
         )
 
-    def post_project(self, name):
+    def post_project(self, name, follow_redirects=False):
         """Create a fake project"""
         # create the project
-        self.client.post(
+        return self.client.post(
             "/create",
             data={
                 "name": name,
@@ -63,6 +63,7 @@ class BaseTestCase(TestCase):
                 "contact_email": f"{name}@notmyidea.org",
                 "default_currency": "USD",
             },
+            follow_redirects=follow_redirects,
         )
 
     def create_project(self, name):
--- a/ihatemoney/tests/tests.py
+++ b/ihatemoney/tests/tests.py
@@ -142,9 +142,14 @@ class BudgetTestCase(IhatemoneyTestCase):
             self.login("raclette")
 
             self.post_project("raclette")
-            self.client.post(
-                "/raclette/invite", data={"emails": "zorglub@notmyidea.org"}
+            resp = self.client.post(
+                "/raclette/invite",
+                data={"emails": "zorglub@notmyidea.org"},
+                follow_redirects=True,
             )
+
+            # success notification
+            self.assertIn("Your invitations have been sent", resp.data.decode("utf-8"))
 
             self.assertEqual(len(outbox), 2)
             self.assertEqual(outbox[0].recipients, ["raclette@notmyidea.org"])
@@ -226,7 +231,15 @@ class BudgetTestCase(IhatemoneyTestCase):
         self.create_project("raclette")
         # Get password resetting link from mail
         with self.app.mail.record_messages() as outbox:
-            self.client.post("/password-reminder", data={"id": "raclette"})
+            resp = self.client.post(
+                "/password-reminder", data={"id": "raclette"}, follow_redirects=True
+            )
+            # Check that we are redirected to the right page
+            self.assertIn(
+                "A link to reset your password has been sent to you",
+                resp.data.decode("utf-8"),
+            )
+            # Check that an email was sent
             self.assertEqual(len(outbox), 1)
             url_start = outbox[0].body.find("You can reset it here: ") + 23
             url_end = outbox[0].body.find(".\n", url_start)
@@ -251,17 +264,26 @@ class BudgetTestCase(IhatemoneyTestCase):
     def test_project_creation(self):
         with self.app.test_client() as c:
 
-            # add a valid project
-            c.post(
-                "/create",
-                data={
-                    "name": "The fabulous raclette party",
-                    "id": "raclette",
-                    "password": "party",
-                    "contact_email": "raclette@notmyidea.org",
-                    "default_currency": "USD",
-                },
-            )
+            with self.app.mail.record_messages() as outbox:
+                # add a valid project
+                resp = c.post(
+                    "/create",
+                    data={
+                        "name": "The fabulous raclette party",
+                        "id": "raclette",
+                        "password": "party",
+                        "contact_email": "raclette@notmyidea.org",
+                        "default_currency": "USD",
+                    },
+                    follow_redirects=True,
+                )
+                # an email is sent to the owner with a reminder of the password
+                self.assertEqual(len(outbox), 1)
+                self.assertEqual(outbox[0].recipients, ["raclette@notmyidea.org"])
+                self.assertIn(
+                    "A reminder email has just been sent to you",
+                    resp.data.decode("utf-8"),
+                )
 
             # session is updated
             self.assertTrue(session["raclette"])
--- a/ihatemoney/tests/tests.py
+++ b/ihatemoney/tests/tests.py
@@ -53,7 +53,7 @@ class BaseTestCase(TestCase):
             follow_redirects=True,
         )
 
-    def post_project(self, name, follow_redirects=False):
+    def post_project(self, name, follow_redirects=True):
         """Create a fake project"""
         # create the project
         return self.client.post(
--- a/ihatemoney/tests/tests.py
+++ b/ihatemoney/tests/tests.py
@@ -2296,13 +2296,13 @@ class EmailFailureTestCase(IhatemoneyTestCase):
         with patch.object(
             self.app.mail, "send", MagicMock(side_effect=smtplib.SMTPException)
         ):
-            resp = self.post_project("raclette", follow_redirects=True)
+            resp = self.post_project("raclette")
         # Check that an error message is displayed
         self.assertIn(
             "We tried to send you an reminder email, but there was an error",
             resp.data.decode("utf-8"),
         )
-        # Check that we are redirected anyway
+        # Check that we were redirected to the home page anyway
         self.assertIn(
             'You probably want to <a href="/raclette/members/add"',
             resp.data.decode("utf-8"),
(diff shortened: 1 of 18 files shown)


zorun:
Any feedback on this?

I think it should also be backported to a stable release (see #588 where I started collecting fixes)

ihatemoney/utils.py
```diff
@@ -7,9 +7,11 @@
 import operator
 import os
 import re
+from smtplib import SMTPRecipientsRefused
```
indatwood, line 10: Why using this exception rather than the more generic [`SMTPException`](https://docs.python.org/3/library/smtplib.html#smtplib.SMTPException)?

ihatemoney/web.py
```diff
@@ -327,16 +327,18 @@ def remind_password():
...
+            remind_message = Message(
+                "password recovery",
+                body=render_localized_template("password_reminder", project=project),
+                recipients=[project.contact_email],
             )
-            return redirect(url_for(".password_reminder_sent"))
+            result = send_email(
+                remind_message,
+                _("A password reminder email has just been sent to you"),
+                _("Error while sending password reminder email"),
```
indatwood, line 338: I wonder about the message here. Should we be more user friendly, with a message like "Sorry, an error occured when we tried to send you a reminder email, try again and if the problem persist, try contacting the instance administrators", or something like it.

It's longer, for sure, but I feel it makes things more human :-)

ihatemoney/utils.py
zorun: Good question, this is code that I just moved around.

Looking at it, `SMTPException` is a subclass of `OSError`, and `socket.error` is actually equal to `OSError`.  Since I already check for `socket.error` it's redundant anyway...

I will look at what happens with older versions of Python (to ease backporting) and fix this mess, thanks for the review.

ihatemoney/web.py
zorun: User-facing messages are always the hardest part :) I will try to improve it.

zorun:
I have tweaked exception handling and improved error messages. However, the messages shouldn't be too long, because they are only displayed for 4 seconds before disappearing... This is something we probably want to fix in general.

Regarding tests, I agree it would be nice, but do you have any suggestion on how to do it? Just assume that sending email will fail during the test and check that the app doesn't crash?

indatwood:
> I have tweaked exception handling and improved error messages. However, the messages shouldn't be too long, because they are only displayed for 4 seconds before disappearing... This is something we probably want to fix in general.

Yes I agree, they are too short right now. 

> 
> Regarding tests, I agree it would be nice, but do you have any suggestion on how to do it? Just assume that sending email will fail during the test and check that the app doesn't crash?

you can mock the lib that is sending the emails to throw an error and ensure the app reacts appropriately!

zorun:
Some progress report on the tests: I had a look at flask-email but it doesn't seem to support "simulating" failures.

That being said, I don't actually know what "mocking" is, so I may have missed something obvious...

almet:
Here is some information about mocks : https://docs.python.org/3/library/unittest.mock.html#the-mock-class, I think that's the way to go :-)

zorun:
So, I had a hard time getting this to work, but I finally succeeded in mocking the email sending function!

Basically it looks like this:

```
self.create_project("raclette")
with patch.object(self.app.mail, "send", MagicMock(side_effect=smtplib.SMTPException)):
    resp = self.client.post("/password-reminder", data={"id": "raclette"}, follow_redirects=True)
# Check the response
```

I couldn't get anything to work by patching the class `ihatemoney.run.Mail`, possibly because a real `Mail()` object has already been created by the time the test is run.

I will add more tests and update the PR.

indatwood:
That looks great! Let me know when the code is updated, I'll do another review.

Thanks :-)

zorun:
@indatwood cool, thanks for the offer!  The PR is now ready for a second round of review (especially the tests).

ihatemoney/tests/tests.py
```diff
@@ -63,6 +65,7 @@ def post_project(self, name):
                 "contact_email": f"{name}@notmyidea.org",
                 "default_currency": "USD",
             },
+            follow_redirects=follow_redirects,
```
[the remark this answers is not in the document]
zorun: Good question. Before my changes, no call to `post_project` was using the result. So it should make no difference whether redirections are followed or not.

After looking around a bit, a use-case for **not** following redirects is to test where the page redirects to (see https://stackoverflow.com/a/25354698). We currently do that by checking the content of the final page, which is indeed brittle. But fixing that is clearly outside the scope of this PR.

Given this, I can make `follow_redirects=True` the default for `post_project` but I think it makes sense to keep the argument.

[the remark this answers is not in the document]
zorun: Actually, the "clean" method to check for redirects doesn't work in our cases (at least the one I checked), because we also need to test that a success/error message is correctly displayed, and that involves following the redirection...

Anyway, I have switched `follow_redirects=True` by default, it should make things simpler.

zorun:
Thanks for the review!

ihatemoney/utils.py
```diff
@@ -28,6 +30,28 @@ def slugify(value):
...
+    the email or not.
+    """
+    # Since Python 3.4, SMTPException and socket.error are actually
+    # identical, but this was not the case before.  Also, it is more clear
+    # to check for both.
+    try:
+        current_app.mail.send(mail_message)
+    except (smtplib.SMTPException, socket.error):
+        if flash_error:
+            flash(flash_error, category="danger")
```
Glandos, line 47:
```suggestion
        if flash_error:
            flash(**{"category": "danger", **(flash_error if isintance(flash_error, dict) else {"message": flash_error})})
```

ihatemoney/utils.py
```diff
@@ -28,6 +30,28 @@ def slugify(value):
...
+        return False
+    # Email was sent successfully
+    if flash_success:
+        flash(flash_success, category="success")
```
Glandos, line 51:
```suggestion
    if flash_success:
        flash(**{"category": "success", **(flash_success if isintance(flash_success, dict) else {"message": flash_success})})
```

ihatemoney/web.py
```diff
@@ -308,11 +308,17 @@ def create_project():
...
+            if not success:
+                # Display the error as a simple "info" alert, because it's
+                # not critical and doesn't prevent using the project.
+                flash(
+                    _(
+                        "We tried to send you an reminder email, but there was an error. "
+                        "You can still use the project normally."
+                    ),
+                    category="info",
+                )
```
Glandos, line 321:
```suggestion
            send_email(msg, _("A reminder email has just been sent to you"), {"message": _(
                        "We tried to send you an reminder email, but there was an error. "
                        "You can still use the project normally."
                    ), "category": "info")
```

ihatemoney/web.py
```diff
@@ -328,16 +334,23 @@ def remind_password():
...
-            return redirect(url_for(".password_reminder_sent"))
+            success = send_email(
+                remind_message,
+                "",
+                _(
+                    "Sorry, there was an error while sending you an email "
+                    "with password reset instructions. "
+                    "Please check the email configuration of the server "
+                    "or contact the administrator."
+                ),
```
Glandos, line 350:
```suggestion
                flash_error=_(
                    "Sorry, there was an error while sending you an email "
                    "with password reset instructions. "
                    "Please check the email configuration of the server "
                    "or contact the administrator."
                ),
```

ihatemoney/utils.py
```diff
@@ -28,6 +30,28 @@ def slugify(value):
     return re.sub(r"[-\s]+", "-", value)
 
 
+def send_email(mail_message, flash_success="", flash_error=""):
+    """Send an email using Flask-Mail, returning False if there was an error.
+
+    Optionally display a "flash alert" message to the user.  The flash
+    message is different depending on whether we could successfully send
+    the email or not.
```
Glandos, line 38:
```suggestion
    the email or not.
    
    flash_success and flash_error can be either a simple string flashed with respectively success and error category, or a dict with all parameters as kwargs for flash() function, e.g. {"message": "Alert", "category": "danger"}
```

zorun:
@Glandos err, thanks for the suggestion, but this looks quite complex, including in its usage...

Now that I think about it again, it may be better to just move flashing out of this utility function, and only keep `send_mail()` for, well, sending emails (as its name implies) with proper error checking.

The first idea was that displaying a success / error message is a common pattern when sending emails. But in the three places where we send emails, we already have one special case, so maybe it's not that common...

So, I would just check the return value of `send_email()` and flash a message accordingly as needed by the local context.

What do you think?

Glandos:
I also thought about it. I have to admit that I like to play with Python's dynamic way of interpretation (e.g. `kwargs`), so I had fun :smile: 

But you're right: either we have a clear pattern that displays flash message, or we don't have it. Keep the `send_email` clear of `flash` calls, as you said, and it will be better.

zorun:
Great then :)  I have pushed the simpler solution.  Feel free to merge if it's good for you.

ihatemoney/web.py
```diff
@@ -328,17 +338,25 @@ def remind_password():
...
+            success = send_email(remind_message)
+            if success:
+                return redirect(url_for(".password_reminder_sent"))
+            else:
+                flash(
+                    _(
+                        "Sorry, there was an error while sending you an email "
+                        "with password reset instructions. "
+                        "Please check the email configuration of the server "
+                        "or contact the administrator."
```
Glandos, line 355: What do you think of writing the email address from `MAIL_DEFAULT_SENDER`?
It *should* be reliable, but not always. However, I am always puzzled by message that says to contact the administrator without a clue on who it can be.
Maybe we can add:
> contact the administrator. Try send an email to `MAIL_DEFAULT_SENDER[1]`.

zorun: Hmm, good point. For I hate money I expect users would know their administrators (small instances), but this may not always be the case.

I'm a bit hesitant because it reveals some (potentially private) information. True, this information is used as `From:` for outgoing emails, but still.

Maybe only do that if public project creation is allowed? Also, we should avoid showing the email address if it's budget@notmyidea.org ;)

In any case, I suggest doing that in another PR. The initial goal of this one was to fix a crash, and it is the last blocking point before going ahead with the stable release (#588)

zorun: See #626

Issue #597, opened 2020-04-28
Point to playstore/fdroid app rather than nerdy git


JocelynDelalande:
The smartphone icon points to https://gitlab.com/eneiluj/moneybuster

This is not nice for « simple users ». As, on earth, « simple users » are more numerous than « app developpers », maybe we could point to something more user-friendly :-)

Abo7atm:
Is anyone working on this? I'd like to work on it. Please give me a little explanation about this issue.

Glandos:
![Screenshot_2020-04-29 Account manager](https://user-images.githubusercontent.com/297578/80640181-545d0880-8a63-11ea-9f86-ab63e97cf7ff.png)
The red icon at the bottom is a link to a smartphone application. Right now, the link is pointing to the source code, and for a better user experience, it should point to the application store where the user can install the application in one click.

MagicFab:
A bigger dowload button at the top linking to a download page pointing to F-Droid, Play Store and Apple Store would be more intuitive IMO. 

See this as an example : https://jitsi.org/#download

JocelynDelalande:
Thanks for suggestion @MagicFab ; I have been thinking about the big download button and I think this is too much clutter for regular users.
It is as like all those users-hungry websites which choose to show subscription form at 1-click distance but login form at 2-click distance.

Those are patterns prioritizing agressively new user adhesion over regular users comfort.

I'd like we take care about regular users, we are not running after users or state, but making convivial tool (or I hope so :-)).

Suggestion :

## mobile apps Icons (bubles)

Replacing the smartphone icon with two bubble icons : 

- one f-droid icon
- one google-play

Those being direct links to app pages in the stores.

Those icons are targetted at desktop users, to let them know that app exist, not for android users (except as fallback), see after why…

## Adding Meta tag

- adding a `<meta>` tag to indicate semantically the existence of the app https://developers.facebook.com/docs/applinks/metadata-reference AFAIK, the effect of that is prompting the user if s·he wants to install the app on first time he comes on the website.

Any opinions ?

MagicFab:
I understand your comment about looking hungry for users. It's the first thumb down I've earned ! :smile:  At the same time keep in mind the bottom of the page is probably the last place people will look for a download option. 

When you land on https://ihatemoney.org/ it looks like a web app (**only**). I maintain that a single "Download the mobile app" or similar clear option at the top would benefit adoption. Linking to a dedicated download page is important so the apps can be presented by themselves.

[...]

PR #598, opened 2020-04-29
Change mobile link icon to point to Play Store (#597)

Abo7atm:


eMerzh:
not sure we want to do that 🤔 .
the idea here was to link to the "project" of the mobile app where scrolling a bit might link you to the download link you want, but money buster prefers (i think) f-droid to play store.

zorun:
We should definitely add a link to the F-Droid app, and maybe also another link to the play store. Maybe the icons could be adapted?

Glandos:
> not sure we want to do that thinking .
> the idea here was to link to the "project" of the mobile app where scrolling a bit might link you to the download link you want, but money buster prefers (i think) f-droid to play store.

I agree @eMerzh but on the other hand, the goal is to make a smooth experience here.

Maybe we can add a dropdown with the 2 options, shown when clicking the link? But it's clearly more work.

zorun:
The most user-friendly way is to add two links with appropriate icons: one for F-Droid and one for the Play Store. But indeed we may want to link to the "project" instead.

In that case, we could at least use a more user-friendly link:

- https://gitlab.com/eneiluj/moneybuster#moneybuster-for-android : already better than the current link (it starts with a short description of the app, and if you scroll a bit you quickly see the big F-Droid and Play Store buttons)
- https://gitlab.com/eneiluj/moneybuster/-/wikis/home#releases : direct access to the packages, we support the project because there's a Donate button, and there are screenshots. However, it's missing a description of the app

indatwood:
> The most user-friendly way is to add two links with appropriate icons: one for F-Droid and one for the Play Store.

I second this :-)

Abo7atm:
Interesting discussion. I get, now, what's intended from the mobile app button.

> The most user-friendly way is to add two links with appropriate icons: one for F-Droid and one for the Play Store.

This might be a more user-friendly option, but, as mentioned above, the intention of using this button is to link to the project itself, while still keeping the nerdy-ness to a minimum. 

> * https://gitlab.com/eneiluj/moneybuster#moneybuster-for-android : already better than the current link (it starts with a short description of the app, and if you scroll a bit you quickly see the big F-Droid and Play Store buttons)

I think this might be the better choice.

eMerzh:
@Abo7atm thanks for the modifcations :)

lgtm 👍 ... we might think in the future about having a dedicated page or smth for the integrations...

eMerzh:
btw we might ask money buster if they'd like to create a nice welcome page

indatwood:
@eneiluj do you have a preference?

almet:
Thanks ! Should we point to the « install » buttons directly instead? with this link https://gitlab.com/eneiluj/moneybuster#install

Natim:
Yes I thought it was the case already but that's definitely a good idea

Le ven. 8 mai 2020 à 23:03, Alexis Metaireau <notifications@github.com> a
écrit :

JocelynDelalande:
> Thanks ! Should we point to the « install » buttons directly instead? with this link https://gitlab.com/eneiluj/moneybuster#install

Not sure, I tend to avoid directing users to README, even if it is on the right anchor. IMHO We have to stop thinking people feel safe/welcomed in a gitlab/github page. Those are complex interfaces intended for developers. That is what those interfaces are good at.

indatwood:
If we don't feel like putting a link on the readme (even with the right anchor) then we might want to create a page on the ihatmeoney website / project so that users can find the different options :-)

indatwood:
What should we do with this? My insight would be to merge it but still create a new page linking to clients.

Glandos:
It's clearly better than the current situation, even if not ideal.
@JocelynDelalande I tried the new link with a "mobile" screen (using Firefox responsive feature), and it doesn't look as nerdy as on desktop screen. Do you think we should skip this request or keep it?

almet:
Hey, I believe we're good to go here :-) We can still add new enhancements in the future if needed! Let's merge?

zorun:
It's not perfect but it's better than before, so let's merge!

Issue #600, opened 2020-04-30
Currency: duplicate information in "Amount" column when using no currency


zorun:
When using no currency, the new "Amount" columns only shows duplicate information:

![ihm-duplicate-amount](https://user-images.githubusercontent.com/731144/80704272-9712f500-8ae4-11ea-993b-fc9a39faaac5.png)

I see a few alternatives to improve it:

- simply remove the column when no currency is used
- only keep "21.00 each" in the "How much?" column, and move the total amount to the "Amount" column. This may be confusing when having currencies though.

PR #604, opened 2020-05-02
Improve currencies

### LINKED ISSUE #601: Currency: failure with postgresql: "value too long for type character varying(3)"
The currency is declared as a string of length 3:

```
# Project
default_currency = db.Column(db.String(3))
# Bill
original_currency = db.Column(db.String(3))
```
However, an empty currency is encoded as the string `No Currency`, which is longer than 3 characters.

Strangely enough, sqlite doesn't seem to care, and happily stores longer values:

```
sqlite> SELECT * FROM bill;
1|1|42.0|2020-04-30|Test||2020-04-30||42.0|No Currency
2|2|100.0|2020-04-30|Bill en euros||2020-04-30||100.0|EUR
```

But with postgresql, it is not even possible to apply the migration, because the default value is larger than the maximum length:

```
INFO  [alembic.runtime.migration] Running upgrade cb038f79982e -> 927ed575acbd, Add currencies
  File "/home/zorun/tmp/venv3-ihatemoney/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1247, in _execute_context
    self.dialect.do_execute(
  File "/home/zorun/tmp/venv3-ihatemoney/lib/python3.8/site-packages/sqlalchemy/engine/default.py", line 590, in do_execute
    cursor.execute(statement, parameters)
psycopg2.errors.StringDataRightTruncation: value too long for type character varying(3)

The above exception was the direct cause of the following exception:

sqlalchemy.exc.DataError: (psycopg2.errors.StringDataRightTruncation) value too long for type character varying(3)

[SQL: ALTER TABLE bill ADD COLUMN original_currency VARCHAR(3) DEFAULT 'No Currency']
```

Glandos:
Improve #541 by adding the following:
- Rename "No Currency" to ISO4217 "XXX"
- Use Babel to render currency symbols and names in currency lists
- Improve i18n in bill lists

Fix #601 #600 

I discovered Babel number formatting functions, and they are awesome. They are wrapped by flask_babel, undocumented as far as I know.

### DIFF
--- a/ihatemoney/forms.py
+++ b/ihatemoney/forms.py
@@ -23,7 +23,7 @@ from wtforms.validators import (
 
 from ihatemoney.currency_convertor import CurrencyConverter
 from ihatemoney.models import LoggingMode, Person, Project
-from ihatemoney.utils import eval_arithmetic_expression, slugify
+from ihatemoney.utils import eval_arithmetic_expression, slugify, render_localized_currency
 
 
 def strip_filter(string):
@@ -31,18 +31,6 @@ def strip_filter(string):
         return string.strip()
     except Exception:
         return string
-
-
-def get_editprojectform_for(project, **kwargs):
-    """Return an instance of EditProjectForm configured for a particular project.
-    """
-    form = EditProjectForm(**kwargs)
-    choices = copy.copy(form.default_currency.choices)
-    choices.sort(
-        key=lambda rates: "" if rates[0] == project.default_currency else rates[0]
-    )
-    form.default_currency.choices = choices
-    return form
 
 
 def get_billform_for(project, set_default=True, **kwargs):
@@ -124,7 +112,7 @@ class EditProjectForm(FlaskForm):
     default_currency = SelectField(
         _("Default Currency"),
         choices=[
-            (currency_name, currency_name)
+            (currency_name, render_localized_currency(currency_name))
             for currency_name in currency_helper.get_currencies()
         ],
         validators=[DataRequired()],
--- a/ihatemoney/forms.py
+++ b/ihatemoney/forms.py
@@ -279,7 +279,9 @@ class BillForm(FlaskForm):
         self.original_currency.label = Label("original_currency", _("Currency"))
         self.original_currency.description = _(
             "Project default: %(currency)s",
-            currency=render_localized_currency(project.default_currency, detailed=False),
+            currency=render_localized_currency(
+                project.default_currency, detailed=False
+            ),
         )
 
     def set_default(self):
--- a/ihatemoney/forms.py
+++ b/ihatemoney/forms.py
@@ -45,19 +45,8 @@ def get_billform_for(project, set_default=True, **kwargs):
         form.original_currency.data = project.default_currency
 
     if form.original_currency.data != CurrencyConverter.default:
-        choices = copy.copy(form.original_currency.choices)
-        choices.remove((CurrencyConverter.default, CurrencyConverter.default))
-        choices.sort(
-            key=lambda rates: "" if rates[0] == project.default_currency else rates[0]
-        )
-        form.original_currency.choices = choices
-    else:
-        form.original_currency.render_kw = {"default": True}
-        form.original_currency.data = CurrencyConverter.default
+        form.original_currency.choices.remove((CurrencyConverter.default, render_localized_currency(CurrencyConverter.default, detailed=False)))
 
-    form.original_currency.label = Label(
-        "original_currency", "Currency (Default: %s)" % (project.default_currency)
-    )
     active_members = [(m.id, m.name) for m in project.active_members]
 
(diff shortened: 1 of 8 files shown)


ihatemoney/templates/list_bills.html
```diff
@@ -1,5 +1,10 @@
 {% extends "sidebar_table_layout.html" %}
 
+{%- macro bill_amount(bill, currency=bill.original_currency, amount=bill.amount) %}
+    {# Undocumented currencyformat filter from flask_babel is forwarding to Babel format_currency #}
+    {{ amount|currencyformat(currency if currency != 'XXX' else '') }} ({{ _("%(amount)s each", amount=bill.pay_each_default(amount)|currencyformat(currency if currency != 'XXX' else '')) }})
+{% endmacro -%}
+
```
zorun, line 7: Nice that you found out about this `currencyformat` filter!

I think it makes sense to write a small wrapper around it to handle the `XXX` currency, because that is something we want to do everywhere.

Also, small nitpick: with no currency, we still get non-breakable spaces for some locales:
```
In [1]: numbers.format_currency("42", "", locale="fr")
Out[1]: '42,00\xa0'
In [2]: numbers.format_currency("42", "", locale="no")                                                                                                       
Out[2]: '\xa042,00'
```
This looks slightly bad in the list of bills because there is too much space:
```
12,00  (6,00  each)
```

ihatemoney/forms.py
```diff
@@ -124,7 +109,7 @@ class EditProjectForm(FlaskForm):
     default_currency = SelectField(
         _("Default Currency"),
         choices=[
-            (currency_name, currency_name)
+            (currency_name, render_localized_currency(currency_name))
```
zorun, line 112: Maybe add `detailed=True` for clarity.

ihatemoney/forms.py
```diff
@@ -245,7 +230,7 @@ class BillForm(FlaskForm):
     original_currency = SelectField(
         _("Currency"),
         choices=[
-            (currency_name, currency_name)
+            (currency_name, render_localized_currency(currency_name, detailed=False))
```
zorun, line 233: Is there anything that prevents using the detailed format here?

If we switch to something like select2 or equivalent, it would allow to search by name (e.g. "zloty").

That being said, I agree that this field will rarely be used.

zorun:
Here are some additional suggestions:

- use the currency format everywhere (stats, settle, sidebar)
- add a tooltip explaining the currency setting when creating a project. It should explain that if all bills are expected to use the same currency, selecting "No currency" is fine; a default currency should only be selected if you plan to add bills in several currencies.
- improve wording? Some suggestions: "Single currency" instead of "No currency". "Project currency" or "Base currency" instead of "Default currency". This is the hardest part ;)

ihatemoney/forms.py
```diff
@@ -297,6 +282,14 @@ def fill(self, bill):
...
         self.payed_for.data = [int(ower.id) for ower in bill.owers]
 
+        if bill.original_currency != project.default_currency:
+            label = _(
+                "Currency (Default: %(currency)s)", currency=project.default_currency
+            )
+        else:
+            label = _("Currency")
+        self.original_currency.label = Label("original_currency", label)
+
```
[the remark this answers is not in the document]
Glandos: I add a description instead. It should be clearer, tell me what you think.

ihatemoney/templates/list_bills.html
Glandos: It's a side-effect of currency formatting, since there is a non-breaking space before or after currency sign.
I crafted a new custom filter to handle this use case. I named it differently to avoid overwriting, but naming it was… difficult :)

ihatemoney/forms.py
Glandos: When a user is selecting a currency for one bill, he/she should be aware of the currency, and I don't think there is a need to clutter the UI with a lot of text. We will see if we use a filtering select widget in the future.

Glandos:
[**@zorun**](https://github.com/zorun) commented on [3 mai 2020 à 23:32 UTC+2](https://github.com/spiral-project/ihatemoney/pull/604#issuecomment-623185683 "2020-05-03T21:32:02Z - Replied by Github Reply Comments"):
> Here are some additional suggestions:
> 
> - use the currency format everywhere (stats, settle, sidebar)

That will be for another request. This one is already too big.

> - add a tooltip explaining the currency setting when creating a project. It should explain that if all bills are expected to use the same currency, selecting "No currency" is fine; a default currency should only be selected if you plan to add bills in several currencies.

You're right, but same as above.

> - improve wording? Some suggestions: "Single currency" instead of "No currency". "Project currency" or "Base currency" instead of "Default currency". This is the hardest part ;)

I'll favor "Currency", because we already are in a "Project" form, and the explanation from the point above should do give a better explanation about it.

Issue #501, opened 2019-10-19
Split test into little ones.


JoseAntonioTorre:
I have been reading the code of test and I have seen that, for example, in test_project we are asserting different use cases. In my opinion, it is better to divide the test and refactor it in a little methods.

What is your opinion?
For example, instead of testing that first we can create a project, later we can not create the same, it is better to have a test to create a project, a test to test that we can not create a duplicated one, and so on.

almet:
It's probably better to have atomic test functions, because it's easier to detect where the problem is when it arises. I believe this was done this way because it was easier at that time.

For me the main important thing is to have a coverage of all the scenarios. If you feel like splitting every scenario into multiple (atomic) tests, do not hesitate :-)

JoseAntonioTorre:
I think that it is better, yes!, I will try to start working on it and let you know it!

JvPy:
I would like to contribute.

Never worked with tests before, but would like to learn, if possible.

almet:
@JoseAntonioTorre would you like to mentor @JvPy on this?

JoseAntonioTorre:
Yes, I am not an expert in testing but I can try it!
@JvPy First, get used to read the testing code and try to figure out what is happening in each test. After that, you will probably see that there are tests that asserts several things that are not related to the test itself. The idea is to divide those tests in little ones in order to make it easier to check where are the errors (in case we have any error...)

indatwood:
Hi @JoseAntonioTorre and @JvPy. Are you still happy to work on this?

Villaquiranm:
@indatwood @Glandos  Hello, could you give more information about this issue like the most problematic files / test to refactor ?

Thanks

Glandos:
Well @Villaquiranm I'm glad you asked. The few tests I've written were difficult, because it seems to me that a lot of code has to be duplicated, especially for project and bill creation.
The `BudgetTestCase` is clearly too big, and should at least be in its own file.

So as a start, it should be nice if each "real" test case class were in their own file. But it should be done with code factorization in mind, in order to avoid repeating a lot of code. For example, some datasets like bill could be gathered and available in each test (I've spotted the `"id": "raclette"` too many times :yum: )

Villaquiranm:
@Glandos Ok, i will try to do something nice :)

Villaquiranm:
Hello @Glandos , I expect this PR is more or less what you expected, if you have any suggestion to improve it please let me know i'll be glad to continue coding

PR #608, opened 2020-05-04
 Fix order of participants in the statistics page

### LINKED ISSUE #607: Statistics are mixed up between participants
I noticed that the statistics (Paid/Spent) are sometimes attributed to the wrong participant.

Here is an example matching [the current test](https://github.com/spiral-project/ihatemoney/blob/15ab04e636f1c267da3ef1c6bd4cdc68449b2e97/ihatemoney/tests/tests.py#L897) but with simple name changes (fred → zorglub, tata → quentin, toto → manu, you'll see why):

ihm-bug-statistics-1

Something is wrong because Paid - Spent is not always equal to the balance.

Here is the same view but displaying the hidden column on the right-side table (this column is normally only shown on small screens because the two tables are not side-by-side):

ihm-bug-statistics-2

The origin of the bug is clear: the two tables are not ordered in the same way! It's been like this since the beginning (#257) but only became an issue when the tables were "merged" in #535.

The [test](https://github.com/spiral-project/ihatemoney/blob/15ab04e636f1c267da3ef1c6bd4cdc68449b2e97/ihatemoney/tests/tests.py#L897) failed to catch this for two reasons:

- it adds members in alphabetical order, so ordering by ID or by name gives the same result;
- in addition, the test is using the hidden column to check that the amounts are correct for the given participant!

zorun:
This fixes #607 and add a test case for this bug.

It also renames participants in test cases to avoid alphabetical ordering.

Inserting participants in alphabetical order is a special case, because ordering by ID will be the same as ordering by name.  This is a bad idea in test cases, as #607 has shown.

### DIFF
--- a/ihatemoney/templates/statistics.html
+++ b/ihatemoney/templates/statistics.html
@@ -9,7 +9,7 @@
             <th class="balance-value">{{ _("Balance") }}</th>
         </tr>
     </thead>
-    {% for stat in members_stats| sort(attribute='member.name') %}
+    {% for stat in members_stats|sort(attribute='member.name') %}
     <tr>
         <td class="balance-name">{{ stat.member.name }}</td>
         <td class="balance-value {% if stat.balance|round(2) > 0 %}positive{% elif stat.balance|round(2) < 0 %}negative{% endif %}">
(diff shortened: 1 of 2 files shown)


Glandos:
Thanks for that work. The fix was easy, tests were awfully hard. I hope we'll have time to work on #501 some day…

PR #616, opened 2020-05-09
Populate the demo project with defaults.

### LINKED ISSUE #609: Rethink how the demo project works
Currently, the demo project is just a special project that we reset now and then when it's too cluttered.

I think of a least one better ways this could be :

Have a demo project which is unique to each user, generated on the fly and destructed a few hours later. On this project members are already created and a few expenses are there.

We might want to display something special for this project so it's not thought about a "real" one.

What do you think?

indatwood:
Fixes #609 by adding a few members and bills to the default test project.

### DIFF
--- a/ihatemoney/models.py
+++ b/ihatemoney/models.py
@@ -10,7 +10,6 @@ from itsdangerous import (
     TimedJSONWebSignatureSerializer,
     URLSafeSerializer,
 )
-
 import sqlalchemy
 from sqlalchemy import orm
 from sqlalchemy.sql import func
--- a/ihatemoney/models.py
+++ b/ihatemoney/models.py
@@ -10,11 +10,13 @@ from itsdangerous import (
     TimedJSONWebSignatureSerializer,
     URLSafeSerializer,
 )
+
 import sqlalchemy
 from sqlalchemy import orm
 from sqlalchemy.sql import func
 from sqlalchemy_continuum import make_versioned, version_class
 from sqlalchemy_continuum.plugins import FlaskPlugin
+from werkzeug.security import generate_password_hash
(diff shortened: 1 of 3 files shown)


Glandos:
What do you think about using the `import_project` function within `web.py`, using a JSON file?
It can even allow the JSON to be customized within preference file.

indatwood:
Who would use this, what is the use case? Relying on an external file seems more complicated for little gain to me :-)

Glandos:
I was thinking about a new setting in the setting file:
```python
# If set to True, a demonstration project will be activated.
ACTIVATE_DEMO_PROJECT = True

# Fill demonstration project when created automatically
DEMO_PROJECT_INITIAL_DATA = [{'what': 'Bill 1',
  'amount': 11.0,
  'date': '2020-05-03',
  'payer_name': 'User1',
  'payer_weight': 1.0,
  'owers': ['User1', 'User2']},
 {'what': 'Bill 2',
  'amount': 12.0,
  'date': '2020-05-03',
  'payer_name': 'User1',
  'payer_weight': 1.0,
  'owers': ['User1', 'User2']}]
```

But I have to admit that the use case is quite limited :+1:

indatwood:
I see many ways how exposing this via configuration can be error prone or create bugs, for a feature a believe would be not that useful. 

Frankly, I prefer to keep things simple if you don't mind, that's why I like this project : it doesn't try to do too much :-)

Glandos:
OK, go for it :) If someone wants to make it optional, we'll just add a settings to skip data insertion.

You just have to fix CI by running "make black isort".

indatwood:
Updated.

PR #618, opened 2020-05-10
Add a unified balance table for all sidebar pages

Glandos:
We also use the currencyformat_nc to show currency formatting everywhere

### DIFF
--- a/ihatemoney/templates/history.html
+++ b/ihatemoney/templates/history.html
@@ -98,23 +98,7 @@
 
 {% block sidebar %}
     <div id="table_overflow">
-    <table class="balance table">
-    <thead>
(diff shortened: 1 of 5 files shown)


indatwood:
That looks okay to me. That being said, it seems that we could also use an `{% include %}` statement with parameters passed with `{% with %}`, or directly when calling the template.

But this would probably be a big change, it seems that the project uses a lot of macros so far, so we might want to continue this way, just for consistency.

Glandos:
You're right about using `include / with`, but in this case, since we need arguments with default values, I guess it's easier with macros.

If you're OK with this, I'd like to merge it by the end of this week-end.

indatwood:
Go for it!

PR #619, opened 2020-05-10
use currency format everywhere

Glandos:
This should unify the number formats, along with #618

### DIFF
--- a/ihatemoney/templates/settle_bills.html
+++ b/ihatemoney/templates/settle_bills.html
@@ -25,7 +25,7 @@
     <tr receiver={{bill.receiver.id}}>
             <td>{{ bill.ower }}</td>
             <td>{{ bill.receiver }}</td>
-            <td>{{ "%0.2f"|format(bill.amount) }}</td>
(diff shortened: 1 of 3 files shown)


ihatemoney/templates/settle_bills.html
```diff
@@ -25,7 +25,7 @@
     <tr receiver={{bill.receiver.id}}>
             <td>{{ bill.ower }}</td>
             <td>{{ bill.receiver }}</td>
-            <td>{{ "%0.2f"|format(bill.amount) }}</td>
+            <td>{{ bill.amount|currencyformat_nc(g.project.default_currency) }}</td>
```
indatwood, line 28: What do you think about creating a shortcut filter so we could do `bill.amount | currency` for instance?

Glandos: I'll try to have a look at it.

The currency filter should then be able to format without parameter (and use project's default), or accept a currency as parameter (for bill list).

zorun:
Looks good to me, combined with #625  :+1:

Issue #610, opened 2020-05-04
The private code is no longer stored or sent in cleartext


zorun:
Since a few years already (#286), we hash a project password ("private code") and don't store it in clear text. I just looked and we also don't seem to send it by email.

However, the tooltip on the home page (`ihatemoney/templates/home.html`) still says:

> This access code will be sent to your friends. It is stored as-is by the server, so don't reuse a personal password!

This text should be changed. The code is used for connecting to the project and to change its settings. It is never sent to anybody and is no longer stored "as-is" by the server. It is possible to invite people to a project without telling them the private code, this should probably be hinted here.

Also, we call that "private code" everywhere else, not "access code".

rajat641:
Hi @zorun Can I work on this ??

Glandos:
@rajat641 Yes, you can, nobody is currently working on this. Thanks for taking care of it :+1:

rajat641:
Hi @Glandos @zorun ,
I have created a PR for this fix. Request to have a look. Here is the PR link- https://github.com/spiral-project/ihatemoney/pull/623

comradekingu:
@rajat641 "Make an access code and send it to your friends"?

PR #623, opened 2020-05-19
Fix: Change tool-tip message for access code

rajat641:
This is a fix for Issue [#610 ](https://github.com/spiral-project/ihatemoney/issues/610)
I have updated the message and tested locally.

### DIFF
--- a/ihatemoney/templates/home.html
+++ b/ihatemoney/templates/home.html
@@ -91,7 +91,7 @@
 {% endblock %}
 {% block js %}
 $('#creation-form #password').tooltip({
-    title: '{{ _("This access code will be sent to your friends. It is stored as-is by the server, so don\\'t reuse a personal password!")}}',
+    title: '{{ _("")}}',
     trigger: 'focus',
     placement: 'right'
 });
--- a/ihatemoney/templates/home.html
+++ b/ihatemoney/templates/home.html
@@ -91,7 +91,7 @@
 {% endblock %}
 {% block js %}
 $('#creation-form #password').tooltip({
-    title: '{{ _("")}}',
+    title: '{{ _("Don't reuse a personal password. Make an access code and send it to your friends")}}',
     trigger: 'focus',
     placement: 'right'
 });


Glandos:
Thanks for this modification! In my opinion "logged in" would be better than "log-in" in this case, what do you think?

I am also wondering if it's worth saying that the code is needed to access the project on a new computer. It's normally obvious, and it will make the tooltip longer for nothing. Any opinion?

rajat641:
> Thanks for this modification! In my opinion "logged in" would be better than "log-in" in this case, what do you think?
> 
> I am also wondering if it's worth saying that the code is needed to access the project on a new computer. It's normally obvious, and it will make the tooltip longer for nothing. Any opinion?

rajat641:
@Glandos - i think you're right we can simply add `Use this code while loggin`

rajat641:
I think there is no need for any message on the tooltip. @Glandos - Just thought about your point above and it makes a lot of sense.

rajat641:
Hi  @almet @comradekingu @Glandos 
Can you please help in knowing a proper tooltip message. Its been pending for a while now. I would appreciate your response.

Glandos:
Hi @rajat641 I'm deeply sorry for this late answer. You can imagine that working on my free time is sometimes not as much as I'd like…

Anyway, thanks for your final string, it meets all needed requirements to me. And thanks @comradekingu for his suggestion.

zorun:
It's been pending for only 5 days, I would not call that a long time...

This breaks the javascript code and doesn't display because it's missing escaping: `Don\\'t` instead of `Don't`.

The current text is quite good and short, but it should use "private code" instead of "access code" for consistency. Also, I would say "choose" instead of "make", something like this:

> Don\\'t reuse a personal password. Choose a private code that you can share with your friends.

@rajat641 can you fix in another pull request?

indatwood:
Hey @zorun! The best way to not forget about this is probably to open a new issue (if it's not done already!)

Cheers :-)

rajat641:
Hi @indatwood @zorun @Glandos ,
I have raised the Issue and linked my PR with that. It was really a silly bug from my end.
I should have re-tested it again after the change but I got a little lazy about it.
Also, I'll be a little more patient in the future about code reviews as now I know that you guys work on this in free time.

PR #625, opened 2020-05-21
Replace currencyformat_nc with currency filter

Glandos:
This new context filter use the project default currency as default
currency. It is way cleaner.

It is in draft, waiting for #619 first.


indatwood:
Yay!

PR #628, opened 2020-05-24
Build translations

Glandos:


zorun:
You added a sqlite database in the same commit ;)

Glandos:
Thanks for spotting this. I think I clicked too much. But now, I have told weblate to compile PO to MO and merge it in the same request. We'll see, but it should help.

PR #631, opened 2020-05-24
Enable CI builds with postgresql and mariadb

zorun:
This is still a WIP to see if it works.

### DIFF
--- a/.github/workflows/test-docs.yml
+++ b/.github/workflows/test-docs.yml
@@ -10,10 +10,40 @@ jobs:
   build:
 
     runs-on: ubuntu-latest
+
+    # Use postgresql and MariaDB versions of Debian buster
+    services:
+      postgres:
+        image: postgres:11
+        ports:
+          - 5432:5432
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: ihatemoney
+          POSTGRES_DB: ihatemoney_ci
+        options:
+          --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+      mariadb:
+        image: mariadb:10.3
+        env:
+          MARIADB_ROOT_PASSWORD: ihatemoney
+          MARIADB_DATABASE: ihatemoney_ci
+        options: >-
+          --health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=3
+        ports:
+          - 3306:3306
+
     strategy:
       fail-fast: false
       matrix:
         python-version: [3.6, 3.7, 3.8, 3.9]
+        database: [sqlite]
+        # Test other databases only with one version of Python (Debian buster has 3.7)
+        include:
+          - python-version: 3.7
+            database: postgresql
+          - python-version: 3.7
+            database: mariadb
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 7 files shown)


indatwood:
Hey @zorun what should we do with this? Was it working? Should we merge it, or close it? Thanks for the heads up!

zorun:
> Hey @zorun what should we do with this? Was it working? Should we merge it, or close it? Thanks for the heads up!

Hey, sorry for leaving this in a WIP state.  I just rebased it (it needed some manual fix-up).

As far as I remember, it wasn't merged because the tests were failing with mysql (see #632). We need to fix that issue before merging, otherwise all PRs would appear to be faulty.

zorun:
This should fail (because I explictly trigger a known bug): https://travis-ci.org/github/spiral-project/ihatemoney/builds/757617787

This is supposed to work but will probably fail because of #632 : https://travis-ci.org/github/spiral-project/ihatemoney/builds/757618257

almet:
Hey, if the tests are failing with this, and it's the expected behavior, then I think we should merge it and fix the bug that this points out. Right?

So, merge time?

zorun:
> Hey, if the tests are failing with this, and it's the expected behavior, then I think we should merge it and fix the bug that this points out. Right?
> 
> So, merge time?

Well, it depends if fixing the MySQL bug takes 1 day or 1 year :)

If we merge this while not fixing the bug, all pull requests will have failing CI. It's not very nice for people proposing patches.

Glandos:
Is it possible to replace `service: mysql` with `mariadb` to test it as noted in https://github.com/spiral-project/ihatemoney/issues/632#issuecomment-817342784 ?

zorun:
I've added support for the github CI, and switched to mariadb instead of mysql.

I'm not 100% satisfied, because the postgresql and mysql containers are started in all jobs (even those using sqlite). But it only means that the CI is a bit longer than necessary, and fixing it would involve writing multiple actions with some kind of inheritance, and I'm too lazy to learn how to do that.

So, it's good to go for a review @Glandos !

.travis.yml
```diff
@@ -5,8 +5,18 @@ python:
...
+  - psql -c 'create database travis_ci_test;' -U postgres
+  - mysql -e 'CREATE DATABASE travis_ci_test;'
 script: tox
 before_install:
   - python -m pip install --upgrade pip virtualenv
 install:
   - pip install tox-travis
+services:
+  - mysql
+  - postgresql
```
Glandos, line 22: Is this a leftover of an old commit?
I don't have any objection against maintaining old Travis config, but it references mysql, so I guess this should be either dropped or updated.

Glandos (changes requested):
See comment on .travis.yml

.travis.yml
zorun: Right, this is a commit I did when we were using Travis.

I have now removed all Travis changes.

zorun:
Uh, weird, I just rebased and there's a new postgresql error:

```
psycopg2.errors.GroupingError: column "bill.id" must appear in the GROUP BY clause or be used in an aggregate function
```

zorun:
Ah, it's the new check from #661 . I should have merged before you @Glandos ^^

```
return self.get_bills_unordered().group_by(Bill.original_currency).count() > 1
```

I'm looking at it.

Issue #634, opened 2020-05-24
ToolTip Bug Fix


rajat641:
This is a bug related to Issue #610

Note: I have updated the code and raised a PR #635 

![image](https://user-images.githubusercontent.com/8374949/82767472-b4ce3400-9ddc-11ea-87bb-b5b4581fe4a5.png)

PR #635, opened 2020-05-24
fixed bug on tool-tip message

rajat641:
This is related to issue #634 .
This is tested locally.


zorun:
Looks good, thanks for fixing! :+1:

PR #651, opened 2020-07-11
added 'hi' in SUPPORTED_LANGUAGES list in default_settings.py

raghupalash:


zorun:
Thanks for the translation!

PR #652, opened 2020-07-16
Changed @media max-width for .home .card to 450px to fix responsiveness

Panda4817:
I noticed on my iphone 6 that the card divs for the login forms were not central on the screen. It seems to be fixed by increasing the @media max width.

before:
before change

after:
after change


zorun:
Indeed, it seems better, thanks!

PR #659, opened 2020-07-21
Fix remove project history with MySQL

### LINKED ISSUE #632: History test case fails with MySQL
The work-in-progress CI on postgresql and mysql (see #602 and #631 ) uncovered its first bug :)

The testcase that fails with MySQL is `HistoryTestCase.test_disable_clear_no_new_records`:

> sqlalchemy.exc.InternalError: (pymysql.err.InternalError) (1093, "You can't specify target table 'bill_version' for update in FROM clause")
> [SQL: DELETE FROM bill_version WHERE bill_version.id IN (SELECT bill_version.id AS bill_version_id 
> FROM bill_version INNER JOIN person ON bill_version.payer_id = person.id 
> WHERE person.project_id = %(project_id_1)s)]
> [parameters: {'project_id_1': 'demo'}]
> (Background on this error at: http://sqlalche.me/e/2j85)

See https://travis-ci.org/github/spiral-project/ihatemoney/jobs/690649479 for more details.

@Andrew-Dickinson could you have a look when you have time?

Natim:
Fixes #632

### DIFF
--- a/.github/workflows/test-docs.yml
+++ b/.github/workflows/test-docs.yml
@@ -10,10 +10,40 @@ jobs:
   build:
 
     runs-on: ubuntu-latest
+
+    # Use postgresql and MariaDB versions of Debian buster
+    services:
+      postgres:
+        image: postgres:11
+        ports:
+          - 5432:5432
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: ihatemoney
+          POSTGRES_DB: ihatemoney_ci
+        options:
+          --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+      mariadb:
+        image: mariadb:10.3
+        env:
+          MARIADB_ROOT_PASSWORD: ihatemoney
+          MARIADB_DATABASE: ihatemoney_ci
+        options: >-
+          --health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=3
+        ports:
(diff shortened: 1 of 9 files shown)


zorun:
Debugging SQL requires a fresh mind... I'll have a look tomorrow.

If we can't fix this properly before 5.0.0, we can add this as "known issue" to the release notes and fix it in a later point release.

zorun:
This approach does not seem to work (see failing tests), I think it is mixing up IDs.

I tried a simpler version below, it should work for MySQL and PostgresSQL, but it fails with SQLite...

```python
BillVersion.query.filter(BillVersion.payer_id == Person.id).filter(Person.project_id == project.id)
```

With sqlite:

```
NotImplementedError: This backend does not support multiple-table criteria within DELETE
```

https://docs.sqlalchemy.org/en/13/changelog/migration_12.html#orm-support-of-multiple-table-deletes

Natim:
@zorun in that case we might want to use 2 distincts queries checking what DB driver is in use.

indatwood:
Hey @Natim do you still want to finish this or should we close the issue ? Thanks!

almet:
Okay, I believe doing what @Natim proposes here is the way to go. Marking as an easy-win.

zorun:
I have updated to the new CI, and this change fails on all databases...

This will no longer be needed when we drop support for MySQL, see #775

zorun:
Closing because of #775

PR #660, opened 2020-07-25
Relax pin point dependencies

### LINKED ISSUE #658: Please don't pin point releases and future releases
Hello,

Here is the perspective of pinning from the point of view of a downstream packager.

Let me start with an anectdote.

Yesterday I noticed that sqlalchemy-continuum 1.3.9 does not build on nixpkgs-unstable. One of the tests is failing. This has been fixed in version 1.3.11. This is a no brainer: the point release has no incompatibility, let's upgrade to 1.3.11. Oh no, ihatemoney has pinned sqlalchemy-continuum to 1.3.9 ! We are doomed!

**Please don't pin point releases**

More generally, assume that two libs A and B pin a dependency C to distinct point releases. Maybe you think, we could package both versions of C, and be done with it. Nope: enter D, depending on both A and B. With python modules, this does not work, as there can only ever be one version of a module in a python program.

**Please don't pin point releases**

You are not guilty of this, but I'll still mention this:
Sometime developpers add bounds of the form dep >= min <= current version. Just in case a future version is breaking. It happened that we (=nixpkgs maintainers) had to request twice in a single week to a specific python project to update the upper bound because a new version had been released with no incompatibility. (yes some projects release often)

**Please don't reject future releases**

The summary is that your version bounds should only contain
* lower bounds
* known incompatibilities. 
To my knowledge, ihatemoney has one such incompatibility: WTForms https://github.com/spiral-project/ihatemoney/issues/567
As a packager, this knowledge is very valuable, but you just drown it in false alarms like "sqlalchemy-continuum 1.3.9 required instead of 1.3.11".
Then, you ask, how did I catch that I had to downgrade to a specific version of WTForms ? Your test suite, as well as nixpkgs'.  And as you can see our test suite is large enough that it catches bugs you had not found: https://github.com/spiral-project/ihatemoney/issues/577

If you are still afraid of future incompatibilities, you can switch to semantic versioning https://packaging.python.org/guides/distributing-packages-using-setuptools/#semantic-versioning-preferred
This would look like:

```diff
--- temp/setup.cfg.orig	2020-07-19 12:28:43.923844181 +0200
+++ temp/setup.cfg	2020-07-19 12:30:51.287900175 +0200
@@ -23,22 +23,22 @@
 include_package_data = True
 zip_safe = False
 install_requires =
-    blinker==1.4
-    debts==0.5
-    email_validator==1.0.5
-    Flask-Babel==1.0.0
-    Flask-Cors==3.0.8
-    Flask-Mail==0.9.1
-    Flask-Migrate==2.5.3
-    Flask-RESTful==0.3.8
-    Flask-Script==2.0.6
-    Flask-SQLAlchemy==2.4.1
-    Flask-WTF==0.14.3
-	WTForms==2.2.1
-    Flask==1.1.2
-    itsdangerous==1.1.0
-    Jinja2==2.11.2
-    SQLAlchemy-Continuum==1.3.9
+    blinker~=1.4
+    debts~=0.5
+    email_validator~=1.0
+    Flask-Babel~=1.0
+    Flask-Cors~=3.0
+    Flask-Mail~=0.9
+    Flask-Migrate~=2.5
+    Flask-RESTful~=0.3
+    Flask-Script~=2.0
+    Flask-SQLAlchemy~=2.4
+    Flask-WTF~=0.14
+	WTForms~=2.2,<2.3 # see issue #567 
+    Flask~=1.1
+    itsdangerous~=1.1
+    Jinja2~=2.11
+    SQLAlchemy-Continuum~=1.3
 
 [options.extras_require]
 dev =
```

And, oh surprise, with this setup.cfg, I don't have to remove version constraints at all.
(Note the nice comment for downstream packagers proving that your upper bound on WTForms is not just form of zealotry. This is worth many internet points.)
But first you have to check that all your dependencies commit to semantic versioning of course. For those of them which don't, please only use a lower bound until you determine for sure that some later version is breaking.

Glandos:
Fixes #658

I switch to semantic versioning for install dependencies, and lower bound for dev dependencies.


zorun:
It looks like the updated flake8 doesn't like a complex `if` condition formatted by black.

I think we should keep pinning dev dependencies to avoid inconsistencies.

Glandos:
I have mixed feelings about this:
* It's complicated to follow updates for dependencies, and dev dependencies increase this burden
* Just using lower bounds *can* lead to inconsistencies between developers
* Having continuous build and integration (like Travis) makes the previous point less annoying
* I like using the last version of linting and formatting tools because of what they are

So in conclusion, I would like to try keeping pinned dependencies (with semver) for installed dependencies, and just lower bounds for development ones.

I fixed the flake8 error quite easily (OK, I needed 2 tries :yum:), but if you have other arguments, I'm opened to discussion :+1:

zorun:
Ok, fine with me, we will just have to deal with CI breakage from time to time.

Can you add the two additional constraints discussed in #658, and fix this annoying lone tab?

Glandos:
I didn't touch `Flask-Migrate` since it already uses a lower bound. `Flask-Migrate` doesn't seems to use semantic versioning, e.g. supports for Python 3.4 was dropped in 2.5.2.

PR #661, opened 2020-07-26
Rework currency switching

### LINKED ISSUE #599: Currency: switching back to "no currency" gives unexpected results
When using a default currency in a project, it is possible to add bills in various currencies and they will be converted to the default currency.

However, switching back to "No currency" in the options gives unexpected results:

- new bills are added without currencies (this is expected)
- existing bills keep their currency (this is expected)
- however, the converted amounts still reflect the previous default currency!

This is really weird: for existing bills that had a currency, "How much?" and "Amount" may show different values, without any way to know the currency in which "Amount" is displayed. In addition, if changing again to a new default currency, bills with no currency will be converted to the new default currency while they were probably intended as the old default currency. In short: it's a mess.

I see no easy way to fix this in the general case.  I think the simplest way is to disallow switching back to "No currency".

An alternative solution (but more complicated) is to allow switching back to "No currency" only if all bills are in the same currency, and remove the currency from all bills.

Glandos:
Fixes #599


zorun:
It looks good at a first glance, but there are many cases to cover.

Maybe add a few testcases?

Glandos:
OK, here I need reviews, especially on the `switch_currency` function I moved from `web.py` to `models.py`:
* Is it OK to include such code in the model?
* Should the caller call `db.session.commit` or let the function do it?

zorun:
Things are always somewhat confusing with currencies :/

Looking at the code and the tests, I see a few things and invariants that could be clearer:

- what is the relation between `bill.amount` and `bill.converted_amount`?
- in the test, you add bills without specifying the original currency. I would assume that in that case, bills are created with no currency. However, you later say "switch back to USD". Is USD the default currency then? I would always specify the currency to make things clear.
- invariant: from what I can tell, `original_currency` can only change from a currency to "no currency", or from "no currency" to a currency. That is, it is not possible to directly change from one currency to another. Maybe a test could be built around this?

I'm not even sure if these remarks make sense.

indatwood:
Hey, what should we do with this? Is it ready for review?

Glandos:
Nope, I'm still "working" on it. This is the last part I want to polish before releasing 5.0. I'm trying to find free time for this, but I don't forget it!

almet:
Hey, so it seems this is what blocks the 5.0 release. @Glandos do you know how we could help on this?

ihatemoney/models.py
```diff
@@ -264,6 +265,34 @@ def get_pretty_bills(self, export_format="json"):
             )
         return pretty_bills
 
+    def switch_currency(self, new_currency):
+        # Update converted currency
+        if new_currency != self.default_currency:
+            for bill in self.get_bills():
+                if bill.original_currency == new_currency:
```
almet, line 272: I'm not sure this is what we want here. 🤔 

Consider the case where we've got a project with default currency set to USD. We've got three bills : 

Amount | original_currency | converted_amount
--------- | ------------------- | ---------------------
10 | EUR | 11.88
10 | GBP | 13.79
10 | USD | 10

And now we want to convert to EUR. And the expected result is : 

Amount | original_currency | converted_amount
--------- | ------------------- | ---------------------
10 | EUR | 10
10 | GBP | 11.6
10 | USD | 8.4

Here on the first line, even if `bill.original_currency = new_currency` we need to set `bill.converted_amount = bill.amount`

ihatemoney/models.py
```diff
@@ -264,6 +265,34 @@ def get_pretty_bills(self, export_format="json"):
...
+                    )
+                    # Strip currency
+                    bill.amount = bill.converted_amount
+                    bill.original_currency = CurrencyConverter.no_currency
+                else:
+                    # Switch to new currency for everyone
+                    bill.converted_amount = CurrencyConverter().exchange_currency(
+                        bill.amount, bill.original_currency, new_currency
+                    )
+                    # Add the currency for previously un-currency-ied
```
almet, line 288: This comment makes things harder to understand to me, heh.

almet:
Hey, I hope you're all good these days. I've got some time today so here I am!  

I've had a look at the code to have a fresh take on all this. And some stuff isn't really clear to me. It's not really due to the code you wrote @Glandos but probably to the initial implementation of this currency thing.

The model has different fields which makes it complicated to work things out as there is no definition of the fields intention. We might want to add some comments in the model to make things easier to understand.

```python

class Bill(db.Model):
    amount = db.Column(db.Float)
    original_currency = db.Column(db.String(3))
    converted_amount = db.Column(db.Float)
```

Here we have three different fields. I'm just trying to understand what they do to be sure to get things right :

- `amount` is the original amount that's being entered on the form.
- `original_currency` is the currency the amount entered was on.
- `converted_amount` is the conversion of the amount to the currency of the project.

I'm just putting this here for future reference (it helped me understand what I was reading)

ihatemoney/tests/tests.py
```diff
@@ -1567,6 +1567,85 @@ def test_access_other_projects(self):
...
+
+        new_currency = "CAD"
+        project.switch_currency(new_currency)
+        bills = project.get_bills()
+        for bill in bills:
+            if bill.original_currency == new_currency:
+                self.assertEqual(bill.amount, bill.converted_amount)
+            else:
+                self.assertNotEqual(bill.amount, bill.converted_amount)
+
```
almet, line 1648: I believe here we could add the test case I was talking about.

almet:
Hey, I've commented on the pull request. Let me know if I can help any further on this.

almet:
I've added a test for this, removed the problematic line and merged with master.

Let me know if you think it's good to be merged :-) Thanks!

almet:
Hey @Natim, @Glandos or @zorun, if you have some spare time to review this, that would be awesome! Thanks :-)

almet:
Hey, anybody wants to review this so we can land it? Thanks!

zorun:
It's a difficult one, I definitely can't review this at this time of night :-)

I'll try to have a look in the coming days.

zorun:
I had a look and the test helps to understand what should happen.

However, the code (and the test) doesn't address one point of #599: once there are bills in various currencies, it should not be possible to switch back to the "no currency" mode, because there is no good way to do it.

I have other minor reviews, I'll make them directly as comments in the code.

ihatemoney/tests/budget_test.py
```diff
@@ -1446,6 +1448,112 @@ def test_access_other_projects(self):
...
+                "payer": 3,
+                "payed_for": [2],
+                "amount": "13.33",
+            },
+        )
+
+        project = models.Project.query.get("raclette")
+
+        # First, no currency
+        project.switch_currency(CurrencyConverter.no_currency)
```
zorun, line 1502: It's not clear what we are testing here: since the project was created without specifying the currency, I would expect it to already be in "no currency" mode.

It turns out that `post_project` used in the tests defaults to USD! It should rather default to "no currency" to make things clear, and also because "no currency" is the default choice in the UI when creating a project.

zorun:
I think a complete test would look like this:

- start a project with no default currency
- add bills
- check that the bills have no currency and that amount == converted_amount
- switch project currency to EUR
- check that the existing bills are in EUR, and that amount == converted_amount
- add a new bill in EUR, and check that amount == converted_amount
- switch project currency back to "no currency"
- check again that all bills have no currency and that amount == converted_amount
- switch project currency to EUR again
- add a new bill in CAD, check that the converted amount is correct for CAD -> EUR conversion
- switch project currency to USD
- check that all bills stayed in their original currency, and that the converted amount is correct for {EUR,CAD} -> USD conversion
- try to switch project currency to "no currency": it should fail

Glandos:
OK, I rebased, moved to GitHub Actions (Travis.org is dead since June, 15th), and started to seriously work on this.

First, I already changed the default project creation to "No Currency". As @zorun said, it is the default in the UI, so it must be the default in tests. It required some modifications, but nothing too difficult.

Then, I am rewriting the test as @zorun scripted it. But:

> - try to switch project currency to "no currency": it should fail 

I don't really see why it should fail in the last step. In the current algorithm, every bills is first converted to the current currency before stripping the currency (aka set to "No currency"). For example, if we have a project set to `USD` and one bill of `CAD 10`, then the `converted_amount` and the `amount` is set to `8.33` (with a conversion of 1.2), and finally, `original_currency` is set to "XXX" (No Currency).
Is that an unexpected behaviour?

zorun:
Thanks for the update!

> For example, if we have a project set to USD and one bill of CAD 10, then the converted_amount and the amount is set to 8.33 (with a conversion of 1.2), and finally, original_currency is set to "XXX" (No Currency).

One issue is that we lose information, since the exchange rate can vary over time.

In my opinion, `amount` is entered by the user and should never be changed by the system. The amount can be related to a bill or invoice, so changing it makes it basically impossible to cross-reference.

With this constraint, as soon as there are bills in different currencies, I think it's impossible to switch back to "no currency" (it does not make sense). If we relax this constraint and recompute `amount`, then there should be a big warning in the UI before doing that.

Glandos:
Oh, that's a frondend issue then. Let's forward the issue to the frontend team! Joke done.

I think you're right, we can strictly forbid to switch to "No Currency" as soon as we have more than one currency in our bill list. It's simpler, and probably this use case will never happen. We can display a nice error message with the details then.

Glandos:
I need help for one point.

I am creating `validate_default_currency` into `EditProjectForm` to have a clean validation here. However, to validate the `default_currency` field, I need to have access to the project. But in this form, I don't have enough information to query the database, since the project id is missing :/

Do you know if it's possible to attach some context before form validation?

Glandos:
I finally found a possible way: pass the project id in the `EditProjectForm` `__init__` function.
This is a workaround, so I request your kind review on it :smiley:

zorun:
Thanks!

I think there's no need for a hack: the name is passed in the form, so you should be able to use `get_by_name`

I'm still a bit confused that we have both ID and name as strings...

ihatemoney/models.py
```diff
@@ -263,6 +271,37 @@ def get_pretty_bills(self, export_format="json"):
...
+                    bill.amount = bill.converted_amount
+                    bill.original_currency = CurrencyConverter.no_currency
+                else:
+                    # Switch to new currency for everyone
+                    bill.converted_amount = CurrencyConverter().exchange_currency(
+                        bill.amount, bill.original_currency, new_currency
+                    )
+                    # Add the currency for bills that were created without currency
+                    if bill.original_currency == CurrencyConverter.no_currency:
+                        bill.original_currency = new_currency
```
zorun, line 299: Is it even possible to have a bill with no currency here?

It looks like the two changes are in the wrong order: to compute `bill.converted_amount`, we need `bill.original_currency`, but it is only set in the second change.

ihatemoney/forms.py
```diff
@@ -142,6 +147,23 @@ def save(self):
...
+            raise ValidationError(
+                _(
+                    (
+                        "This project already uses different currencies"
+                        "and can't be set to 'No Currency'"
+                    )
+                )
+            )
+        pass
+
```
zorun, line 166: Where is this used? I don't find any call to this function.

Small suggestion for the error message: "This project cannot be set to 'no currency' because it contains bills in multiple currencies."

ihatemoney/forms.py
```diff
@@ -110,6 +111,10 @@ class EditProjectForm(FlaskForm):
     default_currency = SelectField(_("Default Currency"), validators=[DataRequired()])
 
     def __init__(self, *args, **kwargs):
+        if not hasattr(self, "id"):
+            # We must access the project to validate the default currency.
+            # Mimics a StringField. Defaut to empty string to ensure that query run smoothly.
+            self.id = SimpleNamespace(data=kwargs.pop("id", ""))
```
zorun, line 117: Instead of adding an ID, you should be able to query project by name:

```
models.Project.query.get_by_name(name=XXX)
```

ihatemoney/tests/budget_test.py
```diff
@@ -1443,6 +1447,171 @@ def test_access_other_projects(self):
...
+        self.assertEqual(models.Project.query.get("raclette").default_currency, "USD")
+
+    def test_currency_switch_to_bill_currency(self):
+
+        mock_data = {"USD": 1, "EUR": 0.8, "CAD": 1.2, CurrencyConverter.no_currency: 1}
+        converter = CurrencyConverter()
+        converter.get_rates = MagicMock(return_value=mock_data)
+
+        # A project should be editable
+        self.post_project("raclette", default_currency="USD")
```
zorun, line 1586: The comment should be updated

ihatemoney/models.py
```diff
@@ -215,15 +216,25 @@ def exactmatch(self, credit, debts):
...
     def has_bills(self):
         """return if the project do have bills or not"""
-        return self.get_bills().count() > 0
+        return self.get_bills_unordered().count() > 0
 
-    def get_bills(self):
-        """Return the list of bills related to this project"""
+    def has_multiple_currencies(self):
+        """Return if multiple currencies are used"""
+        return self.get_bills_unordered().group_by(Bill.original_currency).count() > 1
```
zorun, line 223: What happens if there are some bills with a currency like EUR, and some bills with no currency? I think it would still count as 2 with this function. But I have no idea what we should do in that case anyway...

zorun (changes requested):

Glandos:
[**@zorun**](https://github.com/zorun) commented on [3 juil. 2021 à 12:32 UTC+2](https://github.com/spiral-project/ihatemoney/pull/661#issuecomment-873384729 "2021-07-03T10:32:49Z - Replied by Github Reply Comments"):
> Thanks!
> 
> I think there's no need for a hack: the name is passed in the form, so you should be able to use `get_by_name`
> 
> I'm still a bit confused that we have both ID and name as strings...

```
❯ sqlite3 /tmp/ihatemoney.db 'SELECT * FROM project WHERE name = \'demonstration\';'
demo|demonstration|pbkdf2:sha256:260000$QkI8WUSFMlJdWot4$58d9f38efced65395067ea200a02ee180b54fc59418c3e9d28785b3af9c96387|demo@notmyidea.org|ENABLED|EUR
hop|demonstration|pbkdf2:sha256:260000$k9xOopW1grARJ0mM$e5e5e41d822e76a76d6d2f820172ffe37b7256cfe3670ee0e9dea49b99ba5963|hop@a.fr|ENABLED|XXX
```

It is possible to change the name after creation. Name is not unique in database, for obvious reasons. On the main ihatemoney.org, a lot of people can have a display name of "Monthly budget", even if the ID is of course unique.

ihatemoney/forms.py
Glandos: This is used by WTForms. A function called `validate_<FIELD_NAME>` will be called to validate `FIELD_NAME`.

Thanks for the suggestion, it will put emphasis on the error itself.

ihatemoney/forms.py
Glandos: Thanks, but name are not unique in database, so we need to discriminate them.

ihatemoney/models.py
Glandos: It is currently impossible to have this. When switching EUR -> XXX (no currency), all bills are converted to XXX.
In XXX default, it's only possible to add bills with XXX.
In any other default (e.g. EUR), it's only possible to add bills with a currency (e.g. **not** XXX).

I don't know if our test cases cover this from the API though.

ihatemoney/models.py
Glandos: Yes, it is possible, when switching from XXX (no currency) -> EUR. `original_currency` is XXX, and we must set it to the new default currency.
If switching USD -> EUR, this is not possible.

The two affectations does the following:
- Compute the *new* `converted_amount` by using the conversion from bill's `original_currency` to `new_currency`.
- Override the `original_currency` if and only if it was set to XXX.

So the first affectation will be a "simple" recomputation, since conversion ratio of XXX is set to 1.

Glandos:
By the way, I thought of disabling the option to set "No currency" in the dropdown menu. But WTForms seems not to handle this case natively : https://phdesign.com.au/programming/wtforms-selectfield-custom-option-attributes/

ihatemoney/forms.py
zorun: Ok, that makes sense.

ihatemoney/forms.py
zorun: Ah, right!

ihatemoney/forms.py
zorun: So, I guess we do need this hack.

I had a hard time understanding the code (which led me to open #774 ), so please add more comments ^^ For instance on why you need the `not hasattr(self, "id")` and what is this `SimpleNamespace` thing.

ihatemoney/models.py
```diff
@@ -263,6 +271,37 @@ def get_pretty_bills(self, export_format="json"):
...
+            for bill in self.get_bills_unordered():
+
+                if new_currency == CurrencyConverter.no_currency:
+                    # Use old currency to flatten all amount before stripping
+                    bill.converted_amount = CurrencyConverter().exchange_currency(
+                        bill.amount, bill.original_currency, self.default_currency
+                    )
+                    # Strip currency
+                    bill.amount = bill.converted_amount
+                    bill.original_currency = CurrencyConverter.no_currency
```
zorun, line 291: I don't get why we need to convert the amount here.  When `new_currency == XXX`, we just need to remove the currency indication on all bills. The amount doesn't need to be converted (because we checked that all bills have the same currency). Also, in principle, we should never touch `bill.amount`.

There is this edge-case that is not completely clear:

- all bills have original_currency=EUR
- current default currency is USD
- new default currency is XXX

In that case, I would still not convert the bills to USD before removing the currency: it's not useful and modifies the original amount for no good reason.

So in the end this case can be simplified to simply:
```
# We are removing the currency, and we already checked that all bills had the same currency:
# it means that we can simply strip the currency without converting the amounts.
# We basically ignore the current default_currency.
if new_currency == CurrencyConverter.no_currency:
    # Reset converted amount in case it was different from the original amount
    bill.converted_amount = bill.amount
    # Strip currency
    bill.original_currency = CurrencyConverter.no_currency
```

ihatemoney/models.py
zorun: Ah, right, it makes sense with the additional knowledge that the conversion rate of XXX is 1 :)

It would clarify the code to make that explicit:

- if original_currency == XXX, just set `converted_amount = amount` and `original_currency = new_currency`
- otherwise, apply the conversion

ihatemoney/tests/budget_test.py
Glandos: Done, thanks for the finding

ihatemoney/models.py
Glandos: OK done

ihatemoney/forms.py
Glandos: OK, done. I didn't explain the SimpleNamespace behavior, since it's in the default library.

zorun: Thanks!

ihatemoney/models.py
```diff
@@ -263,6 +271,37 @@ def get_pretty_bills(self, export_format="json"):
...
 
+    def switch_currency(self, new_currency):
+        # Update converted currency
+        if new_currency != self.default_currency:
+            if (
+                new_currency == CurrencyConverter.no_currency
+                and self.has_multiple_currencies()
+            ):
+                raise ValueError(f"Can't unset currency of project {self.id}")
+            for bill in self.get_bills_unordered():
```
zorun, line 282: I suggest moving the `new_currency == CurrencyConverter.no_currency` test before the `for bill` loop, because the test does not depend on the bill. It's not so much about performance, but about readability.

Glandos: Do you mean either:
- Storing the result of `new_currency == CurrencyConverter.no_currency` into a variable and use it in the loop;
- or testing it and write two different loops in if/else blocks?

ihatemoney/models.py
Glandos: Hmm, I am puzzled by the actual code, but I guess that's because it was done when we still allowed the switch to 'XXX' with multiple used currencies.
In the current scenario, I agree that we could simply strip the currency, full stop.

I will write a test first for this.

Glandos: Now there is a test. This fails (obviously), so please tell me if it fits the use case.

ihatemoney/models.py
zorun: I was thinking about two separate loops.

ihatemoney/models.py
zorun: Cool, the test looks good!

almet (approved):
Hi @zorun and @Glandos. Thanks for taking this over.

It seems good to me. It simplifies the code to not consider the multiple currency to no currency case, so good call 👍 

I don't have much to say on the topic, it all looks good and is understandable to me. I haven't tested the UI though.

ihatemoney/models.py
zorun: Thanks!

zorun (approved):

Issue #656, opened 2020-07-17
Show "External link" less prominently when creating bill


zorun:
Since #479, it is possible to add an URL as "External link" when creating a bill.

However, the field is really big and prominent in the bill creation page, even for people that do not need this feature.

It would be good to find a less prominent design, so that:

- people that don't need External links are not bothered by the additional interface
- people that need External links can still easily add them

A more general discussion about optional features is at #603 (see the example about the weights, I really like it)

almet:
So, what about an "extra fields" clickable link which expends to a few optional fields that could be then filled?

Glandos:
Like a simple [`details/summary`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) tag?

zorun:
It's a simple and pragmatic way to fix this, I like it.

People that want to add an external link on each bill might find it annoying, because it would require one more click for each bill. But we can start with this simple solution, and somebody could propose a better UI/UX design later if this use-case is important for them.

zorun:
Anybody want to work on this? (I suck at web UI)

Glandos:
I hacked something like this:

https://user-images.githubusercontent.com/297578/124379620-703bea00-dcb8-11eb-8fd3-d04bcace02d1.mp4

Does it sound good (except some margins…)

zorun:
Excellent! :+1: 

Yeah, you probably need a slightly bigger margin between "Options avancées" and the actual fields.

PR #666, opened 2020-07-26
Ajouts au changelog 5.0

JocelynDelalande:


JocelynDelalande:
Honêtement, j'ai pu en manquer certains dans le git log , mais j'en aurais au moins attrapé une partie.

Pas clair pour moi : les numéros que l'on met dans le changelog, cesont des n° de ticket ou de PR ?

Glandos:
Quand il y a une PR, le numéro de PR, sinon le ticket, avec l'identifiant version courte du commit.

JocelynDelalande:
OK, merci pour l'éclaircissement :) peut-être qu'on pourrait l'écrire quelque-part.

J'aurais instinctivement pensé l'inverse : le changelog est utile aux admins et aux curieux·se·s, la PR ne l'est qu'aux dévs.

zorun:
Thanks for spotting the missing entries!

I believe these two are already in the changelog for 4.1.4 so they can be omitted for 5.0:

- Fix bug on tool-tip message
- Change the color of the add a member button

I admit it can be hard to be sure because of the backports.

JocelynDelalande:
> Thanks for spotting the missing entries!

Thanks for spotting my spotting issues ! 🤪

PR is fixed. Archeologists will thank us ;-)

indatwood:
Hi, thanks for this! Where are we with this? Should we merge it and then release 5.0?

Glandos:
I'll wait for a resolution of #597 (via #688 if everything goes well) and merging of #661.

zorun:
I had completely forgotten that 5.0 is still not released :/

For the changelog, we have two new translations: Portuguese and Japanese.

We have a milestone with a list of bug / PR that should be part of 5.0: https://github.com/spiral-project/ihatemoney/milestone/5

Of course it's open to discussion. Those that seem important to me:

- #659 is a bug fix, but I think it's stuck. If we don't find an easy way to fix it, we shouldn't delay the release because of it (just add as known issue in changelog)
- #656 is a UX regression and should be fixed (in my opinion)
- #587 would improve translations, but it needs time for all translators to adapt. We can either merge it and wait a few weeks before releasing, or keeping it for another release. I think I prefer keeping it for another release.

zorun:
And of course, there's #599 / #661 which is really tricky...

PR #673, opened 2020-08-23
Update apache-vhost config generation

kobus-v-schoor:
The current configuration generated uses the old syntax for Apache access control. On a Debian 10 install I recently did the generated config just resulted in an access denied error when using Apache and mod_wsgi. Making the changes in this pull request to my Apache config solved the issue (got the hint looking at Flask's documentation here: https://flask.palletsprojects.com/en/1.1.x/deploying/mod_wsgi/#configuring-apache).


Glandos:
Thanks a lot!
I guess that Apache is not used by users of this project :)

kobus-v-schoor:
Guess not :) Thanks for the fantastic project!

PR #680, opened 2020-09-25
Partial resolution to #597 #issuecomment-623154318

dcorriveau:
-Added new icons for f-droid and play store
-added them to layout and changed the href
-May require some more testing!


Glandos:
Very nice, thanks a lot!

However, I see some issues:
- Why there are some empty groups in these SVG files? I don't see any valid reason, but I may miss something
- Shouldn't we use the [F-Droid SVG icon](https://f-droid.org/assets/fdroid-logo.svg) for F-Droid?
- The `footer-right` div is now too narrow to contain all icons, so they wrap. The `max-width` rule should be relaxed.

Do you think you can introduce changes yourself for that?

dcorriveau:
Just added them! Let me know if it was what you were looking for! :)

indatwood:
Hi, so this seems to be conflicting with #688. What should we go for?

Glandos:
Yes, I think that a new page for mobile is better than just different icons, because now, we already have 3 different links for mobile. And since the API is opened, this list could grow in the future, so the page can be useful to list everything.

I don't want to close this before anything is merged though.

Glandos:
I am sorry to close this because another solution was accepted in #688.

However, you are free to review it and improve it!

PR #684, opened 2020-09-28
Update setup.py

AzharMithani:


AzharMithani:
Hope it is helpful :)

Glandos:
Hi @AzharMithani thanks for spending some of your time on this project!

Is there any reason for including a comment with your source code disclaimer? It seems at least out of this project's scope.

AzharMithani:
I saw an issue related to the source code disclaimer so thought that the
link would be helpful but wasn't sure if the issue had been resolved so
included the link in the comment section.

Glandos:
Sorry I'll close this because I find it irrelevant. The first two added comments don't add value, since the code is pretty obvious, and the last one doesn't apply to this project.

PR #685, opened 2020-09-30
Simplify tests

Villaquiranm:
Hello I've started working in simplify the unit tests of the project and i wanted to know what is your opinion about what I have done until now (Not a lot).

Do you have any better ideas ? 
I wanted to separate the tests by usecases (depending on the size in code lines of each usecase)

- APITestCase (749 lines)
- BudgetTestCase (1444 lines)
- HistoryTestCase (600 lines)

Also a general test file with usecases that does not have more than 150 lines of code.

I've modified the `make test` command in order to execute each of the files ( this need improvement i wanted to execute all the files in the directory, like that if we add a new file, we won't need to add it manually in the script, but i don't know a lot  about tox config. If you have any ideas let me know, if not I'll investigate about it later.

I've also try to replace "Tartiflette" with a variable and try to modify almost everywhere in the **BudgetTestCase** file but i'm not sure the result is great :)


Villaquiranm:
#501

tox.ini
```diff
@@ -6,7 +6,8 @@ skip_missing_interpreters = True
 
 commands =
     python --version
-    py.test --pyargs ihatemoney.tests.tests
+    py.test --pyargs ihatemoney.tests.general_test ihatemoney.tests.api_test \
+    ihatemoney.tests.budget_test ihatemoney.tests.history_test
```
Glandos, line 10: You should be able to remove all named tests, and just use `py.test --pyargs ihatemoney.tests`. See https://docs.pytest.org/en/latest/goodpractices.html#test-discovery.

Villaquiranm: You're right, i though i've tested that thanks :+1:

ihatemoney/tests/api_test.py
```diff
@@ -0,0 +1,773 @@
...
+
+
+class APITestCase(IhatemoneyTestCase):
+
+    """Tests the API"""
+
+    def api_create(self, name, id=None, password=None, contact=None):
+        id = id or name
+        password = password or name
+        contact = contact or f"{name}@notmyidea.org"
```
indatwood, line 19: In general, I tend to prefer having the tests easy to read, so relying on a variable name here seems weird to me.

indatwood: Also, if we chose to go with the variable name, I believe it should be something we can change by subclassing the tests. So maybe put it in a constant inside the class?

ihatemoney/tests/general_test.py
```diff
@@ -0,0 +1,271 @@
+import io
```
indatwood, line 1: I'm not sure about the "general" name here, but that might just be me. What about `main_test`?

indatwood (changes requested):
Hey, thanks for taking the time to tackle this. It will be handy I think :-) I've added a few remarks that should probably be adressed before merging, if you can. Thanks :-)

ihatemoney/tests/general_test.py
Villaquiranm: Sure, i'll do it :)

ihatemoney/tests/api_test.py
Villaquiranm: Hello, thanks for taking the time of reviewing this PR :D

I'm not really sure of understanding the changes you're asking me. Could you clarify with a little example ?

This part of the tests was already like that when I separated in several files. I've only added the
`project_name = "raclette"` and replace it where it seems adequate to do so.

But If you explain it to me, I'll be glad to help

indatwood: So, my idea was to revert the `project_name = "raclette"`` part, because I believe it's just adding a layer of redirection and makes the tests harder to read in my opinion. What do you think about this?

Villaquiranm: Yeah, i was not sure about the result, I agree it will be better without it, I will fix this when I have some free time next days. thanks 👍

Villaquiranm: @indatwood as you requested I've removed the variable *project_name*, so now the tests are unchanged only separated in many files :)

Glandos:
This is indeed a "simple" step. But necessary.
@indatwood do you see more things to integrate as a start?

Glandos:
Thanks for that work! I'm really bad at writing tests, and so amazed at using them when I write some new code :nerd_face:

PR #687, opened 2020-10-01
Update messages.po

gteti:
Wrong Italian translation -> corrections.

### DIFF
--- a/ihatemoney/translations/it/LC_MESSAGES/messages.po
+++ b/ihatemoney/translations/it/LC_MESSAGES/messages.po
@@ -41,7 +41,7 @@ msgid "Default Currency"
 msgstr "Valuta predefinita"
 
 msgid "Import previously exported JSON file"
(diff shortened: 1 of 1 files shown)


Glandos:
Hi, and thanks for taking care of improving this project!

It would have been better to use the [translation platform](https://hosted.weblate.org/projects/i-hate-money/i-hate-money/it/#overview) to suggest improvements. You can even log in with GitHub credentials if you don't want to create a new account.

Moreover, I see in the last change that you remove a format option (`%(num)` instead of `%(num)d`) which seems incorrect.

zorun:
Indeed, the format string needs to be fixed in the proposed change.

zorun (changes requested):
The format string `%(num)d` must not be changed.

Glandos:
I integrated manually changes into Weblate, pushed as #708. All strings are marked for review, since I'm not fluent in Italian.

PR #691, opened 2020-10-25
Update messages.po

mppmpp315:
Some of these Russian translations make sense literally, like they are taken from Google Translate, but not in a practical sense. I've edited some of them here. Thank you for your consideration!


Glandos:
Hi, and thanks for your contribution!
I have no skills in Russian, but is it possible for you to edit this in https://hosted.weblate.org/projects/i-hate-money/i-hate-money/ru/ ? You can log in easily using your GitHub credentials, if you like.

The advantage is that Weblate performs additional checks and compilation before merging.

PR #703, opened 2020-12-07
Changed the navbar element spacing to make it look more even.

mckerracher:
Hello! I thought that maybe the navbar element would look better evenly spaced, so I changed the list element class to "navbar-nav nav-fill w-100".


Glandos:
I agree that the navbar is about to explode, but with this modification, I found that the top right menu is a bit weirder:
image

To remove some entries, we could move "Settings" and "History" inside the drop-down menu at the top right.

Thanks for opening the discussion on this :+1:

almet:
I'm +1 on moving these two items on the menu.

almet:
I've added a few changes in a subsequent PR #739

PR #766, opened 2021-06-06
Fix test for generate_hash_password

Glandos:
Fix test by using the function that check password instead of just checking the output length. This length can be variable and will change with Werkzeug 2

### DIFF
--- a/ihatemoney/manage.py
+++ b/ihatemoney/manage.py
@@ -34,10 +34,14 @@ def runserver(ctx):
 
 
 @click.command(name="generate_password_hash")
-def password_hash():
+@click.option(
+    "--method",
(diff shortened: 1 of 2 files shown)


almet:
Hey, it doesn't seem to be enough to fix it. Here's the answer from Travis CI : 

```
    def test_generate_password_hash(self):
        runner = self.app.test_cli_runner()
        with patch("getpass.getpass", new=lambda prompt: "secret"):
            result = runner.invoke(password_hash, ["--method", "pbkdf2:sha256:260000"])
>           self.assertEqual(len(result.output.strip()), 102)
E           AssertionError: 94 != 102
```

Glandos:
Yep, it seems that werkzeug changed the salt length from 8 to 16 in version 2, and in my current venv, I am mixing them. I finally change the test to use the method that really check the password.

almet:
👍

PR #675, opened 2020-08-27
Changed the favicon image to something more clear

Sammed98:
This is regarding the issue #543 . I have change the icon to something more clear and on the lines of "I hate money".

Issue #543, opened 2020-03-13
Icon is too small


indatwood:
*Hi, and thanks for this project, it's great :-) I just discovered it and I'm opening a few issues here and there to propose feedback. Hope it's the right way to do things around here!*

The [icon used as a favicon](https://raw.githubusercontent.com/spiral-project/ihatemoney/master/ihatemoney/static/favicon.ico) is too small for my eyes, or maybe it depends on the colors used, but I find it almost useless, and even counter-productive!

There are plenty of existing icons provided with CC license by the Nount project. What about this one, for instance?

https://thenounproject.com/search/?q=money&i=1406358

almet:
Hi, thanks for opening the issue, that's appreciated, and thanks for the kind words :-) 

I'm okay with updating the current icon, this one is not really handy! Not sure about the one you're proposing though, it looks good, but I'm not sure it'll show properly as a favicon.

najann:
Hi there!
I just came across this project and already added some translations using weblate :)
What do you think of the following icons as an alternative for the current favicon?
<table>
<td><img width="90%" alt="Screenshot 2020-05-26 at 12 23 23" src="https://user-images.githubusercontent.com/47032941/82890161-0943f180-9f4c-11ea-9b48-2300783a070a.png"></td>
<td><img width="70%" alt="Screenshot 2020-05-26 at 12 25 15" src="https://user-images.githubusercontent.com/47032941/82890168-0ba64b80-9f4c-11ea-9f60-b8184beafa6c.png"></td>
<td><img width="90%" alt="Screenshot 2020-05-26 at 12 25 15" src="https://user-images.githubusercontent.com/47032941/82889662-5ecbce80-9f4b-11ea-9fa3-4ddab92df343.png"></td>
</table>

Glandos:
I like the second one. There are also nice ones in https://thenounproject.com/search/?q=money%20share
Like https://thenounproject.com/term/average/2028785

allanqin:
Hi, I'm looking for my 1st open source project to contribute to. Is anyone currently working on this?

najann:
@allanqin I had the same intention but feel free to work on this issue now!

Sammed98:
I have made a pull request for this issue #675 . Please have a look.

Glandos:
I feel a little uncomfortable with a single € or $, for being less inclusive. However, money lacks a universal symbol (¤ is know by nearly no one), and the dollar sign fit this role since the end of the WWII now…
[...]

PR #771, opened 2021-07-01
Added New Icon

moepotatoe:
If the project is called I Hate Money, don't have money as your icon.  Use something closer to the gold standard.


Glandos:
Then this icon should replace the existing one, not being thrown out randomly into the project root.

This comment is a spam test.

moepotatoe:
No, this isn't spam.  Where should I submit the icon then?

Glandos:
The discussion for this happened in #543 
You could have a look at #675 for an example of paths. However, this was not merge due to licensing. So the question now is: who designed the icon you want to submit?

PR #774, opened 2021-07-04
Improve project edit form

zorun:
Currently, the private code is mandatory in the edit form, and it's not clear whether it's meant to **check** for the right private code or to **change** the private code.

Make the field optional and rename the help message to make it clearer it's only meant to **change** the private code.

Also change the field type to "password" instead of simple string, and refactor some code to clarify.


zorun:
@Glandos while reviewing #661, I took the opportunity to clarify the form code, and I also remembered this minor UI issue. Please review when you have time :)

ihatemoney/forms.py
```diff
@@ -131,9 +132,11 @@ def update(self, project):
         """Update the project with the information from the form"""
         project.name = self.name.data
 
-        # Only update password if changed to prevent spurious log entries
-        if not check_password_hash(project.password, self.password.data):
-            project.password = generate_password_hash(self.password.data)
+        if self.password.data:
+            # Only update password if different from the previous one,
+            # to prevent spurious log entries
+            if not check_password_hash(project.password, self.password.data):
```
Glandos, line 138: Those two conditions should be merged into one I guess

Glandos (changes requested):
Only one "minor" issue ;)

ihatemoney/forms.py
```diff
@@ -102,7 +102,8 @@ def process_formdata(self, valuelist):
 
 class EditProjectForm(FlaskForm):
     name = StringField(_("Project name"), validators=[DataRequired()])
-    password = StringField(_("Private code"), validators=[DataRequired()])
+    # If empty -> don't change the password
+    password = PasswordField(_("New private code"))
```
Glandos, line 106: I wouldn't change the name of of the field, but rather add a description:
```suggestion
    # If empty -> don't change the password
    password = PasswordField(_("Private code"), description=_("Enter a code only if you want to change it"))
```

ihatemoney/forms.py
Glandos:
```suggestion
        # Only update password if different from the previous one,
        # to prevent spurious log entries
        if self.password.data and not check_password_hash(project.password, self.password.data):
```

ihatemoney/forms.py
zorun: I really find it clearer to have "New private code": it completely resolves the ambiguity.

ihatemoney/forms.py
zorun: Right, it was kind of intended because the comment only applies to the second test, but I will improve it then :)

zorun: Let me know what you think of the new version.

ihatemoney/forms.py
Glandos: OK for "New private code", but I also really think that a short description could help.

ihatemoney/forms.py
Glandos: Nice integration of comments!

PR #783, opened 2021-07-07
Enhance bill form

### LINKED ISSUE #656: Show "External link" less prominently when creating bill

Glandos:
Fix #656

Rework UI for new bill form. Additional options are hidden by default (currency and external link)
I also tried to make multiple columns of payee, but this is still WIP

### DIFF
--- a/ihatemoney/api/common.py
+++ b/ihatemoney/api/common.py
@@ -69,7 +69,7 @@ class ProjectHandler(Resource):
         return "DELETED"
 
     def put(self, project):
-        form = EditProjectForm(meta={"csrf": False})
+        form = EditProjectForm(id=project.id, meta={"csrf": False})
         if form.validate() and current_app.config.get("ALLOW_PUBLIC_PROJECT_CREATION"):
             form.update(project)
             db.session.commit()
--- a/ihatemoney/forms.py
+++ b/ihatemoney/forms.py
@@ -1,5 +1,6 @@
 from datetime import datetime
 from re import match
+from types import SimpleNamespace
 
 import email_validator
 from flask import request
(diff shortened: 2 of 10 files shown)


Glandos:
Here is a quick overview of what changed:

https://user-images.githubusercontent.com/297578/124924197-7a811f80-dffb-11eb-9966-adbacae9fea7.mp4

zorun:
It looks really good! :+1: 

Just one thing, are you sure about "noone"? I would write it "no one", or just keep "nobody" like it was before.

Glandos:
[**@zorun**](https://github.com/zorun) commented on [9 juil. 2021 à 01:38 UTC+2](https://github.com/spiral-project/ihatemoney/pull/783#issuecomment-876810525 "2021-07-08T23:38:35Z - Replied by Github Reply Comments"):
> It looks really good! :+1: 
> 
> Just one thing, are you sure about "noone"? I would write it "no one", or just keep "nobody" like it was before.

Yes, it sounds odd, but it is valid. My quick online research found that "Everyone/Noone" is more formal than "Everybody/Nobody", and we were already using "Everyone" and "Everyone but XXX,YYY" in the bill list. So I chose to stick to the same wording. "Noone" was not used, but it is aligned with "Everyone".

And… as I type this message, I found that "Noone" is accepted by some translation machines, but they always produce "No one" with a space. I'll use that form, this is more eye-friendly.

zorun:
Thanks! Yeah "noone" without space looks weird.

PR #784, opened 2021-07-08
Fix SQL query in has_multiple_currencies

zorun:
The previous query was working fine on sqlite and mariadb, but not on
postgresql :

    return self.get_bills_unordered().group_by(Bill.original_currency).count() > 1

    psycopg2.errors.GroupingError: column "bill.id" must appear in the GROUP BY clause or be used in an aggregate function

The failing SQL query was:

    [SQL: SELECT count(*) AS count_1
    FROM (SELECT bill.id AS bill_id, ...
    FROM bill JOIN person ON person.id = bill.payer_id JOIN project ON project.id = person.project_id
    WHERE bill.payer_id = person.id AND person.project_id = project.id AND project.id = %(id_1)s GROUP BY bill.original_currency) AS anon_1]

The problem is that sqlalchemy creates a subquery for postgresql, and the
internal query does not make sense.

GROUP BY is not actually useful, we simply need to count the number of
distinct currencies in the list of bills.

### DIFF
--- a/ihatemoney/models.py
+++ b/ihatemoney/models.py
@@ -219,8 +219,14 @@ class Project(db.Model):
         return self.get_bills_unordered().count() > 0
 
     def has_multiple_currencies(self):
-        """Return if multiple currencies are used"""
-        return self.get_bills_unordered().group_by(Bill.original_currency).count() > 1
+        """Returns True if multiple currencies are used"""
+        # It would be more efficient to do the counting in the database,
+        # but this is called very rarely so we can tolerate if it's a bit
+        # slow. And doing this in Python is much more readable, see #784.
+        nb_currencies = len(
+            set(bill.original_currency for bill in self.get_bills_unordered())
+        )
+        return nb_currencies > 1
 
     def get_bills_unordered(self):
         """Base query for bill list"""


Glandos:
Yes, I went for a simple code, not the smartest one. Good catch.
If this works, it's OK for me.

One other way to do that is to use Python, because it's readable:
```python
return len(set(bill.original_currency for bill in self.get_bills_unordered())) <= 1
```
From a DBA point of of view, it's less performant. However, I highly doubt it will change anything for bill list under 10k entries.

zorun:
Right, I like that the Python version is understandable.

This code is basically only called when changing currencies, so it's not a big problem if it's a bit slow.

So, it pains me a bit because I spent time on the nice ORM solution, but I think I'll go with your version!

Glandos:
[**@zorun**](https://github.com/zorun) commented on [9 juil. 2021 à 01:45 UTC+2](https://github.com/spiral-project/ihatemoney/pull/784#issuecomment-876812729 "2021-07-08T23:45:55Z - Replied by Github Reply Comments"):> 
> So, it pains me a bit because I spent time on the nice ORM solution, but I think I'll go with your version!

I felt this before your answer. So, in good remembrance of your work, here is the version you wrote from the sweat of your brain:

```python
    def has_multiple_currencies(self):
        """Return if multiple currencies are used"""
        return self.get_bills_unordered().group_by(Bill.original_currency).count() > 1
        # It's a bit complicated, but this query is just counting the
        # number of distinct currencies.  We can't easily use a subquery
        # because of PostgreSQL, and grouping is not useful (we want a
        # single result, while group_by would give us a count for each group).
        # Adapted from:
        # https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.count
        nb_currencies = (
            db.session.query(func.count(distinct(Bill.original_currency)))
            .select_from(Bill)
            .join(Person, Project)
            .filter(Bill.payer_id == Person.id)
            .filter(Person.project_id == Project.id)
            .filter(Project.id == self.id)
            .scalar()
        )
        return nb_currencies > 1
```

That's partly a joke. I feel this can help us in the future!

zorun:
Eh, thanks :)

PR #788, opened 2021-07-10
Test CI with minimum supported dependencies

zorun:


zorun:
This does not catch all cases (e.g. having one dependency at latest version and other dependencies at minimum version), but it's already a good start to ensure we have some consistency.

setup.cfg
```diff
@@ -31,27 +31,27 @@ install_requires =
     Flask-Cors~=3.0
     Flask-Mail~=0.9
     Flask-Migrate>=2.5.3,<4  # Not following semantic versioning (e.g. https://github.com/miguelgrinberg/flask-migrate/commit/1af28ba273de6c88544623b8dc02dd539340294b)
-    Flask-RESTful~=0.3
+    Flask-RESTful~=0.3,>=0.3.9
```
Glandos, line 34: Can I suggest:
```suggestion
    Flask-RESTful>=0.3.9,<0.4.0
```

setup.cfg
```diff
@@ -31,27 +31,27 @@ install_requires =
...
     Flask-WTF~=0.14,>=0.14.3  # See b76da172652da94c1f9c4b2fdd965375da8a2c3f
     WTForms>=2.3.1,<2.4
-    Flask>=1.1,<3.0
-    itsdangerous>=1.1,<3.0
+    Flask>=2.0,<3.0
+    itsdangerous>=2.0,<3.0
     Jinja2>=3.0,<4.0
     requests~=2.22
-    SQLAlchemy-Continuum~=1.3
+    SQLAlchemy-Continuum~=1.3,>=1.3.9
```
Glandos, line 42: And:
```suggestion
    SQLAlchemy-Continuum>=1.3.9,<1.4.0
```

Glandos (commented):
I am wondering if it's better to specify bounds with `>=` and `<` instead of a mix of `>=` and `~`.

So I am suggesting some changes, feel free to deny them (with comments!)

Glandos:
That's a super nice enhancement to quality. Thanks!

setup.cfg
symphorien: According to the table in https://pip.pypa.io/en/stable/user_guide/#fixing-conflicting-dependencies `~=0.3,>=0.3.9` is equivalent to `~=0.3.9`

setup.cfg
symphorien: could be simplified in `~=1.3.9`

setup.cfg
zorun: From my understanding, `~=0.3,>=0.3.9` allows `0.4.X` while `~=0.3.9` does not

zorun: > Can I suggest:

The current requirement is a bit different, it's equivalent to:

`Flask-RESTful>=0.3.9,<1.0.0`

which makes sense for semantic versioning.

But in general, I'm all for getting rid of this confusing `~=` requirement, it's too implicit.

PR #790, opened 2021-07-11
tests: ensure that /tmp/ihatemoney.db does not get overwritten

zorun:
Most of the tests are using a separate database, but we have a few tests
that are loading default values and are writing to /tmp/ihatemoney.db.

This is annoying because it's also the database used for development:
running the test suite breaks the dev database.

To fix this, always use a separate testing database to avoid interference.

### DIFF
--- a/ihatemoney/tests/common/ihatemoney_testcase.py
+++ b/ihatemoney/tests/common/ihatemoney_testcase.py
@@ -10,6 +10,9 @@ from ihatemoney.run import create_app, db
 class BaseTestCase(TestCase):
 
     SECRET_KEY = "TEST SESSION"
+    SQLALCHEMY_DATABASE_URI = os.environ.get(
+        "TESTING_SQLALCHEMY_DATABASE_URI", "sqlite://"
+    )
 
     def create_app(self):
         # Pass the test object as a configuration.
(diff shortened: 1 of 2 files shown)


Glandos:
So, if I understand correctly, all tests will now use a non persistent in memory database?

That may solve some issues I had when running tests in parallel…

zorun:
Yes, this should now be all in-memory sqlite, but I don't know if the same one is reused for all test cases or if a new one is created each time.

How do you run tests in parallel?

Glandos:
Using https://pypi.org/project/pytest-xdist/ (installed with `pip install pytest-xdist[psutil]`) and then `pytest -n auto ihatemoney.tests`. This is really faster on a modern machine with many CPUs.

I'll try to see if the database is reused, but it should be easy to drop it between two tests if needed.

zorun:
Nice, it goes from 1 minute to 30 seconds on my desktop, and down to 10 seconds on a big server :) And it seems to work fine with this change.

Should we document it and even put that in the CI?

Glandos:
[**@zorun**](https://github.com/zorun) commented on [14 juil. 2021 à 11:03 UTC+2](https://github.com/spiral-project/ihatemoney/pull/790#issuecomment-879724203 "2021-07-14T09:03:12Z - Replied by Github Reply Comments"):
> Should we document it and even put that in the CI?

We can document it. But no, the CI is not on our infrastructure, so I don't want to put CPU time pressure on it. And I'm not even sure it has multiple CPU (even virtual) available.

Issue #792, opened 2021-07-12
Change Dockerfile to setup the app at build time instead of runtime


wrecker:
The current Dockerfile is written to download the python dependencies and setup the full app when the container starts up. This means everytime the container starts up it goes the whole process of downloading the python dependencies from PyPi and setting up the app. I was trying to bring this up in an isolated environment and realized it wouldn't work. There are other side-effects like users cannot run the `generate_hashed_password` with a built image in a single step. 

A better way to do this would be move the python setup process from entrypoint.sh into the Dockerfile and simplify the entrypoint.sh (and in turn the startup process). If the current behavior is useful for dev purposes, the project can have a separate Dockerfile.dev with a corresponding dev-entrypoint.sh.

wrecker:
I have a PR if you would like to see review it.

zorun:
Hi, welcome! This is definitely a change we want to have, I also don't like the current process (but I'm not a big Docker user). I'll comment on the PR.

For reference, we have another issue about Docker here: #332

ra-ju:
Since the PR #793 has been merged, can we close this issue?

PR #793, opened 2021-07-12
Change Dockerfile to compile python dependencies at build time.

wrecker:
WIP PR for #792

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+.git
+.github
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,6 @@
 FROM python:3.7-alpine
 
-ENV NIGHTLY="" \
-    DEBUG="False" \
+ENV DEBUG="False" \
     SQLALCHEMY_DATABASE_URI="sqlite:////database/ihatemoney.db" \
     SQLALCHEMY_TRACK_MODIFICATIONS="False" \
     SECRET_KEY="tralala" \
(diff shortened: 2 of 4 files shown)


zorun:
This looks good so far, thanks. You can remove the NIGHTLY variable as it's now unused (it's the default)

I'm not convinced we want to keep the old Dockerfile around. For release builds we should probably just build it once and publish it on the dockerhub or something.

Can you also have a look at the documentation and see what needs to be updated? https://ihatemoney.readthedocs.io/en/latest/installation.html#with-docker

I see you bump Flask-SQLalchemy, why did you need that?

wrecker:
> I see you bump Flask-SQLalchemy, why did you need that?

I ran into this SQLAlchemy issue that was causing the startup to fail https://github.com/pallets/flask-sqlalchemy/issues/910#issuecomment-802098285

wrecker:
Ok. Made the changes to remove the original Dockerfile and updated the Docker section of the docs.

zorun:
Thanks. You need to squash your commits before you rebase to master, otherwise you'll get conflicts.

Here are some more suggestions:

- add `.git/` to a `.dockerignore` file to avoid including 40 MB of git history for nothing
- remove all apk packages: there's no need for them now
- `generate_hashed_password` in the doc looks like a typo, it should be `generate_password_hash`

wrecker:
> You need to squash your commits before you rebase to master, otherwise you'll get conflicts.

Done.

> Here are some more suggestions:
> 
>     * add `.git/` to a `.dockerignore` file to avoid including 40 MB of git history for nothing

Good tip!. Done.

>     * remove all apk packages: there's no need for them now
> 
Done.

>     * `generate_hashed_password` in the doc looks like a typo, it should be `generate_password_hash`
Done.

zorun:
Excellent, thanks!

PR #796, opened 2021-07-12
Add more validation to destructive actions (CSRF + additional password checking)

Glandos:
This also switches all such actions to POST requests.

Remaining topics:

- [ ] Use form.errors instead of translated strings
- [ ] Project deletion
- [ ] Project history deletion

### DIFF
--- a/ihatemoney/templates/edit_project.html
+++ b/ihatemoney/templates/edit_project.html
@@ -1,9 +1,17 @@
 {% extends "layout.html" %}
 
 {% block js %}
-    $('#delete-project').click(function ()
-    {
-        $(this).html("<a style='color:red; ' href='{{ url_for('.delete_project') }}' >{{_("you sure?")}}</a>");
+
+    $('#delete-project').click(function(){
+        var link = $(this).find('button');
+        link.click(function(){
+            if ($(this).hasClass("confirm")){
+                return true;
+            }
+            $(this).html("{{_("Are you sure?")}}");
+            $(this).addClass("confirm");
+            return false;
+        });
     });
 
     $('.custom-file-input').on('change', function(event) {
@@ -21,6 +29,10 @@
         {{ forms.edit_project(edit_form) }}
     </form>
 
+    <h2>{{ _("Delete project") }}</h2>
+    <form id="delete-project" class="form-horizontal" action="{{ url_for(".delete_project") }}" method="post" >
+        {{ forms.delete_project(delete_form) }}
+    </form>
 
     <h2>{{ _("Import JSON") }}</h2>
     <form class="form-horizontal" method="post" enctype="multipart/form-data">
--- a/ihatemoney/templates/edit_project.html
+++ b/ihatemoney/templates/edit_project.html
@@ -2,16 +2,20 @@
 
 {% block js %}
 
-    $('#delete-project').click(function(){
-        var link = $(this).find('button');
-        link.click(function(){
-            if ($(this).hasClass("confirm")){
-                return true;
-            }
-            $(this).html("{{_("Are you sure?")}}");
-            $(this).addClass("confirm");
-            return false;
-        });
+    let link = $('#delete-project').find('button');
+    let deleteOriginalHTML = link.html();
+    link.click(function() {
+        if (link.hasClass("confirm")){
+            return true;
+        }
+        link.html("{{_("Are you sure?")}}");
+        link.addClass("confirm btn-danger");
+        return false;
+    });
+
+    $('#delete-project').focusout(function() {
+        link.removeClass("confirm btn-danger");
+        link.html(deleteOriginalHTML);
     });
 
     $('.custom-file-input').on('change', function(event) {
--- a/ihatemoney/templates/forms.html
+++ b/ihatemoney/templates/forms.html
@@ -130,9 +130,6 @@
     {% include "display_errors.html" %}
     {{ form.hidden_tag() }}
     {{ input(form.password, inline=True) }}
-    <div class="actions">
-        <button class="btn btn-danger">{{ _("Confirm deletion") }}</button>
-    </div>
 
 {% endmacro %}
 
@@ -141,9 +138,6 @@
     {% include "display_errors.html" %}
     {{ form.hidden_tag() }}
     {{ input(form.password) }}
-    <div class="actions">
-        <button class="btn btn-danger">{{ _("Confirm deletion") }}</button>
-    </div>
 
 {% endmacro %}
 
--- a/ihatemoney/templates/forms.html
+++ b/ihatemoney/templates/forms.html
@@ -111,7 +111,7 @@
     {{ form.hidden_tag() }}
     {{ input(form.password) }}
     <div class="actions">
-        <button class="btn btn-danger">{{ _("Delete project") }}</button>
+        <button class="btn btn-primary">{{ _("Delete project") }}</button>
     </div>
 
 {% endmacro %}
--- a/ihatemoney/templates/forms.html
+++ b/ihatemoney/templates/forms.html
@@ -100,7 +100,18 @@
     {{ input(form.default_currency) }}
     <div class="actions">
         <button class="btn btn-primary">{{ _("Edit the project") }}</button>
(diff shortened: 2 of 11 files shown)


Glandos:
I was surprised by the fact that on authentication form, the CSRF error message is displayed directly from Flask_WTF, but this is always in English. In fact, there is an issue on that: https://github.com/wtforms/flask-wtf/issues/401
The main issue with providing our own string is that we lose some information, since Flask_WTF is using different strings for different cases. And this error **should** never happen in a normal environment, so displaying it in English is not a real issue.
On the other hand, it seems better to have a translated application.

What do you think?

zorun:
I have reworked project deletion to have its own form and to ask for the private code:

2021-07-14-161729_1245x817_scrot

zorun:
Same thing for the "Clear project history" modal, but I need help from someone who knows how to talk to bootstrap:

2021-07-14-161829_520x343_scrot

zorun:
> I was surprised by the fact that on authentication form, the CSRF error message is displayed directly from Flask_WTF, but this is always in English. In fact, there is an issue on that: [wtforms/flask-wtf#401](https://github.com/wtforms/flask-wtf/issues/401)
> The main issue with providing our own string is that we lose some information, since Flask_WTF is using different strings for different cases. And this error **should** never happen in a normal environment, so displaying it in English is not a real issue.
> On the other hand, it seems better to have a translated application.
> 
> What do you think?

Yes, I was also surprised about the built-in CSRF error messages.

I think it's fine to handle errors through flash messages, it shouldn't happen in normal cases (and practically speaking it's hard to handle this as form errors...)

Let me know what you think of the new code.

ihatemoney/templates/edit_project.html
```diff
@@ -1,9 +1,17 @@
...
+    $('#delete-project').click(function(){
+        var link = $(this).find('button');
+        link.click(function(){
+            if ($(this).hasClass("confirm")){
+                return true;
+            }
+            $(this).html("{{_("Are you sure?")}}");
+            $(this).addClass("confirm");
+            return false;
+        });
```
Glandos, line 14: The event listener is added every time the form is clicked. Let's simplify this:
```suggestion
    let link = $('#delete-project').find('button');
    let deleteOriginalHTML = link.html();
    link.click(function(){
        if (link.hasClass("confirm")){
            return true;
        }
        link.html("{{_("Are you sure?")}}");
        link.addClass("confirm btn-danger");
        return false;
    });

    $('#delete-project').focusout(function() {
        link.removeClass("confirm btn-danger");
        link.html(deleteOriginalHTML);
    });
```

ihatemoney/templates/forms.html
```diff
@@ -100,7 +100,18 @@
...
+{% endmacro %}
+
+{% macro delete_project(form) %}
+
+    {% include "display_errors.html" %}
+    <p><strong>{{ _("This will remove all bills and participants in this project!") }}</strong></p>
+    {{ form.hidden_tag() }}
+    {{ input(form.password) }}
+    <div class="actions">
+        <button class="btn btn-danger">{{ _("Delete project") }}</button>
```
Glandos, line 114: At first the button is normal. Confirmation will become `btn-danger`.
```suggestion
        <button class="btn btn-primary">{{ _("Delete project") }}</button>
```

Glandos (commented):
For the edit project form, I changed the interaction:
- Button is normal at first, and changed to danger when clicked. Otherwise, it is hard to notice a change with the mouse over it.
- If the button lose focus, confirmation state is lost.
- Attach the event at document load, and not on each form click.

ihatemoney/web.py
```diff
@@ -511,11 +515,18 @@ def import_project(file, project):
...
-    flash(_("Project successfully deleted"))
-
+    form = DestructiveActionProjectForm(id=g.project.id)
+    if form.validate():
+        g.project.remove_project()
+        flash(_("Project successfully deleted"))
+        return redirect(url_for(".home"))
+    else:
+        flash(
+            _("Error deleting project: wrong private code or wrong CSRF token"),
```
Glandos, line 527: Use `flash_form_errors`

ihatemoney/web.py
```diff
@@ -644,6 +657,12 @@ def add_member():
 
 @main.route("/<project_id>/members/<member_id>/reactivate", methods=["POST"])
 def reactivate(member_id):
+    # Used for CSRF validation
+    form = EmptyForm()
+    if not form.validate():
+        flash(_("CSRF Token: The CSRF token is invalid."), category="danger")
```
Glandos, line 663: Use `flash_form_errors`

ihatemoney/web.py
```diff
@@ -658,6 +677,12 @@ def reactivate(member_id):
 
 @main.route("/<project_id>/members/<member_id>/delete", methods=["POST"])
 def remove_member(member_id):
+    # Used for CSRF validation
+    form = EmptyForm()
+    if not form.validate():
+        flash(_("CSRF Token: The CSRF token is invalid."), category="danger")
```
Glandos, line 683: Use `flash_form_errors`

ihatemoney/web.py
```diff
@@ -715,9 +740,14 @@ def add_bill():
...
 
 
-@main.route("/<project_id>/delete/<int:bill_id>")
+@main.route("/<project_id>/delete/<int:bill_id>", methods=["POST"])
 def delete_bill(bill_id):
-    # fixme: everyone is able to delete a bill
+    # Used for CSRF validation
+    form = EmptyForm()
+    if not form.validate():
+        flash(_("CSRF Token: The CSRF token is invalid."), category="danger")
```
Glandos, line 748: Use `flash_form_errors`

ihatemoney/web.py
```diff
@@ -781,27 +812,48 @@ def history():
...
+        delete_form=delete_form,
     )
 
 
 @main.route("/<project_id>/erase_history", methods=["POST"])
 def erase_history():
     """Erase all history entries associated with this project."""
+    form = DestructiveActionProjectForm(id=g.project.id)
+    if not form.validate():
+        flash(
```
Glandos, line 824: Use `flash_form_errors`

ihatemoney/web.py
```diff
@@ -781,27 +812,48 @@ def history():
...
+    flash(_("Deleted project history."))
     return redirect(url_for(".history"))
 
 
 @main.route("/<project_id>/strip_ip_addresses", methods=["POST"])
 def strip_ip_addresses():
     """Strip ip addresses from history entries associated with this project."""
+    form = DestructiveActionProjectForm(id=g.project.id)
+    if not form.validate():
+        flash(
```
Glandos, line 843: Use `flash_form_errors`

Glandos (commented):
Forms are already translating interesting strings, so I'd like to use them instead of flashing a new message.
For this, we need a function (in `utils.py`?) like this:
```python

def flash_form_errors(form, title):
    if len(form.errors) == 0:
        return None
    elif len(form.errors) == 1:
        # I18N: Form error with only one error
        return _("{title}: {description}").format(title=title, error=form.errors.popitem()[1][0])
    else:
        error_list = "</li><li>".join(str(error) for (field, errors) in form.errors.items() for error in errors)
        errors = f"<ul><li>{error_list}</li></ul>"
        # I18N: Form error with a list of errors
        return Markup(_("{title}<br />{errors}").format(title=title, errors=errors))
```

And it can be used:
```python
        flash(
            flash_form_errors(form, _("Error deleting project history")),
            category="danger",
        )
```
Then all strings from validators are displayed as-is, potentially in a list, if there more than one of them. I voluntarily removed the field name, otherwise the flashing message would have been too long.

Glandos:
I'll gladly provide my help for the form, but this will for another evening.

ihatemoney/templates/edit_project.html
zorun: I'm a javascript noob, so I trust you on this ;)

ihatemoney/web.py
zorun: Good idea!

zorun:
Many thanks for the review! I have integrated all your suggestions, and I figured out what was wrong with the form layout.

PR #799, opened 2021-07-13
Update the Flask-SQLAlchemy version to v2.5

### LINKED ISSUE #798: Update Flask-SQLAlchemy version in setup.cfg
This is needed to pickup the fix made in Flask-SQLAlchemy https://github.com/pallets/flask-sqlalchemy/issues/910

wrecker:
The upstream SQLAlchemy project made a breaking change in v1.4 and
v2.5 of Flask-SQLAlchemy is now needed.

Fixes #798

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,7 +32,7 @@ install_requires =
     Flask-Mail~=0.9
     Flask-Migrate>=2.5.3,<4  # Not following semantic versioning (e.g. https://github.com/miguelgrinberg/flask-migrate/commit/1af28ba273de6c88544623b8dc02dd539340294b)
     Flask-RESTful~=0.3
-    Flask-SQLAlchemy~=2.4
+    Flask-SQLAlchemy~=2.5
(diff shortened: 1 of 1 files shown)


Glandos:
Hi, and thanks for your involvement in this project (especially Docker, that I don't use :wink:)

There is something that I don't understand. Our current `setup.cfg` specifically pin SQAlchemy strictly below 1.4, because SQLAlchemy-continuum is not yet compatible. Does it get installed somehow?

Note however that we plan to use SQLAlchemy 1.4 whenever it will be possible, so your finding is still valuable.

wrecker:
@Glandos Ignore this change.. When I initially ran into this issue,  I was using the Dockerfile before the changes in #793. At that point the build was getting IHM from PyPI - this version of IHM didn't have a pinned version for SQLAlchemy and so it was installing v1.4. This led to the error and I ended up fixing it with the Flask-SQLAlchemy change. 
But when building from the HEAD of the repo, this is no longer an issue. 
So the current setup.cfg with SQL-Alchemy pinned to v1.3 is working fine in my setup.

Glandos:
It is then more than needed that we release a new version. Thanks for this investigation.

PR #805, opened 2021-07-17
Display all languages in dropdown list, including current one

zorun:
Currently we remove the current language from the list.

Changing the list depending on the language is quite confusing, it's
better if it's always the same.

### DIFF
--- a/ihatemoney/templates/layout.html
+++ b/ihatemoney/templates/layout.html
@@ -65,9 +65,7 @@
           <div class="dropdown-menu" aria-labelledby="langMenuButton">
               <h6 class="dropdown-header">{{ _('Languages') }}</h6>
               {% for lang in config['SUPPORTED_LANGUAGES'] %}
-                {% if g.lang != lang %}
-                  <a class="dropdown-item" href="{{ url_for("main.change_lang", lang=lang)}}">{{ locale_from_iso(lang).display_name | capitalize }}</a>
-                {% endif %}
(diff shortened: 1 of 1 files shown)


ihatemoney/templates/layout.html
```diff
@@ -65,9 +65,7 @@ <h1><a class="navbar-brand" href="{{ url_for("main.home") }}"><span>#!</span> mo
           <div class="dropdown-menu" aria-labelledby="langMenuButton">
               <h6 class="dropdown-header">{{ _('Languages') }}</h6>
               {% for lang in config['SUPPORTED_LANGUAGES'] %}
-                {% if g.lang != lang %}
                   <a class="dropdown-item" href="{{ url_for("main.change_lang", lang=lang)}}">{{ locale_from_iso(lang).display_name | capitalize }}</a>
-                {% endif %}
```
Glandos, line 70: What about writing a mark in front of the current language?

```html
<a class="dropdown-item" href="{{ url_for("main.change_lang", lang=lang)}}">
  {% if g.lang == lang %}
  <span class="select-mark">•</span>
  {% endif %}
  {{ locale_from_iso(lang).display_name | capitalize }}
</a>
```

And in CSS we need:
```css
.dropdown-item > .select-mark {
  margin-left: -14px;
  margin-right: 4px;
}
```

Rendered:
Screenshot_2021-07-18 Gestion de comptes - demonstration

I chose an Unicode character, but feel free to suggest anything else, it's just the idea.

zorun: Right, good idea! I've simply added the "active" bootstrap class:

2021-07-18-143017_274x784_scrot

Issue #811, opened 2021-07-18
Language list is not scrollable on mobile


zorun:
I just tested on Android (with a somewhat weird browser), and I cannot scroll in the language list. French is high enough in the list to be selectable, but many languages are not that lucky.

See screenshot:

![image](https://user-images.githubusercontent.com/731144/126067813-94ddbaf3-8477-4bc6-9c09-9594a16d21d8.png)

shreyash00007:
hey, I just got the fix of it you can add the some code to the class "navbar-collapse"
    
  .navbar-collapse {
    -ms-flex-preferred-size: 100%;
    flex-basis: 100%;
    -ms-flex-positive: 1;
    flex-grow: 1;
    -ms-flex-align: center;
    align-items: center;
    margin: 5px;
    padding: 5px;
    background-color: #343a40;
    width: 500px;      //adjust in desktop view
    height: 110px;        //adjust in desktop view
    overflow: auto;
    text-align: justify;
}
I tested it works, but adjust the width properly as per the desktop view .

Flecart:
@shreyash00007 you have to open a pull request to collaborate!
Fork the repository, make new branch and open pull request

Glandos:
[**@Flecart**](https://github.com/Flecart) commented on [18 juil. 2021 à 18:47 UTC+2](https://github.com/spiral-project/ihatemoney/issues/811#issuecomment-882084809 "2021-07-18T16:47:43Z - Replied by Github Reply Comments"):
> @shreyash00007 you have to open a pull request to collaborate!
> Fork the repository, make new branch and open pull request

For a simple file edition, you can even:
- Navigate to the file, display it
- Click on the “Pen” icon at the top right
- Edit & save
- Submit the PR

For small contributions, this is really a good way to go.

zorun:
Fixed by #813, thanks.

PR #813, opened 2021-07-18
Fix language list scrollable on mobile

olejech:
#811

### DIFF
--- a/ihatemoney/static/css/main.css
+++ b/ihatemoney/static/css/main.css
@@ -536,6 +536,10 @@ tr:hover .extra-info {
 .navbar-dark .navbar-nav .show > .nav-link svg {
   fill: white;
 }
+.navbar-dark .dropdown-menu {
+  height: 50vh;
+  overflow-y: auto;
+}
 .icon svg {
   display: inline-block;
   border-bottom: 0.2em solid transparent;


ihatemoney/static/css/main.css
```diff
@@ -536,6 +536,10 @@ tr:hover .extra-info {
 .navbar-dark .navbar-nav .show > .nav-link svg {
   fill: white;
 }
+.navbar-dark .dropdown-menu {
+  height: 50vh;
```
Glandos, line 540:
```suggestion
  max-height: 60vh;
```
Use max-height to have a menu smaller if needed.
I also put 60vh, because on large screens, it's a bit confusing to have "only" `50vh`. But maybe `60vh` is too big on small screens?

Glandos:
Thanks for this quick and lean solution!

ihatemoney/static/css/main.css
zorun: I was about to ask about this 50vh, because it can already be too big for very small screens...

Glandos: Screenshot_2021-07-18 Account manager - demonstration
I don't think it's too big, but I'm not a real smartphone user :thinking:

zorun:
Is it possible to use bootstrap classes to only apply it on small screens?

ihatemoney/static/css/main.css
zorun: Right, 50vh is the maximum that fits on my phone.

zorun:
Merged with max-height and 50vh, thanks for the fix!

PR #814, opened 2021-07-18
Add currency to import and export formats

zorun:


ihatemoney/web.py
```diff
@@ -447,17 +448,45 @@ def import_project(file, project):
...
+    attr = ["what", "payer_name", "payer_weight", "amount", "currency", "date", "owers"]
     attr.sort()
+    currencies = set()
     for e in json_file:
+        # Add compatibility with versions < 5 that did not have currencies
+        if "currency" not in e:
+            e["currency"] = CurrencyConverter.no_currency
+        # Empty currency should be converted to "XXX"
+        if e["currency"] == "":
+            e["currency"] = CurrencyConverter.no_currency
```
Glandos, line 460:
```suggestion
        # Add compatibility with versions < 5 that did not have currencies
        # Empty currency should be converted to "XXX"
        if e.get("currency", "") == "":
            e["currency"] = CurrencyConverter.no_currency
```

Just to have a better readability

Glandos, line 460:
```suggestion
        # If currency is absent, empty, or explicitly set to XXX
        # set it to project default.
        if e.get("currency", "") in ["", "XXX"]:
            e["currency"] = project.default_currency
```

ihatemoney/web.py
```diff
@@ -447,17 +448,45 @@ def import_project(file, project):
...
         list_attr.sort()
         if list_attr != attr:
-            raise ValueError
+            raise ValueError(_("Invalid JSON"))
+        # If the project has a default currency, convert bills that have no currency
+        if (
+            project.default_currency != CurrencyConverter.no_currency
+            and e["currency"] == CurrencyConverter.no_currency
+        ):
+            e["currency"] = project.default_currency
```
Glandos, line 474:
```suggestion
```
This part has been resolved by the previous step.

ihatemoney/web.py
```diff
@@ -447,17 +448,45 @@ def import_project(file, project):
...
+        # If bills have currencies, they must be consistent
+        if len(currencies - {CurrencyConverter.no_currency}) >= 2:
+            raise ValueError(
+                _(
+                    "Cannot add bills in multiple currencies to a project without default currency"
+                )
+            )
+        # Strip currency from bills (since it's the same for every bill)
+        for e in json_file:
+            e["currency"] = CurrencyConverter.no_currency
```
Glandos, line 489: This seems complicated. I would say that if a project is set to no currency, then no other currency is allowed in the imported file.

In this case, currencies don't have to be tracked. A simple check in the previous loop can raise an exception.

Glandos (changes requested):
I would simplify the import logic.

ihatemoney/web.py
zorun: The current code would allow to export a project using EUR, and then re-import it in a project with no currency. But I agree this could also be done by first converting the initial project to "no currency" and then exporting.

I guess I like Postel's law too much :) https://en.wikipedia.org/wiki/Robustness_principle

zorun: Still, being too strict during an import can be a problem. If a project ends up with inconsistent currencies (through the API or because of a bug), then exporting it and reimporting it would fail. This is bad.

If we are 100% sure that currencies are always consistent within a project, then I'm OK to be strict when importing.

Glandos: For me the main issue is not really the robustness principle, but the user experience. Let's summarize edge cases when importing into a "no currency" project:
- Import bills with different currencies: it must fail
- Import bills with `XXX` or empty currency: it must succeed
- Import bills with the same currency: user expects success, even if it's not `XXX`
- Import bills with a mix of `XXX` and another single currency: user expects success, but honestly, what happened? Manual edition or merging?

So... Yes, because of the 3rd point, I would say that your code fits better.

PR #818, opened 2021-07-25
"a free software" -> "free software"

georgefst:
In English, "software" is an [uncountable noun](https://dictionary.cambridge.org/grammar/british-grammar/uncountable-nouns). You could say "a free piece of software" or "some free software", but in this context it feels most natural to just drop the article entirely.

### DIFF
--- a/ihatemoney/templates/layout.html
+++ b/ihatemoney/templates/layout.html
@@ -159,7 +159,7 @@
 
           <div class="footer-left">
             <p>
-              <a href="https://github.com/spiral-project/ihatemoney">{{ _("\"I hate money\" is free software") }}</a><span class="d-none d-sm-inline"></span>,
+              <a href="https://github.com/spiral-project/ihatemoney">{{ _("\"I hate money\" is a free software") }}</a><span class="d-none d-sm-inline"></span>,
               {{ _("you can contribute and improve it!") }}</span>
             </p>
           </div>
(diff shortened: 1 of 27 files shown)


ihatemoney/templates/layout.html
```diff
@@ -159,7 +159,7 @@ <h6 class="dropdown-header">{{ _('Languages') }}</h6>
 
           <div class="footer-left">
             <p>
-              <a href="https://github.com/spiral-project/ihatemoney">{{ _("\"I hate money\" is a free software") }}</a><span class="d-none d-sm-inline"></span>,
+              <a href="https://github.com/spiral-project/ihatemoney">{{ _("\"I hate money\" is free software") }}</a><span class="d-none d-sm-inline"></span>,
```
Glandos, line 162: Only this part should be left in the PR

Glandos (changes requested):
Thanks for the limpid explanation, and the correction.

However, `.pot` and `.po` files are managed automatically by our Weblate instance, so the must be excluded from the PR itself.

georgefst:
> However, `.pot` and `.po` files are managed automatically by our Weblate instance, so the must be excluded from the PR itself.

Ah okay, I'll have a proper look at that tomorrow. Is it easy to make bulk edits?

Glandos:
> > However, `.pot` and `.po` files are managed automatically by our Weblate instance, so the must be excluded from the PR itself.
> 
> Ah okay, I'll have a proper look at that tomorrow. Is it easy to make bulk edits?

Yes, no problem. Once the string is edited in the source file, we'll run the string extractor that will re-generate all `.po(t)` files.

georgefst:
(this should be squash-merged)

PR #819, opened 2021-07-25
Miscellaneous improvements to installation instructions

georgefst:
These would have stopped me from making a few mistakes when trying to install.

(After about an hour, I hit some inscrutable SQL error, and decided to use the hosted version, donate, and move on with my life. But hopefully I can make the path a bit clearer for the next person.)

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -26,7 +26,7 @@ format used can be found on `the SQLAlchemy documentation`_.
 
 For example, if you're using MariaDB, use a configuration similar to the following::
 
-    SQLALCHEMY_DATABASE_URI = 'mysql+pymysql://user:pass@localhost/dbname'
+    SQLALCHEMY_DATABASE_URI = 'mysql+pymysql://user:pass@localhost/mydb'
 
 If you're using PostgreSQL, your client must use utf8. Unfortunately,
 PostgreSQL default is to use ASCII. Either change your client settings,
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -33,7 +33,7 @@ PostgreSQL default is to use ASCII. Either change your client settings,
 or specify the encoding by appending ``?client_encoding=utf8`` to the
 connection string. This will look like::
 
-    SQLALCHEMY_DATABASE_URI = 'postgresql://myuser:mypass@localhost/dbname?client_encoding=utf8'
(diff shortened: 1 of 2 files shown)


georgefst:
> decided to use the hosted version, donate, and move on with my life

And of course, I'll buy you a beer if you ever find yourself in West London...

docs/configuration.rst
```diff
@@ -33,7 +33,7 @@ PostgreSQL default is to use ASCII. Either change your client settings,
 or specify the encoding by appending ``?client_encoding=utf8`` to the
 connection string. This will look like::
 
-    SQLALCHEMY_DATABASE_URI = 'postgresql://myuser:mypass@localhost/dbname?client_encoding=utf8'
+    SQLALCHEMY_DATABASE_URI = 'postgresql://myuser:mypass@localhost/mydb?client_encoding=utf8'
```
Glandos, line 36: Is there any rationale for that change? In this case, the example for MariaDB should be adapted in the same way.

docs/installation.rst
```diff
@@ -36,15 +36,18 @@ most small to medium setups.
...
-Choose an installation path, here `/home/john/ihatemoney`.
+Choose an installation path, here the current user's home directory (`~`).
 
 Create a virtual environment::
 
-    python3 -m venv /home/john/ihatemoney
+    cd ~
+    mkdir ihatemoney
+    cd ihatemoney
+    python3 -m venv .
```
Glandos, line 46: That's more generic indeed, but a bit too long:
```suggestion
    python3 -m venv ~/ihatemoney
    cd ~/ihatemoney
```

The `cd` can be omitted if we keep an absolute path for the following activation.

Glandos (changes requested):
Documentation! Neat!

However, I have some (little) questions :smile:

docs/installation.rst
georgefst: Sure. I have a slight preference for doing things in the current directory, with relative paths, but I don't feel strongly about it.

docs/configuration.rst
georgefst: Because it matches this:
https://github.com/spiral-project/ihatemoney/blob/6bacae1cd9d136e99a818b0e7f4f1f829803ed88/docs/installation.rst#L101

I know the intention is for the user to use their own names, but in practice they may not. And `myuser` and `mypass` match already, so we may as well be consistent.

Glandos:
Thanks!

PR #832, opened 2021-09-01
Update the email template to feel more inclusive.

almet:


Glandos:
Erf… I don't know why action failed. It says that module `dateutil` is not found, but it is part of the standard library…

almet:
That's weird, but adding it as a direct dependency fixes it. I wonder if some dependency we used dropped its use?

almet:
Ah nice, I didn't know about `pip show` =D.

Is it okay for you if we merge this as part of this PR?

Glandos:
It's not conventional, but the merge will link here for explanation. Let's go.

PR #839, opened 2021-09-15
Fix #838: Round weight to 2 decimal cases

### LINKED ISSUE #838: Weight is being rounded on balance member list
Currently the user's weight are being rounded on to 1 decimal case only, so if we select 0.75 as the weight for one user and 0.25 for the other we see on the balance member list as 0.8 and 0.2 image below:
image

DanielNetoP:
After this fix the weight values are kept and are not rounded
image


Glandos:
That was quick!

Just run `make black` and you're done.

By the way @zorun do you have an opinion on this? For me, it's all good.

DanielNetoP:
@Glandos Should be ready now 🤞

zorun:
This has bugged me in the past, so definitely +1 for this change!

Issue #841, opened 2021-10-05
Protect project creations with a captcha


almet:
The hosted version of ihatemoney.org is currently spammed by robots, and AlwaysData (our hoster) disabled project creation until we are able to filter them out.

I believe we could solve this issue by using a captcha. I don't want to use ReCaptcha or proprietary solutions, though. I unfortunately don't have so much spare time these days, so if anybody want to take this over that would be awesome.

I did some research, and I think we could integrate with https://pypi.org/project/flask-session-captcha/, but we will need to make it compatible with Flask-WTF.

Glandos:
Honestly, an optional CAPTCHA is not a bad idea. But I can only put a veto for a solution with reCaptcha.
However, there are alternatives:
- https://www.hcaptcha.com/ Used by CloudFlare. It's also a service from a big company.
- https://friendlycaptcha.com/ The demo seems interesting. Especially because it has a warning when JS is disabled
- A custom field. I gathered a lot of positive feedback from a simple thing like this. You ask a simple question like "Which is a real currency: Euro or Petro dollar?". It won't block targeted bots, but it will block automated ones.

almet:
Hi, I dont like proprietary solutions either.

We have spammers and our hoster service asks us to filter them. They do not care about the solution as long as It works.

The impact on them is increased database size and outflux of emails.

Unless someone would like to tackle this (I dont), I'm fine with landing this. It's off by default and solves an existing issue.

If somebody wants to take time to investigate further that would be awesome !

PR #842, opened 2021-10-05
Add optional support for Google ReCAPTCHA service.

almet:
Cf #841


almet:
I know this is not ideal, as it uses a google service. I consider it to be okay as it's optional. I'll be glad to see something done without the help of this service, but I went for the easy solution first in order to re-enable the online service.

I think I'm going to deploy this as a quick fix, until we find a better solution.

almet:
Not sure how to solve this black issue. Ideas?

PR #843, opened 2021-10-10
Change token path authentication to /PROJECT/join/TOKEN

Glandos:
See #802 for initial conversation.

I left `project_id/token` since we need to validate that the project exist before validating the token.

Otherwise, it means extracting it from the token without verifying the signature in a first step, and I'm not comfortable with this. With this, the project is checked with `pull_project` preprocessor, as every other endpoint.

Discussion is opened :smile:

### DIFF
--- a/ihatemoney/web.py
+++ b/ihatemoney/web.py
@@ -227,7 +227,7 @@ def authenticate(project_id=None):
         form.id.data = request.args["project_id"]
     project_id = form.id.data
 
-    project = Project.query.get(project_id)
+    project = Project.query.get(project_id) if project_id is not None else None
     if not project:
         # If the user try to connect to an unexisting project, we will
         # propose him a link to the creation form.


almet:
Thanks for opening a new discussion on this.

> Otherwise, it means extracting it from the token without verifying the signature in a first step, and I'm not comfortable with this.

Could you please elaborate on why this would be bad from your standpoint? Thanks :-)

If the URL is related to a project id, then the name could be `{project_id}/join/{token}` (or without `{project_id}` if we decide against it)

almet:
> If the URL is related to a project id, then the name could be {project_id}/join/{token} (or without {project_id} if we decide against it)

Disregard this, I haven't understood this was a pull request with code attached and was thinking about it the wrong way.

I find it a bit hackish to check on the dot in the URL, but that could work!

almet:
I feel we should go this way with /join. The odds that this is used are fairly low in my opinion. I see two ways to solve this :

- Create a migration in the database to see if the "join" project exists, and if so tell the operator to do something about it. We could even migrate it to another name and send en email to the contact email of the "join" project ; 
-  When /join is used without a token, redirect to the project page. That could be the simpler solution, because `/join` without anything after is not used, only `/join/{token}`.

But… whatever works! :-)

Glandos:
[**@almet**](https://github.com/almet) commented on [10 oct. 2021, 22:02 UTC+2](https://github.com/spiral-project/ihatemoney/pull/843#issuecomment-939543883 "2021-10-10T20:02:56Z - Replied by Github Reply Comments"):
> -  When /join is used without a token, redirect to the project page. That could be the simpler solution, because `/join` without anything after is not used, only `/join/{token}`.

In this case we need to keep the regexp converter for `/join/settle_bills`, `/join/statistics` and all other subpaths for a project. I am really lacking an argument to decide a best solution here :)

almet:
Or it could mean that we need to change the URL structure altogether but… yeah, this would kinda suck. Sorry for making this wrong choice in the first place.

> I feel like /<project_id>/join/<token> is the most clear and clean way.

Okay, I think I'm coming to the same conclusion. I would be against the `/j`. After all explicit is better than implicit :-)

Glandos:
The last change here is on invalid token. At first, I mimic the old behavior by redirecting to the authentication form, but it was error prone from both the code and the user point of view.
Instead, let's redirect to the home with a flash message.

PR #844, opened 2021-10-10
Add optional support for a simple CAPTCHA.

almet:
Add support for a simple CAPTCHA. Alternate solution for #841.


almet:
Thanks for taking over @Glandos :-)

PR #845, opened 2021-10-10
Implement security best practices using Flask-Talisman

zorun:


zorun:
> Looks good to me. Thanks, didn't know about Flask-Talisman, it's long overdue… :-)

Hey, thanks for your review :) Yes, Flask-Talisman is kind of magical, but fortunately simple enough to trust it...

What do you think about the default setting?

zorun:
I was concerned about the dev environment, because it's using plain HTTP. But browsers apparently have exceptions and they accept to send secure cookies to `localhost` over HTTP, which should cover most dev cases. I have added a note in the documentation for the few people not developping over localhost.

So, let's merge this!

PR #850, opened 2021-10-10
Fix currency handling in API

zorun:
Most notably, we broke backwards compatibility of the API without realizing it.


zorun:
@almet did you try adding a bill through MoneyBuster? Some tests are still broken and I think that's when we try to create bills. I'll have a closer look in the coming days hopefully.

PR #851, opened 2021-10-10
Continuous integration build and push to dockerhub

youegraillot:
This pr adds continuous integration with github actions
Image is built and pushed to dockerhub using secrets (DOCKERHUB_USERNAME, DOCKERHUB_REPOSITORY, DOCKERHUB_PASSWORD) each time a new tag is pushed. Multiple tags will be available : versions and 'latest'

Additionally, the dockerhub description is updated to mirror the repository one
Using the account password is needed (instead of a token, more at https://github.com/peter-evans/dockerhub-description), and the README file has been updated to markdown for compatibility

The last changes can be reverted if only pushing the image is desired (updating the description would be a manual process) or using a token is preferred


YoueGraillot:
You can check out the result at : https://hub.docker.com/r/youegraillot/ihatemoney

almet:
> As far as I know, the README is used for the pypi page, that's why it's in rst format. It's probably better to have a small description specific to the docker hub and point to the existing documentation.

I just checked and [PyPI supports Markdown for READMEs](https://packaging.python.org/guides/making-a-pypi-friendly-readme/). But I tend to agree that we should probably put some docker specific instructions in there, if that's possible!

YoueGraillot:
Yeah I assumed you would create an new account to make it more "official"
I'd like you to add me to the collaborators so I can edit the description : youegraillot

almet:
Just added you there.

almet:
I just did. I also added you there. Let me know if you need something else :-)

PR #852, opened 2021-10-12
Fix docker image and docker image test

youegraillot:
- Fix missing env SESSION_COOKIE_SECURE in docker image, preventing launch
- Run basic test before pushing the image (ping the home page from another image)
- Run pipeline for 'master'
- Run pipeline for pr (only test)
- Add an example docker-compose.yml

I'm planning to fix/add some functionalities to the image for compatibility (rootless containers, podman) in a following pr
But for now the container launch successfully, and we'll know in a minute if a change break it :)

Also I'm still wondering how you manage the '4.1-stable' branch, image is still crashing there, should we merge ?
It would be nice to run the pipeline for the 4.1.5, and have an image available on dockerhub

### DIFF
--- a/.github/workflows/dockerhub.yml
+++ b/.github/workflows/dockerhub.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
-      - name: Test image
+      - name: Test image # Checks we are able to run the container.
         run: docker compose -f docker-compose.test.yml run sut
 
   build:
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -1,3 +1,4 @@
+# This is used by Github actions to ensure docker images build properly.
 version: "3.9"
 
 services:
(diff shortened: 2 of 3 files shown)


zorun:
Oops, good catch for the missing SESSION_COOKIE_SECURE, I always forget to add new settings in the Dockerfile.

I think variables are getting out of control: we already had 4 places where we define them, and that's now 5! If I understand correctly, the docker-compose file sets example variables that are supposed to be changed by the person using it? But docker-compose uses the Dockerfile which also has example variables? I'm feeling we're missing some simplification somewhere, but I'm no expert on containers.

Other than that, it looks very good, thanks!

zorun:
And yes, stable-4.1 is a branch where we backported some fixes, but we don't plan any new release on it. Next step is 5.0.0!

almet:
I've added some comments. If everything is green then we should merge :-)

PR #856, opened 2021-10-13
Flash messages must be dimissed manually

### LINKED ISSUE #849: Error (aka danger) flash message should be persistent.
All flash messages are not equals. A normal message could disappear after a reasonable delay, but an error should be manually dismissed. Otherwise, it's confusing for the user.

Glandos:
Fix #849

There is now even less JS as stated in https://github.com/spiral-project/ihatemoney/pull/292#issue-284277396

Note however that it uses plugin from bootstrap, so… JS is used anyway.


zorun:
Looks good, thanks!

PR #871, opened 2021-10-19
Rename all strings to "participants".

almet:


almet:
@Glandos this brings the question of API backwards compatibility. We need to have something which still work for the existing clients.

We can have multiple endpoints with the same goal without issues, because the only client for these URLs is Flask itself. So why not changing everything to participants. 

**But** for the REST API, this is kinda harder. We probably want to have multiple versions of the API running at the same time, with different names for the same ideas. I wonder if it's that's so important to be "pure" here.

The amount of work this will require from us, money buster and cospend is quite large, and for quite no benefits. So… we might want to keep the API "as-is". I'm open for discussion though :-)

almet:
Thanks for your feedback though :-) This needs to be rebased and we can change the commit message as @Glandos pointed out.

Glandos:
OK for the API, let's not change it, it doesn't worth it.

PR #873, opened 2021-10-19
Improve pagination style

zorun:
The style for top and bottom pagination was different, and the top part
was confusing.

Also translate bottom pagination text.


zorun:
Before: https://github.com/spiral-project/ihatemoney/pull/480#issuecomment-587176604

After (with only 2 items per page for screenshot): 
ihatemoney-new-pagination-style

PR #879, opened 2021-10-20
Update release instructions.

almet:


zorun:
Thanks, looks good!

I think we should remove `Update known good versions of dependencies`: it's definitely not a good idea to change that just before a release...

almet:
Yeah, you're right, it's not the proper time to do this :-)

PR #885, opened 2021-10-20
Display monthly statistics for the range of months where the project was active

zorun:
Currently, we display a hard-coded "one year" range of monthly statistics
starting from today.  This generally is not the intended behaviour: for
instance, on an archived project, the bills might all be older than one
year, so the table only displays months without any operation.

Instead, display all months between the first and last bills.  There might
be empty months in the middle, but that's intended, because we want all
months to be consecutive.

If there are no bills, simply display an empty table.

### DIFF
--- a/ihatemoney/models.py
+++ b/ihatemoney/models.py
@@ -1,5 +1,5 @@
 from collections import defaultdict
-from datetime import datetime
+import datetime
 
 from dateutil.parser import parse
 from debts import settle
@@ -608,8 +608,8 @@ class Bill(db.Model):
     owers = db.relationship(Person, secondary=billowers)
 
     amount = db.Column(db.Float)
-    date = db.Column(db.Date, default=datetime.now)
-    creation_date = db.Column(db.Date, default=datetime.now)
+    date = db.Column(db.Date, default=datetime.datetime.now)
+    creation_date = db.Column(db.Date, default=datetime.datetime.now)
     what = db.Column(db.UnicodeText)
     external_link = db.Column(db.UnicodeText)
 
@@ -623,7 +623,7 @@ class Bill(db.Model):
     def __init__(
         self,
         amount: float,
-        date: datetime = None,
+        date: datetime.datetime = None,
         external_link: str = "",
         original_currency: str = "",
         owers: list = [],
--- a/ihatemoney/models.py
+++ b/ihatemoney/models.py
@@ -1,7 +1,9 @@
 from collections import defaultdict
 import datetime
+import itertools
 
 from dateutil.parser import parse
+from dateutil.relativedelta import relativedelta
 from debts import settle
 from flask import current_app, g
 from flask_sqlalchemy import BaseQuery, SQLAlchemy
@@ -263,6 +265,41 @@ class Project(db.Model):
             .order_by(Bill.date.desc())
             .order_by(Bill.id.desc())
         )
+
+    def get_newest_bill(self):
+        """Returns the most recent bill (according to bill date) or None if there are no bills"""
+        # Note that the ORM performs an optimized query with LIMIT
+        return self.get_bills_unordered().order_by(Bill.date.desc()).first()
+
+    def get_oldest_bill(self):
+        """Returns the least recent bill (according to bill date) or None if there are no bills"""
+        # Note that the ORM performs an optimized query with LIMIT
+        return self.get_bills_unordered().order_by(Bill.date.asc()).first()
+
+    def active_months_range(self):
+        """Returns a list of dates, representing the range of consecutive months
+        for which the project was active (i.e. has bills).
+
+        Note that the list might contain months during which there was no
+        bills.  We only guarantee that there were bills during the first
+        and last month in the list.
+        """
+        oldest_bill = self.get_oldest_bill()
+        newest_bill = self.get_newest_bill()
+        if oldest_bill is None or newest_bill is None:
+            return []
+        oldest_date = oldest_bill.date
+        newest_date = newest_bill.date
+        newest_month = datetime.date(
+            year=newest_date.year, month=newest_date.month, day=1
+        )
+        # Infinite iterator towards the past
+        all_months = (newest_month - relativedelta(months=i) for i in itertools.count())
+        # Stop when reaching one month before the first date
+        months = itertools.takewhile(
+            lambda x: x > oldest_date - relativedelta(months=1), all_months
+        )
+        return list(months)
 
     def get_pretty_bills(self, export_format="json"):
         """Return a list of project's bills with pretty formatting"""
--- a/ihatemoney/tests/budget_test.py
+++ b/ihatemoney/tests/budget_test.py
@@ -1,4 +1,5 @@
 from collections import defaultdict
+import datetime
 import re
 from time import sleep
 import unittest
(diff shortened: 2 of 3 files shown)


zorun:
Hmm, yes, it could be useful

ihatemoney/models.py
```diff
@@ -257,6 +257,14 @@ def get_member_bills(self, member_id):
             .order_by(Bill.id.desc())
         )
 
+    def get_most_recent_bill(self):
+        """Returns the most recent bill (according to bill date) or None if there are no bills"""
+        return self.get_bills_unordered().order_by(Bill.date.desc()).first()
+
+    def get_least_recent_bill(self):
```
almet, line 264: Shouldn't we name this `get_oldest_bill`?

ihatemoney/web.py
```diff
@@ -895,12 +896,23 @@ def strip_ip_addresses():
 @main.route("/<project_id>/statistics")
 def statistics():
     """Compute what each member has paid and spent and display it"""
-    today = datetime.now()
+    if g.project.has_bills():
+        first_date = g.project.get_least_recent_bill().date
+        last_date = g.project.get_most_recent_bill().date
+    else:
+        first_date = datetime.now() - relativedelta(months=12)
+        last_date = datetime.now()
```
almet, line 904: I don't quite understand why we need to have last and first date when we don't have bills yet?

ihatemoney/web.py
```diff
@@ -895,12 +896,23 @@ def strip_ip_addresses():
 @main.route("/<project_id>/statistics")
 def statistics():
     """Compute what each member has paid and spent and display it"""
-    today = datetime.now()
+    if g.project.has_bills():
+        first_date = g.project.get_least_recent_bill().date
+        last_date = g.project.get_most_recent_bill().date
```
almet, line 901: first and last doesn't really work for me when dealing with dates. I would prefer if we used `oldest_date` and `newest_date` I guess.

ihatemoney/web.py
```diff
@@ -895,12 +896,23 @@ def strip_ip_addresses():
...
+    else:
+        first_date = datetime.now() - relativedelta(months=12)
+        last_date = datetime.now()
+    # Infinite iterator towards the past
+    all_months = (last_date - relativedelta(months=i) for i in itertools.count())
+    # Stop when reaching one month before the first date
+    months = itertools.takewhile(
```
almet, line 908: Should we name this `active_months`?

almet:
Hey @zorun, here's a gentle ping on this so we can merge it before it rots (if you have the time, of course!).

zorun:
I don't have time to add tests for this anytime soon. You decide if it can be merged without tests or not :)

ihatemoney/web.py
zorun: This is just to mirror the current behaviour (last one year), but yes, we can simply display nothing in this case.

almet:
Rebased.

ihatemoney/web.py
zorun: They are not all active (I explained this in the commit message).

ihatemoney/models.py
zorun: You're right, I like compliated names apparently :)

ihatemoney/web.py
zorun: Changed to display nothing instead.

zorun:
@almet I took into account your remarks, fixed the corner cases, and added a test. A second quick review would be appreciated, thanks!

ihatemoney/models.py
```diff
@@ -262,6 +262,14 @@ def get_member_bills(self, member_id):
             .order_by(Bill.id.desc())
         )
 
+    def get_newest_bill(self):
+        """Returns the most recent bill (according to bill date) or None if there are no bills"""
```
almet, line 266: I find it weird to have two return types (None and a list), but that's probably not a problem ? :thinking:

almet: If we always return a list, we might want to get rid of the comments.

ihatemoney/tests/budget_test.py
```diff
@@ -917,6 +917,11 @@ def test_statistics(self):
         # Add a participant with a balance at 0 :
         self.client.post("/raclette/members/add", data={"name": "pépé"})
 
+        # Check that the "monthly expenses" table is empty
+        response = self.client.get("/raclette/statistics")
+        regex = r"<thead>\s*<tr>\s*<th>Period</th>\s*<th>Spent</th>\s*</tr>\s*</thead>\s*<tbody>\s*</tbody>"
```
almet, line 922: I'm probably rusted for not reading regexp for some time, but are we sure that a non-empty table will not pass the regex?

ihatemoney/tests/budget_test.py
```diff
@@ -982,6 +987,85 @@ def test_statistics(self):
...
+            ("December 2011", r"\$10.00"),
+            ("November 2011", r"\$0.00"),
+            ("October 2011", r"\$0.00"),
+            ("September 2011", r"\$0.00"),
+            ("August 2011", r"\$50.00"),
+        ]
+        months_fmt = [
+            r"<tr>\s*<td>{}</td>\s*<td>{}</td>\s*</tr>".format(month, amount)
+            for month, amount in months
+        ]
```
almet, line 1025: Actually, I find it kinda weird to check the output of the HTML like this, because it means that if we change the format of the table we also need to change the tests, even though we're not changing the actual behavior of the code.

I wonder if we might instead create a model function, and test that it works properly with python-related tests (more unit tests than functional tests, though). We might also want to have this information on the `/api` endpoint.

What do you think?

almet (commented):
Hey, here's a quick review. Overall, I think we're good and I'm okay merging it as-is. That being said, if you want to polish things a bit more I've left a few comments. Don't hesitate to ignore them if you feel I'm too picky.

ihatemoney/models.py
zorun: It returns a single bill, not a list. Is there a way to clarify this? I mean, the function is called get_something_bill, without S :)

ihatemoney/tests/budget_test.py
zorun: Definitely, yes, the \s* only matches whitespace and newlines. So it will only match if there is an empty body.

ihatemoney/models.py
almet: Oh my bad, it was very clear but I… I don't know I was somewhere else I guess :sleeping:

ihatemoney/models.py
```diff
@@ -262,6 +262,14 @@ def get_member_bills(self, member_id):
...
+        return self.get_bills_unordered().order_by(Bill.date.desc()).first()
```
Glandos, line 267: I am not familiar enough with SQLAlchemy. Could you check that it doesn't fetch a full batch of records from the database? Otherwise, we could manually add a `LIMIT 1` to the query for obvious performance reasons.

zorun: Good idea, I checked and sqlalchemy correctly adds a `LIMIT` itself. I added a comment to make that clear.

zorun:
I have changed the approach a bit:

- move the logic (that computes active months) as a method in the model, because it makes it easier to test
- change the tests to use that new method instead of checking the HTML output
- I am still checking the HTML table once (the empty one) to make sure it's displayed

@almet I guess another review is in order :)

ihatemoney/tests/budget_test.py
zorun: As explained below, I just implement this new testing approach!

For the API, I'm not sure. The client could just request the list of bills and compute any statistics itself.

almet:
It's far better this way, thanks!

PR #893, opened 2021-10-26
Update release documentation

zorun:


almet:
If we change to markdown I suggest that we convert all the files and do not keep rst and md.

PR #899, opened 2021-10-28
Reduce the size of showcase images

### LINKED ISSUE #875: Reduce the size of showcase images
Spinoff of #870 by @zorun.

Glandos:
Fixes #875

Original showcase images are kept in another directory, and compressed
in webp in the distributed directory.
Since they won't change so much, let's keep them in git.


almet:
Thanks !

PR #900, opened 2021-10-28
Consistently use questions in bill form

JocelynDelalande:
I suggest to be consistent with the question-style of our bill form.

Extra-bonus, this will be consistent with the expense table column names.

Isn't that beautiful ?


almet:
Works for me, thanks

PR #901, opened 2021-10-28
Avoid showcase-guy to point his finger at nothing

JocelynDelalande:
That occured on some screen sizes.

Example : 
image


almet:
Little things that make the difference. 😉

PR #906, opened 2021-11-01
Convert all the rst files to markdown.

### LINKED ISSUE #902: Convert all documentation to markdown
We currently use the python rst standard.

almet:
Update sphinx to work with the new format.

https://myst-parser.readthedocs.io/en/latest/sphinx/use.html#automatically-create-targets-for-section-headers 
was a valuable resource along the way.

Fixes #902.


almet:
If we're happy with using MD files, then I suggest we don't let this rot, it took some manual validation to have everything integrated together and not doing it again would be great :-)

Glandos:
Well, I thought it were easier. I'm glad it's done.

PR #916, opened 2021-11-12
Support WTForms 3

Glandos:
This bring the latest version, **but** some breaking changes are introduced.
Especially, StringField now default to `None` instead of `''`, so I changed the tests, without knowing the real impact.

@zorun do you have an idea of a real use case where it can break? All columns are nullable, as far as I know, but I want to be extra sure.


Glandos:
Ah, of course, this breaking change cannot be backported, at least in tests… I'll see it later, then.

setup.cfg
```diff
@@ -35,7 +35,7 @@ install_requires =
     Flask-SQLAlchemy>=2.4,<3
     Flask-Talisman>=0.8,<1
     Flask-WTF>=0.14.3,<2
-    WTForms>=2.3.1,<2.4
+    WTForms>=2.3.1,<3.1
```
almet, line 38: Shouldn't this be `>=3.0` ?

ihatemoney/forms.py
```diff
@@ -312,6 +326,7 @@ class BillForm(FlaskForm):
     original_currency = SelectField(_("Currency"), validators=[DataRequired()])
     external_link = URLField(
         _("External link"),
+        default="",
```
almet, line 329: So it means that by default it stores an empty string?

ihatemoney/forms.py
```diff
@@ -9,9 +9,23 @@
...
+    Label,
+    PasswordField,
+    SelectField,
+    SelectMultipleField,
+    StringField,
+    SubmitField,
+)
+
+try:
+    # Compat for WTForms <= 2.3.3
```
almet, line 25: Why should we keep backwards compatibility?

ihatemoney/forms.py
almet: Disregard, I understand now :-)

almet (approved):

ihatemoney/forms.py
Glandos: That's exactly my question. My experience showed me that jumping early on the "new backward incompatible release" train can be really annoying. So I tend to prefer solutions that introduce a backward compatibility layer, but I think this is not possible here, due to the default change for string values.

almet: > My experience showed me that jumping early on the "new backward incompatible release" train can be really annoying.

Ah, that seems bad indeed. Could you explain why?

symphorien:
can this be merged ? I kind of need that for the downstream nixos package.

almet:
Okay, let's do that for now, and drop it later on :-)

symphorien:
thanks !

PR #919, opened 2021-11-14
Production ready docker

YoueGraillot:
Some changes in this PR to make the docker image production ready :

- /healthcheck endpoint usefull for monitoring, ci test also uses this
- customizable PORT with environment variable
- customizable PUID/PGID, reduce attack surface and allow better integration in rootless environments
- size optimization
- update to python 3.10
- add postegresql compatibility

PUID/PGID default as root to not break current user environments


Dockerfile
```diff
@@ -1,5 +1,8 @@
-FROM python:3.7-alpine
+FROM python:3.9-alpine
```
Glandos, line 1: What about using 3.10-alpine? Is it because we don't officially support it yet?

docker-compose.test.yml
```diff
@@ -4,10 +4,8 @@ version: "3.9"
 services:
   ihatemoney:
     build: .
-    ports:
-      - "8000:8000"
   sut:
     image: alpine
-    command: wget --spider ihatemoney:8000
+    command: sh -c 'wget -qO- ihatemoney:8000/healthcheck | grep "OK"'
```
Glandos, line 9: Why this move is needed?

This add a new route for Docker only, and I feel that just fetching the home page with checking that the return code is 200 is enough.

Glandos (commented):
Some small remarks :)

Dockerfile
YoueGraillot: Exactly, didn't test with 3.10, but once tests are updated we should update this

docker-compose.test.yml
YoueGraillot: As I said in the PR, /healthcheck is also usefull for monitoring, kubernetes environments ect

Not fetching the whole page is the point

Dockerfile
Glandos: You can go ahead with 3.10 now that #921 was merged

YoueGraillot: Alright !

Dockerfile
```diff
@@ -21,13 +28,22 @@ ENV DEBUG="False" \
...
 ADD . /src
 
-RUN pip install --no-cache-dir -e /src
+RUN echo "**** install runtime packages ****" && \
+    apk add --no-cache \
+    shadow && \
+    echo "**** create runtime folder ****" && \
+    mkdir -p /etc/ihatemoney &&\
+    echo "**** install pip packages ****" && \
+    pip install --no-cache-dir gunicorn pymysql && \
```
Nikos410, line 39: How about installing `psycopg2` too, so you can use a postgres database with the docker image?

Nikos410: I tried adding it afterwards by running from a simple local Dockerfile:
```Dockerfile
FROM ihatemoney/ihatemoney:5.1.1
RUN pip install --no-cache-dir psycopg2
```

but building fails, not sure why, I don't really have experience using python.

almet: Can you tell us how it fails, maybe you have some logs?

Nikos410: Sure, I've opened a proper Issue: #924

almet:
This seems good to me, thanks ! but as I don't understand a word about docker, I'll let @Glandos or @zorun review and merge this !

zorun:
Well done for the lightweight postgresql integration :)

I have one nitpick: tests are run with constrained versions of psycopg2 and pymysql defined in setup.py, currently:
```
psycopg2-binary>=2.9,<3
PyMySQL>=0.9,<1.1
```

It would be good to use the same constraints in the Docker image to avoid surprises.

I suggest moving these two dependencies out of the `[dev]` group and into a new `[database]` group, update the tox config to install this group instead of `[dev]`, and also install this group in your Dockerfile (I think `pip install -e /src[database]` should work)

YoueGraillot:
> Well done for the lightweight postgresql integration :)
> 
> I have one nitpick: tests are run with constrained versions of psycopg2 and pymysql defined in setup.py, currently:
> 
> ```
> psycopg2-binary>=2.9,<3
> PyMySQL>=0.9,<1.1
> ```
> 
> It would be good to use the same constraints in the Docker image to avoid surprises.
> 
> I suggest moving these two dependencies out of the `[dev]` group and into a new `[database]` group, update the tox config to install this group instead of `[dev]`, and also install this group in your Dockerfile (I think `pip install -e /src[database]` should work)

Good call, done !

zorun:
Looks good, thanks!

YoueGraillot:
I updated the dockerhub readme to include tags description, compose example and env list : https://hub.docker.com/r/ihatemoney/ihatemoney

YoueGraillot:
Also, we may want to trigger another deploy to make PORT,PGID and PUID options available to `latest`.

zorun:
You mean overwriting the last tagged release on the dockerhub? That does not look like a good idea. I guess we simply need to do a release soon.

YoueGraillot:
That is what I meant ;)

Creating a new release after 5.1.1, thus pushing a new `latest` tag on docker

PR #921, opened 2021-11-15
add tests for python 3.10

Glandos:


Glandos:
For Python 3.10, pytest needs https://github.com/pytest-dev/pytest/pull/8540 that was backported in https://github.com/pytest-dev/pytest/pull/8629 And anyway, official support for Python 3.10 is in https://github.com/pytest-dev/pytest/pull/8494 that was released in [pytest 6.2.5](https://docs.pytest.org/en/stable/changelog.html#pytest-6-2-5-2021-08-29)

We are currently on `>=5.4.1`. I'm not against a switch to 6.2.5. But I need opinions on this. pytest 6.2.5 is already installed on "normal" tests, so everything runs fine on this.

zorun:
We currently claim to support Python 3.6 to 3.9, so if the new version of pytest works on Python 3.6, go for it!

Glandos:
[**@zorun**](https://github.com/zorun) commented on [15 nov. 2021, 21:54 UTC+1](https://github.com/spiral-project/ihatemoney/pull/921#issuecomment-969311484 "2021-11-15T20:54:16Z - Replied by Github Reply Comments"):
> We currently claim to support Python 3.6 to 3.9, so if the new version of pytest works on Python 3.6, go for it!

Yes, Python 3.6 was the minimal requirement updated in [pytest 6.2.0](https://docs.pytest.org/en/stable/changelog.html#pytest-6-2-0-2020-12-12)

PR #923, opened 2021-11-21
Add tagging functionality

Nikos410:
Hey there!

This PR contains only the tagging functionality from #557.
I also rebased the changes against the current master.


almet:
Thanks for proposing this :-). For it to be merged, we're lacking tests showing that it works as expected, though. Are you willing to write some?

almet:
@glandos, @zorun what do you think about adding this feature? I'm not sure we really need it, so you input would be helpful :-)

Nikos410:
> Thanks for proposing this :-). For it to be merged, we're lacking tests showing that it works as expected, though. Are you willing to write some?

Of course, I'll try to find the time this week :)

> what do you think about adding this feature? I'm not sure we really need it, so you input would be helpful :-)

I think it can be really helpful if we add a "browse-by-tag" feature, so you can easily see how much you paid for food vs. how much you paid for "fun" things. But of course if you think that is not what this project is aiming to do I understand :)

Nikos410:
Hmm, I just realized that this PR in it's current state does not really make a lot of sense. Since I removed the filter functionality, you can only create tags, but have nowhere to see them.

I would volunteer to implement something like the "browse-by-tag" feature I mentioned above if you would like to have a functionality like that.

Glandos:
Tagging is somewhat requested by some users. I guess that we can integrate it without making the UI more complex. This is a task that requires more work in the UI than in the backend itself.

zorun:
Agreed with Glandos, the most difficult part here is the UI/UX.

almet:
Hi, I'm closing this as it's been stalled for the last 3 months. Feel free to reopen if you want to continue your work on this.

PR #926, opened 2021-11-21
Check that project creation via the API doesn't required a captcha.

### LINKED ISSUE #922: Do not require captcha when creating a project via the API

almet:
Fixes #922


almet:
Not sure if this really solves the issue, but merging still, as it adds a lacking test.

Issue #903, opened 2021-10-30
Notifications are getting in the way


almet:
See ![image](https://user-images.githubusercontent.com/67073/139556604-865fea3b-923a-4735-8d34-501272e82c1b.png).

I'm not sure what to do about this: we could probably make the notifiications in the flow, and not floating on top of the rest of the content.

Glandos:
I explored some solutions, but none of them seems better than the others:
- It's not *simple* to make it in the flow, because the main content is sometime customized, and I would much prefer that flash messages are defined only in `layout.html`
- I tried in the top navbar (not absolute), but it can be difficult not to have it in the way either:
![image](https://user-images.githubusercontent.com/297578/139602186-78b3a1ae-a182-4dbe-94b1-6a68e4962b65.png)
- It can be put absolute **on top** of the navbar links. Navigation is usually less important than flash messages

almet:
Thanks! Do you think it's possible to have it in the layout.html but after the header with the links and logo?

Glandos:
The current screenshot was made with the `messages` div right after the `h1` element in the navbar.

Do you mean to have it on the right in the top navbar, after the project menu? Or at the bottom of this?

almet:
I'm thinking about putting it after the top navbar and the project menu. At the bottom of it, yes.

JocelynDelalande:
A design warning : if flash messages are to be put back in flow : be sure that its vanishing after a couple of seconds does not make the whole page content move up. Some UI act like that, and that is very annoying.

(a solution to that is to remove notification from auto-hiding)

Glandos:
Since #856 there is no auto-hide anymore.

But that's the main problem I encountered: how to integrate flash message without moving existing content. Even when I click on some "Close" action, I don't want the whole content to move. On top of that, moving the content can bring some other design issues.

Glandos:
After seeing the issue in many angle, here is my best proposal so far:

https://user-images.githubusercontent.com/297578/141375139-c30029ca-90bd-4ec8-83a4-120efd2c173d.mp4

I didn't modify the absolute positioning. The trick was to add an empty div of 4.5rem height (as before) on top of messages, but only for screen larger than `md` in bootstrap.

almet:
Sorry I didn't came back to you sooner @Glandos. It seems good enough to me :+1:

Nikos410:
[...]

PR #929, opened 2021-11-21
Allow clicking buttons behind messages div

Nikos410:
If an alert is shown, any elements at the same height cannot be clicked, which is pretty frustrating, because it hides the "New Bill" button. This happens because the "messages" div is positioned absolutely with a width of 100%. 

This PR solves this issue by disabling pointer-events for that div and enabling them again specifically for alerts in that div.

### DIFF
--- a/ihatemoney/static/css/main.css
+++ b/ihatemoney/static/css/main.css
@@ -470,6 +470,11 @@ tr.payer_line .balance-name {
   position: absolute;
   top: 4.5rem;
   width: 100%;
+  pointer-events: none;
+}
+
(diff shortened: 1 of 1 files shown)


almet:
This is related to the discussion happening in #903, you probably should cooperate with @Glandos on this.

Nikos410:
Of course! Thanks for the heads-up, I did check existing issues but apparently missed that one :)

almet:
In any case, I believe this is currently blocking, and merging it solves an existing issue :-) Thanks!

PR #931, opened 2021-11-22
Do not require a captcha when using the API

almet:
This was trickier than expected, due to some side effects : when the
captcha is set to `True` via configuration, it doesn't change the
behavior directly of the ProjectForm class, but does so only when the
project form is used in the `web.py` module.

So, when just using the API (and not using the web.py module, for
instance during tests — manual or functional), no problem was shown,
and everything was working properly.

But at soon as somebody sees the "/" endpoint, the captcha was
required, by both the API and the `web.py` module.

This fixes it by adding a way to bypass the captcha with a new
`bypass_captcha` property on the form.

### DIFF
--- a/ihatemoney/api/common.py
+++ b/ihatemoney/api/common.py
@@ -50,7 +50,7 @@ def need_auth(f):
 
 class ProjectsHandler(Resource):
     def post(self):
-        form = ProjectForm(meta={"csrf": False})
+        form = ProjectForm(meta={"csrf": False}, bypass_captcha=True)
         if form.validate() and current_app.config.get("ALLOW_PUBLIC_PROJECT_CREATION"):
             project = form.save()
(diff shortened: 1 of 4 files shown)


zorun:
Wow, this is a good heisenbug! Maybe we should get rid of that `enable_captcha` class method to avoid this kind of very subtle issues in the future? Isn't there another way to do this, maybe with form inheritance?

almet:
Hmm, that's correct. I believe we should merge this and change the way we handle this captcha thing in another pull request. I opened a new bug for this.

almet:
Actually, I changed the way things were implemented for the Captcha. It's a lot more clear this way, thanks @zorun for the review :+1:

zorun:
Great, thanks for the change, it looks good :) I think you just need to reformat

PR #943, opened 2021-11-25
Rework flash messages

### LINKED ISSUE #903: Notifications are getting in the way

Glandos:
Fixes #903

They are now a macro, and can be called within other templates.
This give the ability to fit them anywhere, at the cost of template complexity.

### DIFF
--- a/ihatemoney/static/css/main.css
+++ b/ihatemoney/static/css/main.css
@@ -464,17 +464,8 @@ tr.payer_line .balance-name {
   margin-top: 10px;
 }
 
-.messages {
-  display: flex;
-  justify-content: center;
-  position: absolute;
-  top: 4.5rem;
-  width: 100%;
-  pointer-events: none;
-}
-
-.messages .alert {
-  pointer-events: auto;
+.flash {
+  z-index: 1031;
 }
 
 .light {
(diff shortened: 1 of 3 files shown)


Glandos:
https://user-images.githubusercontent.com/297578/143436041-9f76856d-e5a0-4336-b792-997988fc3b76.mp4


https://user-images.githubusercontent.com/297578/143436045-c364b06b-bd17-483a-9b1b-161ff65c85bf.mp4


https://user-images.githubusercontent.com/297578/143436049-c2d75961-2b13-4434-b57f-98cc2da1b4b4.mp4

This is different from https://github.com/spiral-project/ihatemoney/issues/903#issuecomment-966657184 since notifications can be after the navbar and the sidebar in mobile layout.

almet:
Hey, since this doesn't change anything to how they appear (it seems, at least?) I wonder if you have a real use case for this change?

Glandos:
In this PR, most of messages are now display **after** the sidebar, so on mobile, they are on top of bill list, and not on top of the participant list.

In https://github.com/spiral-project/ihatemoney/issues/903#issuecomment-966657184 the example works fine, because this is a small messages, and there is only one of it. If there are at least 2 messages with a long text, it will cover the screen anyway, and it seems to me to be a better place to integrate it in the `sidebar_table_layout.html`.

almet:
Okay, your call: it seems you put more thoughts than me on this one ;-)

Issue #909, opened 2021-11-06
Study features and UX of tricount VS ihatemoney


almet:
Most of my friends are using the paid, closed-source and ad-powered Tricount to handle their shared expenses, and I wonder why.

It's partly because it's a "well known" solution, but I do think there are other reasons, and I would like to list them in order to craft a good alternative to it. I don't think we should copy tricount, but I believe studying it could help us improve in some ways.

I haven't used Tricount yet, so I'm telling this from reading their documentation. We might want to install it and check by ourselves as well. 

Here are some stuff that tricount does differently : 

- Accounts : might be useful to browse existing projects. IHM doesn't have accounts because we believe it's simpler, but that might not be the case (it means you have to remember a code for each project, which might be counter-productive) 
- Easy to find mobile app : at least it has the same commercial name. I believe it's an issue for our project to have apps with different names. I understand where it's coming from, but at some point I wonder if we should unite, create a kind of brand or something (with cospend, that is)
- Notifications : when somebody adds an expense on Tricount, you get a notification (I believe).
- Balances with graphs : You can see quickly the balance in a graphic way. It's a quick win

I also think we could do a user survey / UX survey, with people using the app in real time and us watching how they do stuff and taking notes. That would probably help us have a better ecosystem.

I wonder if you have some thoughts about this? Cheers!

Glandos:
## Accounts
I don't like them. They introduce another line in my password manager.
But people are so used to surveillance based applications that they take it for granted: creating an account is the norm.
Anyway, I have to agree that a code for each project is more complicated. So here my point of view:
- Keep project code, without account
- Add optional account
- Within an account, you can add a project, typing the code only once (or the invitation link), and then you never enter it anymore. We'll have to find out how to make this work when the code is changed.

## Mobile application
I have to say that I don't have a smartphone. But the goal fulfilled by all those applications really fit the mobile usage: you enter them right after the expense, and forget. We can discuss with @eneiluj but I guess that the only acceptable solution will be to add a suffix, like a label. E.g. "MoneyBuster (IHM)" "Cospend (IHM)". As I write this, I'm not even convinced. Idea needed. Maybe an icon/color code common to every project can be more useful.

## Notifications
If I understand correctly, to have notifications, you need Firebase, which is Google. Fast search leads to https://stackoverflow.com/questions/41511475/android-push-notification-without-firebase which is clearly non trivial…

## Graphs
[...]

PR #946, opened 2021-11-25
Hide project code from user

### LINKED ISSUE #525: Simplify the creation of a new project
I like how Splittypie does this : https://splittypie.com/new.

It asks the user to select a name for the project, and in the same window, proposes to add the new participants.

Glandos:
This is done to fix #525 and to implement https://github.com/spiral-project/ihatemoney/issues/909#issuecomment-967686397

- [X] Remove password from forms
- [X] Replace it with id in destructive actions
- [ ] Fix tests
- [ ] Use token in API
- [x] Show token/link in admin dashboard
- [ ] Reset password changed with "resend invite link"
- [ ] Update API docs to use only token
- [ ] Discuss token invalidation

### DIFF
--- a/ihatemoney/api/common.py
+++ b/ihatemoney/api/common.py
@@ -65,7 +65,8 @@ class ProjectsHandler(Resource):
             db.session.add(project)
             db.session.commit()
             send_creation_email(project)
-            return project.id, 201
+            token = project.generate_token()
+            return {"token": token, "id": project.id}, 201
         return form.errors, 400
 
 
--- a/ihatemoney/forms.py
+++ b/ihatemoney/forms.py
@@ -2,6 +2,7 @@ from datetime import datetime
 import decimal
 from re import match
 from types import SimpleNamespace
+import secrets
 
 import email_validator
 from flask import request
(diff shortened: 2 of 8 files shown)


Glandos:
I am glad with first steps. This is a more a proof-of-concept for now, but nothing seems a blocker for now.
The most important part will be to use the token instead of the password within the API.

One very important thing is that email server should be nearly mandatory now, to be sure that creator get the invitation link on creation. Otherwise, (s)he has to go in the invitation panel to have it to avoid getting locked out. I plan to make it more obvious after project creation anyway.

Glandos:
OK, so now, I've adapted most of the tests to avoid using `/authenticate` and two tests are left:
- `test_invite_code_invalidation`: Should we delete this test? Since project code won't change, at least not like that anymore
- `test_authentication`: This is testing the `/authenticate` endpoint that… seems to be a good candidate for deletion

The main end point for authentication will be `/project/join/TOKEN` so `/authenticate` can die. Do you agree?

Regarding token invalidation, the main use case is to be able to revoke some members access, am I right? I have still to think about a solution for this.

almet:
Hey, seems you did quite a good amount of work in here. Do you need any help on this?

almet:
@Glandos are you still on this or should we close it? Cheers!

almet:
Hey, I'm doing some cleanup on the repo, so closing this for now, as it has no activity in the past years.

Cheers!

PR #161, opened 2017-01-02
Optimise sql queries

zorun:
These two commits mostly fix issue #159.

Issue #889, opened 2021-10-23
Computing the balance or statistics is slow on big projects


zorun:
Somebody reported that ihatemoney is now much slower than the previous version (4.1.5), and I can confirm using a big project.

I identified two slow actions:

- computing the balance (as shown in the list of bills or in the "settle bills" page)
- computing statistics

For a project with 1000 bills, both actions create thousands of SQL queries like this:

```
SELECT sum(person.weight) AS sum_1 
FROM person JOIN billowers ON person.id = billowers.person_id JOIN bill ON bill.id = billowers.bill_id 
WHERE bill.id = ?
```

I traced this query back to the `Bill.pay_each_default()` function, which is used to basically compute `bill_amount / total_owers_weight` (what is shown as "42 € each" in the list of bills). Each call to this function triggers the above SQL query.

Some ideas to improve things:

- add a new field `Bill.total_owers_weight` so that we can cache the result of this expensive computation. We would need to make sure this field is recomputed when necessary.
- rethink the data model, so that simple things don't need complicated JOIN operations.

PR #949, opened 2021-11-27
[Performance] Get weight sum along with bills to scale

### LINKED ISSUE #159: Optimise SQL queries
A project with a few hundreds bills loads quite slowly: rendering the `list_bills` view takes several seconds.

Debugging with the Flask toolbar, it seems that the number of SQL queries issued by `list_bills` is very high: number of bills * number of owers.  For instance, with 100 bills and an average of 5 owers for each bill, there will be 500 SQL queries.

See also #156 for some initial thoughts on this issue

Glandos:
Otherwise, we need to get the weight sum for each displayed bill.
Here, we are much more scalable.

With a pagination of 500, we previously had 507 queries if bill currencies were project's default, and nearly the double if not.
Now, we are down to 7.
On my Ryzen 4750G, Flask-debug toolbar profiler told me that displaying 500 out of 1000 bills drop from 2200ms to less than 1000ms.

More performance are expected in other PR.

Finally fixes #159 by continuing #161

### DIFF
--- a/ihatemoney/models.py
+++ b/ihatemoney/models.py
@@ -256,6 +256,12 @@ class Project(db.Model):
         )
 
     def get_bill_weights(self):
+        """
+        Return all bills for this project, along with the sum of weight for each bill.
+        Each line is a (float, Bill) tuple.
+
+        Result is unordered.
+        """
         return (
             db.session.query(func.sum(Person.weight), Bill)
             .options(orm.subqueryload(Bill.owers))
@@ -267,6 +273,7 @@ class Project(db.Model):
         )
 
     def get_bill_weights_ordered(self):
+        """Ordered version of get_bill_weights"""
         return self.ordered_bills(self.get_bill_weights())
 
     def get_member_bills(self, member_id):
--- a/ihatemoney/models.py
+++ b/ihatemoney/models.py
@@ -246,12 +246,28 @@ class Project(db.Model):
 
     def get_bills(self):
         """Return the list of bills related to this project"""
+        return self.ordered_bills(self.get_bills_unordered())
+
+    def ordered_bills(self, query):
         return (
-            self.get_bills_unordered()
-            .order_by(Bill.date.desc())
+            query.order_by(Bill.date.desc())
             .order_by(Bill.creation_date.desc())
             .order_by(Bill.id.desc())
         )
+
+    def get_bill_weights(self):
+        return (
+            db.session.query(func.sum(Person.weight), Bill)
+            .options(orm.subqueryload(Bill.owers))
+            .select_from(Person)
+            .join(billowers, Bill, Project)
+            .filter(Person.project_id == Project.id)
+            .filter(Project.id == self.id)
+            .group_by(Bill.id)
+        )
+
+    def get_bill_weights_ordered(self):
+        return self.ordered_bills(self.get_bill_weights())
 
     def get_member_bills(self, member_id):
         """Return the list of bills related to a specific member"""
--- a/ihatemoney/models.py
+++ b/ihatemoney/models.py
@@ -246,9 +246,10 @@ class Project(db.Model):
 
     def get_bills(self):
         """Return the list of bills related to this project"""
-        return self.ordered_bills(self.get_bills_unordered())
+        return self.order_bills(self.get_bills_unordered())
 
-    def ordered_bills(self, query):
+    @staticmethod
+    def order_bills(query):
         return (
             query.order_by(Bill.date.desc())
             .order_by(Bill.creation_date.desc())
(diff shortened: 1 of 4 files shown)


Glandos:
Please review and test. I feel like I did some SQLAlchemy magic without fully understanding what I did. But normally, all bills are here with much less queries.

Glandos:
I integrated it on my own instance (Intel Atom D2550), and I dropped from 6,2s to 1,5s page loading in a budget with 197 bills.

zorun:
Nice one :) This is an improvement I had left out of #889 / #890 because I was not sure how to go about it

ihatemoney/web.py
```diff
@@ -654,12 +653,7 @@ def list_bills():
     # set the last selected payer as default choice if exists
     if "last_selected_payer" in session:
         bill_form.payer.data = session["last_selected_payer"]
-    # Preload the "owers" relationship for all bills
-    bills = (
-        g.project.get_bills()
-        .options(orm.subqueryload(Bill.owers))
-        .paginate(per_page=100, error_out=True)
-    )
+    bills = g.project.get_bill_weights_ordered().paginate(per_page=100, error_out=True)
```
zorun, line 656: `bills` is actually a list of tuples here, right? Please rename the variable accordingly :)

ihatemoney/models.py
```diff
@@ -246,13 +246,29 @@ def get_bills_unordered(self):
...
+    def ordered_bills(self, query):
         return (
-            self.get_bills_unordered()
-            .order_by(Bill.date.desc())
+            query.order_by(Bill.date.desc())
             .order_by(Bill.creation_date.desc())
             .order_by(Bill.id.desc())
         )
 
+    def get_bill_weights(self):
```
zorun, line 258: Add a comment about what this function returns, it's not exactly clear (it returns a list of tuples?)

zorun (changes requested):
The principle looks good and the generated SQL query also :)

I have left minor nitpick comments.

Glandos:
Thanks for the review. Now I see that most of the time is template generation which is clearly less trivial.

ihatemoney/web.py
Glandos: I hope this is clearer. I didn't go for a very long name like `tuples_of_weights_and_bills`, but decided to add a comment.

However, the template still uses `bills`. I hoped to integrated the weight sum directly into the `Bill` object, it should much much clearer, but… I don't know (yet?) how to do it.

zorun: In Django, it's done with a function called `annotate`, I don't know if something like this exists in SQLAlchemy

zorun: After a quick search, `column_property` looks like it could be useful, but I don't understand the documentation... https://docs.sqlalchemy.org/en/14/orm/mapped_sql_expr.html

Simpler way adapted from https://stackoverflow.com/a/14064418 :

```python
def annotate(query):
    for bill, weight in query:
        bill.weight = weight
        yield bill
```

If it works it could even be done directly in `get_bill_weights`.

ihatemoney/web.py
```diff
@@ -654,16 +653,15 @@ def list_bills():
...
     if "last_selected_payer" in session:
         bill_form.payer.data = session["last_selected_payer"]
-    # Preload the "owers" relationship for all bills
-    bills = (
-        g.project.get_bills()
-        .options(orm.subqueryload(Bill.owers))
-        .paginate(per_page=100, error_out=True)
+
+    # Each item will be (weight_sum, Bill) tuple
+    weights_bills = g.project.get_bill_weights_ordered().paginate(
```
almet, line 658: should we change the naming for this to be `bills_with_weights`, or `weighted_bills`? `weight_bills` isn't clear enough for me.

ihatemoney/web.py
zorun: Hi @Glandos, I hope I didn't discourage you with this annotation thing :)

In my opinion it's optional: it would clarify the code to have it, but the most important part of your changes is the performance improvement patch.

Glandos: OK, I wrote this in the comment. Let's not focus on being perfect at first shot.

ihatemoney/models.py
```diff
@@ -246,13 +246,36 @@ def get_bills_unordered(self):
 
     def get_bills(self):
         """Return the list of bills related to this project"""
+        return self.ordered_bills(self.get_bills_unordered())
+
+    def ordered_bills(self, query):
```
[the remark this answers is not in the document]
Glandos: Done. I took the liberty to make it static, because it doesn't use anything from the instance. And make it usable from outside… if ever needed.

ihatemoney/web.py
Glandos: I went with `weighted_bills`.

zorun:
Hi @Glandos , is there anything blocking this PR? I think it's good to go, no need to fix all performance issues (e.g. templates are another story entirely)

zorun:
For some reason Github still says that I requested some changes, but I can't find anything related to that.

PR #950, opened 2021-11-28
Legal link fixes

YoueGraillot:
Two minor fixes :
- 68a4aa05e8f63ab1501a32ae3b39ad212c7d5c13 set `LEGAL_LINK` default value to "False", icon was displayed and redirected to /False if not overwritten
- displaying legal link in footer caused an overflow

Before :
image
After :
image


almet:
Thanks!

Issue #152, opened 2016-09-22
Recurring bills (ISP, electricity, water, charges...)


Vincent14:
I would know if it's possible to add bills that could have a frequency (every day/week/month//year)?

Djabal:
That will be very convenient !

aavenel:
This could be useful ! However, it's not really clear how do we implement this. Mockups are welcome.

Although it's quite difficult to configure, it's possible to do this now using the API.

```
curl --basic -u demo:demo https://ihatemoney.org/api/projects/demo/members
[{"activated": true, "id": 31, "name": "Arnaud"},
 {"activated": true, "id": 32, "name": "Alexis"},
 {"activated": true, "id": 33, "name": "Olivier"},
 {"activated": true, "id": 34, "name": "Fred"}]
```
Note your member id, as it it required to configure the next request.

```
curl --basic -u demo:demo -X POST\
https://ihatemoney.org/api/projects/demo/bills\
-d "what=raclette&payer=31&payed_for=31&payed_for=32&amount=200"
```

If you omit the date field, we will use the current time at bill's creation.

You can find the documentation of the API here : https://github.com/spiral-project/ihatemoney/blob/master/docs/api.rst

To automate this, you can use a Cron script, or even use some service like IFTTT. 
I have set up an applet on IFTTT (Maker channel) to make the REST request.

This is pretty mush the same thing as using curl, except that you need to pass the authentification data like this :
`https://projectname:password@ihatemoney.org/api/projects/projectname`

almet:
We've not implemented this in the current version of the software in order to keep things easy to grasp (and the UX simple), but I can see why this would be useful. 

As @aavenel says, mockups (or implementations!) would be welcome for this. For now, I don't think anyone volunteered.

almet:
The way to do this in Flask would be to create a task (in `manage.py`)  that would check all projects and see the recurring bills, each day, and mark them with the last date is was run.

I believe it's simple enough for a good first issue, so labeled as such.

Andrew-Dickinson:
Hi @almet 

I'm thinking of working on this issue, it looks like the first step would be mockup of the UI changes in the "Add a bill" dialog. It seems like you guys put a lot of emphasis on keeping the default workflow as simple as possible. For this feature, maybe the best way to implement that would be a "Make recurring" hyperlink just above the submit button which expands to reveal the interface for specifying the frequency, first occurrence, etc. 

[...]

PR #951, opened 2021-11-28
CSV bills import (cospend compatible)

youegraillot:
Hi, bigger pr this time to implement CSV project import
- CSV compatible import, same form input as JSON, we sort that out server-side
- Cospend compatible import
- Fix errors logs/flash in edit page
- Updated localization (best effort)
- Optimized Person fetching (get_by_ids and get_by_names)

This is a draft for now, I'm planning to update and add some more tests for this feature


Glandos (changes requested):
I am glad to see more Cospend compat code :+1: 

However, I have mixed feelings about the new `__init__` for `Bill`. In some case, it is helpful, but I feel that it is sometimes less readable than previous code with explicit affectations.

Great move though for replacing `fake_form` by an explicit `export`.

As a last comment, you can remove all translations from this, Weblate will handle this for us.

almet:
That's cool! Thanks for taking the time to implement this :-)

I wonder if this will break importing previously exported project data?

YoueGraillot:
> That's cool! Thanks for taking the time to implement this :-)
> 
> I wonder if this will break importing previously exported project data?

As far as I tested nope, same behavior as for json : only add missing transactions

almet:
Hey, let us know if you need some help on this, we might be able to find some time to take over if you want!

youegraillot:
Hi, I'm planning to update this in the following days when I get some time, but for this PR or any future one feel completely free to update it !

ihatemoney/models.py
```diff
@@ -435,16 +473,19 @@ def create_demo_project():
...
+                Bill(
+                    amount,
+                    None,
+                    None,
+                    "XXX",
+                    [members[name] for name in owers],
+                    members[payer].id,
+                    project.default_currency,
+                    what,
+                )
```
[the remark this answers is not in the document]
youegraillot: Modified init with named parameters, if it's still a problem I can revert it

ihatemoney/translations/bn_BD/LC_MESSAGES/messages.po
```diff
@@ -52,8 +52,11 @@ msgid ""
 "multiple currencies."
 msgstr ""
 
-msgid "Import previously exported JSON file"
-msgstr "পূর্বে রপ্তানি করা JSON ফাইল আমদানি করুন"
+msgid "Compatible with Cospend"
+msgstr "Cospend সঙ্গে সামঞ্জস্যপূর্ণ"
+
+msgid "Import previously exported project"
+msgstr "পূর্বে এক্সপোর্ট প্রকল্প আমদানি করুন"
```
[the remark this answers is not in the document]
youegraillot: Reverted all translation files, I think messages.pot is updated automagically ? I'm not sure

ihatemoney/forms.py
```diff
@@ -240,7 +242,7 @@ def validate_captcha(form, field):
             raise ValidationError(Markup(message))
 
 
-class DestructiveActionProjectForm(FlaskForm):
+class DeleteProjectForm(FlaskForm):
```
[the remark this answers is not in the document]
youegraillot: Didn't knew that, it is now reverted

youegraillot:
Now with CSV tests I believe this PR is ready for review :)

I'm not really satisfied with the CSV/JSON tests implementation but I could not find a better way without converting to pytest, btw I'm wondering why this project uses unittest instead ?

ihatemoney/models.py
```diff
@@ -320,6 +322,42 @@ def switch_currency(self, new_currency):
...
+        members_project = [str(m) for m in self.members]
+        members_new = [
+            m for m in get_members(bills) if str(m[0]) not in members_project
+        ]
+        for m in members_new:
+            Person(name=m[0], project=self, weight=m[1])
+        db.session.commit()
+
+        # Import bills not already in the project
+        bills_project = self.get_pretty_bills()
```
almet, line 337: I'm not sure I follow here: this is named `bills_project` but it seems to contain bills instead, right? If that's the case, we should probably rename the variable here.

almet: I think I understand. It seems to be the bills of the existing project. If this is the case, we could name it `project_bills` :-)

ihatemoney/models.py
```diff
@@ -320,6 +322,42 @@ def switch_currency(self, new_currency):
...
+            same = False
+            for b_p in bills_project:
+                if same_bill(b_p, b):
+                    same = True
+                    break
+            if not same:
+                # Create bills
+                db.session.add(
+                    Bill(
+                        amount=b["amount"],
```
almet, line 349: We probably should check that the data contains this key before using it, otherwise it may fail.

almet: Or we might want to try/catch this entire block and display an error message with the format of the required data.

ihatemoney/utils.py
```diff
@@ -150,6 +150,26 @@ def list_of_dicts2csv(dict_to_convert):
...
 
+def csv2list_of_dicts(csv_to_convert):
+    """Take a csv in-memory file and turns it into
+    a list of dictionnaries
+    """
+    csv_file = TextIOWrapper(csv_to_convert, encoding="utf-8")
+    reader = csv.DictReader(csv_file)
+    result = []
+    for r in reader:
+        # cospend filtering
```
almet, line 161: Could you please expand here on why this is required and what's it's doing? Thanks :-)

almet (commented):
Thanks for doing this and tidying up the tests during the process :-) I believe we're not using py.test because we are not familiar with it, but if you want and know how to improve these tests with py.test it will be useful and appreciated, don't hesitate to open a PR (or an issue) about it.

I've not reviewed all the part with the tests due to lack of time right now, but I've left a few comments here and there.

youegraillot:
pytest would indeed simplify code and avoid some ugly duplication, but I honestly don't have the motivation to refactor all tests. I might write new ones with it though if there is no reason not to do so

Thanks for your feedbacks I'll check these later this week

ihatemoney/forms.py
```diff
@@ -319,33 +321,31 @@ class BillForm(FlaskForm):
...
+        return Bill(
+            float(self.amount.data),
+            self.date.data,
+            self.external_link.data,
+            str(self.original_currency.data),
+            Person.query.get_by_ids(self.payed_for.data, project),
+            self.payer.data,
+            project.default_currency,
+            self.what.data,
+        )
```
Glandos, line 334: It could be good to also use named parameters here for consistency

Glandos:
[**@Youe Graillot**](https://github.com/youegraillot) commented on [13 déc. 2021, 04:34 UTC+1](https://github.com/spiral-project/ihatemoney/pull/951#issuecomment-992079481 "2021-12-13T03:34:27Z - Replied by Github Reply Comments"):
> Now with CSV tests I believe this PR is ready for review :)
> 
> I'm not really satisfied with the CSV/JSON tests implementation but I could not find a better way without converting to pytest, btw I'm wondering why this project uses unittest instead ?

But… this project uses pytest: https://github.com/spiral-project/ihatemoney/blob/master/setup.cfg#L57

And I'm running all tests locally with `.venv/bin/pytest --pyargs ihatemoney.tests.main_test -k weight`, it all run fine.

youegraillot:
> [**@Youe Graillot**](https://github.com/youegraillot) commented on [13 déc. 2021, 04:34 UTC+1](https://github.com/spiral-project/ihatemoney/pull/951#issuecomment-992079481):
> 
> > Now with CSV tests I believe this PR is ready for review :)
> > I'm not really satisfied with the CSV/JSON tests implementation but I could not find a better way without converting to pytest, btw I'm wondering why this project uses unittest instead ?
> 
> But… this project uses pytest: https://github.com/spiral-project/ihatemoney/blob/master/setup.cfg#L57
> 
> And I'm running all tests locally with `.venv/bin/pytest --pyargs ihatemoney.tests.main_test -k weight`, it all run fine.

Running tests with pytest does not mean they are written using pytest, which can run unittest but with caveats.
Fixtures, parametrization and hooks are not supported with this method.

ihatemoney/models.py
youegraillot: Ok !

ihatemoney/models.py
youegraillot: import_bills is already called from a try/catch block in web.import_project
but I agree we should add another one in import_bills as a safeguard in case we use the method elsewhere
done !

ihatemoney/utils.py
youegraillot: Yup
by the way I thought about also exporting data in cospend format (with lines for users and a table for categories)
it would simplify our csv import code AND render IHM exports compatible with cospend

ihatemoney/forms.py
youegraillot: Oops, forgot to include them here, fixed !

almet:
Is this ready for review?

almet:
I believe it's alright from my point of view. As soon as we get the green light from @youegraillot, I believe we should merge :-)

youegraillot:
Oups, I didn't notified you last week

I think I have answered all comments, light is green !

almet:
Thanks a bunch!

youegraillot:
My pleasure, I'll move on to #152 then

PR #956, opened 2021-12-02
Top pagination and new bill button are sticky

Glandos:
When there are a lot of bills, this is much easier to access.

Before:
https://user-images.githubusercontent.com/297578/144515015-34d05c27-7117-4357-878f-4199b2172193.mp4

After:
https://user-images.githubusercontent.com/297578/144515040-21b74ca7-a02b-4198-856a-10a4ccd9d166.mp4

Notice also that the pagination and the "New bill button" is aligned with the "Add participant" field.

On mobile, since there are a lot of page, the pagination wrapping is not so nice, but it could be improved by changing "Next/Previous" label by using icon only on mobile.

### DIFF
--- a/ihatemoney/static/css/main.css
+++ b/ihatemoney/static/css/main.css
@@ -283,14 +283,6 @@ footer .footer-left {
   }
 }
 
-#new-bill {
-  margin-top: 30px;
-}
-
-#pagination-top {
-  margin-top: 30px;
-}
-
 /* Avoid text color flickering when it loose focus as the modal appears */
 .btn-primary[data-toggle="modal"] {
   color: #fff;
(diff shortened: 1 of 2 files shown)


Glandos:
Fun fact: this was an improvement suggested by my wife.

zorun:
Very nice, I like it! :+1: 

Does this mean that we no longer need the pagination at the bottom? (real question, I don't know)

almet:
Wow, thanks I like it too :-) If we have the pagination on top, we might indeed want to remove it from the bottom of the page, from my perspective!

Glandos:
Et voilà!

almet:
Yeah, thanks !

Issue #847, opened 2021-10-10
Expose list of allowed currencies in API


zorun:
While updating the API doc for 5.0, I came accross an inconsistency: we can choose the currency while creating a project through the API, but there is no way to know which currencies are allowed.

It would be nice to expose the list of currencies in the API.

PR #961, opened 2021-12-04
List supported currencies in API under api/currencies

petermaksymo:
For #847 

Returns a list of supported currencies under `/api/currencies`.

This is my first PR for this project, any feedback is greatly appreciated :)

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -242,3 +242,23 @@ You can get some project stats with a `GET` on
             "balance": -10.5
         }
     ]
+
+### Currencies
+
+You can get a list of supported currencies with a `GET` on
+`/api/currencies`:
+
+    $ curl --basic https://ihatemoney.org/api/currencies
+    [
+        "XXX",
+        "AED",
+        "AFN",
+        .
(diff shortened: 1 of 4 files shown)


almet:
Hi, thanks for opening a pull request here. It's appreciated :-) I think adding a test for this would be helpful in order to test that it's behaving the way we think it is.

Let me know if you need help on how to do this properly.

petermaksymo:
Thanks for the feedback, I'm happy to add some tests. 

Just to confirm before I do it, I'm thinking of putting them at the start of this [test_currencies](https://github.com/spiral-project/ihatemoney/blob/master/ihatemoney/tests/api_test.py#L578) and add asserts for a good status and check that `XXX` is within the list. Anything else I should look for?

almet:
Please, add them in the api tests instead, it's better to keep all the API-related tests together.

almet:
Works for me, thanks! We can merge as soon as the build is green :-)

almet:
Thanks :-)

PR #962, opened 2021-12-06
Added Length validator

mazzolaa:
Added wtforms Length validator. Applied this validator to the bill amount because large values will crash an instance of the project.

### DIFF
--- a/ihatemoney/forms.py
+++ b/ihatemoney/forms.py
@@ -20,6 +20,7 @@ from wtforms.validators import (
     NumberRange,
     Optional,
     ValidationError,
+    Length,
 )
 
 from ihatemoney.currency_convertor import CurrencyConverter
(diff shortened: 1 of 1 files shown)


almet:
Thanks, can you please add a test for this, on the API and on the main tests? Thanks!

almet:
(The build fails because you also have to run `make black` on your codebase).

Let us know if you need help or if you don't have the time on this, we can continue the work if needed :-)

mazzolaa:
Thanks for letting me know!
I have the time and knowledge to go back and reformat using make black on my codebase.
I may not have the knowledge to add a proper test for this but I can definitely give it a try. 

I should have time to work on this in the upcoming week!

Glandos:
The use of the `Length` validator is clever, but it could limit some users with long arithmetic operations. However, I tried one with more than 20 characters, and I had to be **very** creative. So I guess this could be OK. I just wanted to raise this potential issue.

Glandos:
I'm sorry to close this in favor of #989
However, thanks for your PR that was a quick and working patch!

PR #963, opened 2021-12-09
Constrain dependency for Babel >= 2.9.1 to address CVE-2021-42771

zorun:
This makes sure that people upgrading their installation using `pip install -U ihatemoney` will get the fixed version of Babel. By default, `pip` doesn't upgrade dependencies if the currently installed version fulfills dependency constraints.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,6 +28,7 @@ install_requires =
     debts>=0.5,<1
     email_validator>=1.0,<2
     Flask-Babel>=1.0,<3
+    Babel>=2.9.1,<3  # CVE-2021-42771
     Flask-Cors>=3.0.8,<4
     Flask-Mail>=0.9.1,<1
     Flask-Migrate>=2.5.3,<4  # Not following semantic versioning (e.g. https://github.com/miguelgrinberg/flask-migrate/commit/1af28ba273de6c88544623b8dc02dd539340294b)


zorun:
I'm not sure about the impact in our case but the issue is kind of severe: https://tenable.com/security/research/tra-2021-14

@Glandos @almet do you have any opinion on constraining indirect dependencies like this?

almet:
That looks good to me, even if I dont understand how this could lead to a security issue for us?

almet:
To be more precise (with a real keyboard it's easier): I believe we should check that this CVE applies to our codebase and the way we use it. How an attacker could leverage this? I don't understand how atm, and as a result I don't think we need to solve anything urgently.

In other words : if we think this doesn't expose us, we might want to not pin indirect dependencies like this.

Glandos:
Well I tried some things like:
- http://127.0.0.1:5000/lang/kikoo
- http://127.0.0.1:5000/lang/../../../../../../../../../../tmp/en
- http://127.0.0.1:5000/lang/..%2F../../../../../../../../../tmp/en

But the last two are caught by the web server, and the path is normalized.
However, the first case lead to a bad behavior: the server generates an error 500, and the user can't get out of it without calling manually `/lang/en` or deleting the session cookie. But it means that it *could* load files, only within the current directory.

almet:
> However, the first case lead to a bad behavior: the server generates an error 500, and the user can't get out of it without calling manually /lang/en or deleting the session cookie

I believe this is another bug, and that we also should fix it. Checking that the file exists before setting the session cookie seems a wise choice for this. I opened #970 about it.

Glandos:
I guess this is fixed by #971 since language must be in the server whitelist. I don't think it can be set from elsewhere.

almet:
Okay, I suggest we close this then? Feel free to reopen if you think it was the wrong move.

Issue #957, opened 2021-12-03
High integer can corrupt demo database


TechyTommy3:
So yesterday, I was trying out the demo on ihatemoney.org. And, I entered in a really high integer (something as big as 9347242149381274732472348728748723473278472843.) Now, it is easy to repair, but it requires editing your database. I did this on my own database again (nothing important is on it), and it gave me an internal server error. And yes, the ihatemoney.org demo I accidently corrupted still isn't working.
**Steps to reproduce**
1. Go into demo
2. Add a new bill
3. Add a big number, like 9347242149381274732472348728748723473278472843
Some error handling needs to be added.

Glandos:
I don't know if it can be solved upstream, but see https://github.com/python-babel/babel/issues/821 that I just opened.

Glandos:
And in this project, I don't know what should be a real solution. Capping the max value is not really good, there will always be a user that have a use case for this.
Maybe check the value before formatting?

almet:
> Capping the max value is not really good, there will always be a user that have a use case for this.

I think we can cap to something really high, like 100 000 000 000. If people with this amount of money want to still use our software, they could open an issue ;-)

Glandos:
I'll go with #989. The `decimal.MAX_EMAX` seems to be lower than required, by really high enough for our use case. @almet do you think it's the right way to workaround?

Issue #948, opened 2021-11-26
History from previous project is shown


Glandos:
- Create a project
- Add some participants
- Show participants creation in history
- Delete the project
- Create it again
- See that old entries are shown in history

This lead to data leak. History must be cleared upon project deletion.

deepthivalisetti:
Hi. I am trying to solve this issue. Can you let me know if the project is deleted, should the user data also be erased?
As I see, the person is having project id linked and on that basis if the project name is the same as the old one, its fetching the data. I have come up with e options.

- remove the Person data linked to the project
- remove the Project linked to the Person

almet:
> Hi. I am trying to solve this issue. Can you let me know if the project is deleted, should the user data also be erased? As I see, the person is having project id linked and on that basis if the project name is the same as the old one, its fetching the data. I have come up with 2 options.
> 
>     * remove the Person data linked to the project
> 
>     * remove the Project linked to the Person

I don't think you've understood the issue we're facing, so let me explain it differently : the history is stored in a separate database table than the projects. When the project is deleted, history should be deleted too, but it's not the case at the moment.

To solve this issue, you need to also delete the extra information.

Glandos:
Yes, history is handled by SQLAlchemy-continuum, and there are some functions to manage this, but I still have to read its documentation :)

Glandos:
First investigations results:
- This should be done manually: https://github.com/kvesteri/sqlalchemy-continuum/issues/124#issuecomment-195017874
- Deleting project history on project delete seems a bit hard: what if we need an undelete on it?
- Maybe we can delete project history on project creation? Only when the old one is old enough? How much days? One week seems enough to me.

almet:
> Deleting project history on project delete seems a bit hard: what if we need an undelete on it?

Interesting :-) At the moment we don't have any way to do this, unless I missed it. I believe it's alright. If we want to be sure that it's meant to be deleted, then we could maybe let only the person with an access to the specified email receive a confirmation email to delete the project.

PR #964, opened 2021-12-09
Prepare next release

zorun:


zorun:
I think it's good to release soon so that @youegraillot 's work on Docker (#919) can be made available.

Some pending changes that I think should also make it into this new release:

- #963 security-related (to be discussed)
- #949 that fixes the rest of the performance issues
- #962 that fixes a validation bug #957
- #956 more UI fixes
- possibly #961 (welcome API feature) if it's ready, but that could also go into a future "feature" release 5.2.0

What do others think?

zorun:
And I forgot about this bug: #948 
But I think we don't have a fix just yet.

zorun:
Updated with latest changes, the next release number will be 5.2.0 (new features).

almet:
So we're good to go here, right?

zorun:
It seems so, thanks a lot @Glandos for fixing the last bugs!

I'll update the release notes and see if we're good to go.

zorun:
I have updated the release notes and everything looks good, but we have a regression that was just reported: #1005

It should be fixed before releasing, it causes an error with the default config on Docker.

zorun:
And release 5.2.0 is out!

PR #966, opened 2021-12-16
Add functionality to new project creation

### LINKED ISSUE #525: Simplify the creation of a new project

seanlu99:
Closes #525

Changed the new project creation on both the home page and create page (/create route). By changing both pages, we could have one creation flow kept simple and the other creation flow more versatile and customizable. 

Changes:

- Added placeholders to all fields for a better explanation
- For the creation page, added a few more fields for further customization (inspired by [https://splittypie.com/new](SplittyPie))

New home page:
Screen Shot 2021-12-16 at 5 22 47 AM

New creation page:
Screen Shot 2021-12-16 at 5 22 54 AM

### DIFF
--- a/ihatemoney/forms.py
+++ b/ihatemoney/forms.py
@@ -23,7 +23,7 @@ from wtforms.validators import (
 )
 
 from ihatemoney.currency_convertor import CurrencyConverter
-from ihatemoney.models import LoggingMode, Person, Project
+from ihatemoney.models import LoggingMode, Person, Project, db
 from ihatemoney.utils import (
     eval_arithmetic_expression,
     render_localized_currency,
@@ -196,6 +196,8 @@ class ProjectForm(EditProjectForm):
     # This field overrides the one from EditProjectForm
     password = PasswordField(_("Private code"), validators=[DataRequired()])
     submit = SubmitField(_("Create the project"))
+    # Field to create a member for the project
+    member_name = StringField(_("Your name"), filters=[strip_filter])
 
     def save(self):
         """Create a new project with the information given by this form.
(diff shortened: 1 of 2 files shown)


ihatemoney/forms.py
```diff
@@ -216,6 +218,15 @@ def save(self):
             logging_preference=self.logging_preference,
             default_currency=self.default_currency.data,
         )
+        # Check that the member name is not empty so you don't insert an empty
+        # Person object if no member name is given
+        if self.member_name.data != "":
```
almet, line 223: I think that this is Falsey, so we could use `if self.member_name.data:` rather than checking for an empty string.

ihatemoney/forms.py
```diff
@@ -216,6 +218,15 @@ def save(self):
...
+            # Create a person object and link it to this project
+            member = Person(project_id=self.id.data, name=self.member_name.data)
+            # Add the new Person object to the database
+            db.session.add(member)
+            # Commit the database changes
```
almet, line 228: I believe the code is self-explanatory here, we could remove the last two comments I believe.

ihatemoney/templates/forms.html
```diff
@@ -69,12 +69,19 @@
...
+    <!-- Only display private code if not home page -->
     {% if not home %}
-    {{ input(form.id) }}
+    {{ input(form.id, placeholder="Unique ID used to login") }}
+    {% endif %}
+    <!-- Fields that are always shown -->
+    {{ input(form.name, placeholder="Example: Trip to Paris") }}
+    {{ input(form.password, placeholder="Password to access project") }}
+    {{ input(form.contact_email, placeholder="Your email") }}
+    <!-- Only display member name and default currency if not home page -->
```
almet, line 80: I feel this is weird. My experience with the project creation page (not on the home) is that it's shown to users only if there is an error when validating the form from the home.

I don't think we should add some specific fields here in this case, and not in the home. From the UX perspective it feels weird : we should have it on both locations or nowhere.

almet (changes requested):
Hi, thanks for working on this! I believe it's heading in the right direction !

However, I don't really like how we now have two project creation flows with no distinct way to choose between the two.

If we really want to have a simple and a complex way of creating projects, we should probably add a link "show more options ?" or something similar which will lead the user to the another creation page with more options.

almet:
@seanlu99 Let us know if we can help you :-)

almet:
@seanlu99 without any answer from your side since 16/12/21, I'm closing your pull request. If you still want to work on this please let us know and we can reopen.

Issue #869, opened 2021-10-19
Allow translation of invitation emails


almet:
Currently, the invitation emails are [directly encoded in templates](https://github.com/spiral-project/ihatemoney/blob/cfbce99f195ea9b373c8b8fbf365e1562262c9de/ihatemoney/templates/invitation_mail.en.j2), and aren't part of the translated strings.

We should change this :-)

seakyith:
Hello, I would like to work on this issue. Can you show me which file that I need to work on?

seakyith:
Thank you for your help! I tried to run test based on the documentation with make test command. I ran into the following error

make: *** No rule to make target `test'.  Stop.

Any helps would be appreciated.

almet:
@seakyith You need to be in the ihatemoney repository for this makefile command to work. Hope you figured this out :-)

jsabherwal:
I could work on it this week if anyone else hasn't. I request you to assign this to me.

almet:
Please go ahead!

anirudhgad:
I have been trying to work on this for a while but I am still struggling to understand exactly what must be done, please let me know if a quick chat is possible. Thanks

toodla:
hello! do you still need help on this? I am a beginner (like, this would be my first time doing this)! Thank you.

ayushanand308:
Hi! I am a first time contributor. Is there anything I can help with?

afriedman412:
hey -- I'm working on this and having a little trouble figuring out an Jinja error

I did this ...

1. Renamed "invitation_mail.en.j2" to "invitation_mail.j2"
2. Added `{% trans %}` and `{% endtrans %}` tags to the beginning and end of "invitation_mail.j2"
3. Changed line 604 in `web.py` from ...
`message_body = render_localized_template("invitation_mail")`
to
`message_body = render_template("invitation_mail.j2")`

but now all the tests for the invite email (specifically the first two in `budget_test.py`) are throwing this error:
```
>   Someone using the email address {{ g.project.contact_email }} invited you to share your expenses for .
E   jinja2.exceptions.TemplateSyntaxError: expected token 'end of print statement', got '.'

ihatemoney/templates/invitation_mail.j2:5: TemplateSyntaxError
```

any ideas?

almet:
Hi Andy,

This seems related to Jinja not able to read the template. Is it possible that you also changed something else in this template ? You can get the diff with `git diff path_to_file` to be sure :-)

[...]

PR #967, opened 2021-12-16
#869 Fix Invitation Mail

anirudhgad:
This is my first contribution so I may have some minor errors but I think the functionality of the code should solve the Issue requested.
I used a google API call to create translation from english to any language without using manual templates by making edits to render_localized_.template function. Unfortunately the requirement of the issue to use jinja2 for automatic translation is not actually a functionality of jinja2 and will not work unless we have manual translations available which I feel defeats the purpose of the change.
Issue is that some times google translate may not be as grammatically correct as a manual translation.

### DIFF
--- a/ihatemoney/templates/inv.j2
+++ b/ihatemoney/templates/inv.j2
@@ -1,0 +1,13 @@
+Salut,
+
+Quelqu'un à l'aide de l'adresse e-mail {{g.project.contact_email}} vous a invité à partager vos dépenses pour "{{g.project.name}}".
+
+C'est aussi simple que de dire ce que vous avez payé, pour qui et combien cela vous a-t-il coûté, nous nous soucions du reste.
+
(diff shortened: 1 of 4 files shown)


almet:
Hi, thanks for taking the time to do this work, however, as you say the "Issue is that some times google translate may not be as grammatically correct as a manual translation".

I believe we should do a manual translation instead. You can work on this if you want, it's described in #869

almet:
Sorry, I feel this lacks some more explanations :-)

I don't think using google translation will solve this issue. We have people ready to volunteer some time in order to translate this project in different languages, so let's use this fantastic power :-)

I believe what we should do is use a `trans` block in the jinja template and then generate the translation files!

Let's continue the discussion in another pull-request if you still want to work on this :-)

anirudhgad:
I initially tried using the trans block in the jinja template to generate them but scrapped it because I thought the requirement was for automation I will make some changes to it and upload again!

almet:
Awesome, thanks !

PR #974, opened 2021-12-22
Fix : use url_for in showcase template.

### LINKED ISSUE #973: Showcase images are referenced without url_for(), resulting in buggy URLs in case of subdir install
The faulty code is here :

https://github.com/spiral-project/ihatemoney/blob/ff4de78b569610e1715859e79fca1dd6ca41b94c/ihatemoney/templates/showcase.html#L36

Here you reference "/static/showcase/1.webp" but it should be `"{{ url_for('static', filename='showcase/1.webp')}}"`.

Salamandar:
Fixes #973


almet:
Thanks!

PR #976, opened 2021-12-22
Some small cleanups

almet:


Glandos:
`make black isort` and it should be fine. I guess that we could put it as precommit hook or something.

almet:
I'm supposed to have this setup but apparently not heh. Thanks for the comment.

almet:
Updated, let's see if the build is green and then merge :-)

PR #981, opened 2022-01-14
Check python compatibility in CI

zorun:
It uses a nice tool called vermin: https://pypi.org/project/vermin/


Glandos:
What do you think about adding it into `dev` requirements and adding a rule into the Makefile, so that it could be run locally?
Or even in tox.ini by the way.

zorun:
No actually, you're right, it would be nicer if it runs locally, I hadn't thought about it. I will fix this.

PR #986, opened 2022-01-23
unpin sqlalchemy to 1.3

symphorien:
sqlalchemy-continuum's latest release supports sqlalchemy 1.4

I notably checked that the history feature works.


almet:
Thanks! Let's see what the tests have the say :-)

symphorien:
looks like the test failures are related to wtforms, not sqlalchemy

Glandos:
Don't took it personally, but dependabot was first on #800 :wink: 

In fact, it was easier to make the rebase after #988 on this than here. See you there!

PR #989, opened 2022-01-30
Check for too high values

### LINKED ISSUE #957: High integer can corrupt demo database

Glandos:
see https://github.com/python-babel/babel/issues/821

fix #957

### DIFF
--- a/ihatemoney/forms.py
+++ b/ihatemoney/forms.py
@@ -1,3 +1,4 @@
+import decimal
 from datetime import datetime
 from re import match
 from types import SimpleNamespace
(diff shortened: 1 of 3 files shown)


almet:
Seems good to me ! Thanks

PR #994, opened 2022-02-05
Purge project history on deletion

### LINKED ISSUE #948: History from previous project is shown

Glandos:
Fix #948

Note that there is no confirmation. The whole project is deleted right away, without any possibility to undo.

This is a bit of an anti feature from using SQLAlchemy-continuum, but the code can easily be adapted to delete the project history later.


almet:
Thanks!

PR #1000, opened 2022-03-07
Almet/qrcode

almet:
Add the ability to display a QRCode when inviting other people. I'm not sure how this should be done specifically, but at least it works.

My main concerns is that not everybody is using Money Buster, nor knows what it is, so we might want to make it explicit that it's an Android app. The URL encoded in the QRCode has a `ihatemoney://` scheme, so it might not work if MoneyBuster is not installed, and we don't want to lose people here.

Let me know what you think :-)

image

### DIFF
--- a/ihatemoney/templates/send_invites.html
+++ b/ihatemoney/templates/send_invites.html
@@ -28,7 +28,7 @@
         </tr>
         <tr>
             <td>
-                <h3>{{ _('Scan the QR Code') }}</h3>
+                <h3>{{ _('Scan QR code') }}</h3>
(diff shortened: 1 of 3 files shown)


Glandos:
> My main concerns is that not everybody is using Money Buster
Then we can add the link to `/mobile` along with the label `Scan QR code`, so that people will be guided to it. By the way, I think that PayForMe (iOS) will support this also, but I don't remember where I found this info…

almet:
Cool, I've included a link, it's now like this :
image

Glandos:
I've reworded to "Scan QR code". Otherwise, it's OK for me!

PR #1008, opened 2022-03-20
Hide "each" amount when there's only one recipient

### LINKED ISSUE #1002: Hide "each" amount when there's only one recipient
When there's only one person in the `For whom?` column (this is the case for the slight majority of my transactions), the `How much?` column is unnecessarily cluttered, with the same value always appearing twice. I think it would read a lot more nicely if instead of e.g. `12.95 (12.95 each)`, we just saw `12.95`.

I was going to try to open a PR for this, assuming it would be pretty simple, but it relies on some sort of template/macro system rather than a proper programming language, and I just don't have the time to look in to it.

Glandos:
Fixes #1002

### DIFF
--- a/ihatemoney/templates/list_bills.html
+++ b/ihatemoney/templates/list_bills.html
@@ -1,7 +1,10 @@
 {% extends "sidebar_table_layout.html" %}
 
 {%- macro weighted_bill_amount(bill, weights, currency=bill.original_currency, amount=bill.amount) %}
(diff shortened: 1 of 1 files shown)


Glandos:
This was quite easy. However, no test existed on this rendering so… do you think there should be one for this?

almet:
I think we're good to go :-)

PR #1016, opened 2022-04-21
Allow Invitation Email Translation

a-angeles13:
Should address Issue #869 

Added a generic invitation email template that uses jinja {% trans %} blocks to facilitate rendering without the render_localized_template function as suggested by almet. 

I also modified the babel.cfg file to remove the jinja extensions as they are now [included by default](https://jinja.palletsprojects.com/en/3.0.x/extensions/#:~:text=With%20Statement,extension%20no%20longer%20does%20anything.) and were causing `make update-translations` to fail. I also modified some f-strings that were being used in conjunction with babel's `gettext` alias `_()` because this[ causes extraction errors with pybabel](https://github.com/python-babel/babel/issues/715). 

Modified budget_test.py to include a test of translated invite emails to ensure quality. 

Language files in `ihatemoney/translations' should be ready to be populated with translations. Until then only English and French will be supported with unsupported languages defaulting to English.


a-angeles13:
The above tests (3.7, minimal, postgresql; 3.7, minimal, mariadb; 3.9, minimal, sqlite; 3.10, minimal, sqlite) are failing due to [depreciation in Werkzeug 2.1](https://werkzeug.palletsprojects.com/en/2.1.x/changes/#:~:text=werkzeug.locals%20and,deprecated%20code.%20%232276) 

Specifically, `self.local.__ident_func__` in line 215 of <test_path>/werkzeug/local.py is now depreciated and will not work in Python versions 3.7+. This was noted in a warning on line 34 of test of 3.7, minimal, postgresql in a previously successful merge. [link](https://github.com/spiral-project/ihatemoney/runs/5528226214?check_suite_focus=true#:~:text=../home/runner/work,self._local.__ident_func__).

I suggest that the merge check test suite be prioritized because any future pull requests will be blocked under any Werkzeug 2.1+.

Glandos:
Hi, thanks for taking care of this!
First, yes, the tests are failing, and it's not your fault, as far as I see.

Secondly, could you remove all modifications done to .mo/.po files? They are handled by Weblate, so we only need modifications in `messages.pot`

Finally, it's an open question: do we want to have a lot of sentences to translate? In my opinion, the whole invitation email is a block in itself, translating it into small pieces will not bring any reuse cases, and it can be harder for translators to find the right wording. It will generate a big box with a loooooong string, but on the other end, the template will be much simpler, with only one `{%trans%}` tag pair.

almet:
Superseded by https://github.com/spiral-project/ihatemoney/pull/1101. Thanks for your work, trying to get it integrated :-)

Issue #528, opened 2020-01-17
Store bills in rational cents rather than floats


Natim:
Storing bills as floats leads to a number of problematic edge cases.

Instead we should handle money as integers in cents and then display them as float in the end.

We did this change on the elm-ihatemoney project during yesterday hackathon with @gdeest and I think we should fix it in the API as well.

Natim:
Refs https://github.com/spiral-project/elm-ihatemoney/pull/16

almet:
I believe this is a duplicate of #165, isn't it?

Natim:
Yes but #165 was closed while we are still using float. And as you well know, we have a couple more bugs with the external lib too.

Natim:
More interesting research here: https://ren.zone/articles/safe-money

almet:
Yes, you're right : we should move away from these silly floats!

LeoMouyna:
Hey @Natim !
Are you sure you want to store integers ? After reading your article we might want to store it as decimal string: https://docs.python.org/3/library/decimal.html and do operations with fractions: https://docs.python.org/3/library/fractions.html
What do you think about this ?

Natim:
Yes you are exactly right.

rom1dep:
@LeoMouyna how do you plan on serializing `Fraction('1.1') + Fraction('0.3') + Fraction(100 ,3)`? as a `(521, 15)` tuple?

PR #1024, opened 2022-05-05
Draft: Cast change from float to decimal

asya99:
Initial commit to test pipeline; change float to rational data type

Original request that was not fulfilled : https://github.com/spiral-project/ihatemoney/issues/528

https://ren.zone/articles/safe-money

### DIFF
--- a/ihatemoney/forms.py
+++ b/ihatemoney/forms.py
@@ -340,7 +340,7 @@ class BillForm(FlaskForm):
 
     def export(self, project):
         return Bill(
-            amount=float(self.amount.data),
+            amount=decimal.Decimal(self.amount.data),
             date=self.date.data,
             external_link=self.external_link.data,
             original_currency=str(self.original_currency.data),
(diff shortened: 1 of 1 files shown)


asya99:
New commit is to implement cancel button in forms.html under ihatemoney/template

asya99:
New commit is to resize favicon.ico file. The file was redesigned from scratch and converted to be 16x16 px

Glandos:
Hi, and thanks for taking care of contributing to this project!
First, please submit one change at a time. This doesn't mean that a Pull Request can't change multiple file at once, but it should target one feature. For example, in your case, favicon and cancel button is not related to decimal store format, so they should be removed from here. If needed they can have their own pull request.

After that, I don't have any issue with starting this huge task by using `Decimal` but I think that more is needed. Let's run the tests, and see what's happening :wink:

almet:
Hi, and thanks for this.

- The return button has been implemented in https://github.com/spiral-project/ihatemoney/commit/0d511e03d8d29470b075e3808e3c79ed1ab3e791
- Thanks for the proposed change of favicon. I opened a specific pull request for it in #1102
- I didn't took your changes for the Decimal change, but don't hesitate to reopen something if you feel you want this to happen 

Cheers!

PR #1025, opened 2022-05-14
escape csv formulae

Glandos:
This is only needed for unsecure spreadsheet applications (hi Google Docs and MS Excel) that load formulae by default.

See https://owasp.org/www-community/attacks/CSV_Injection for some mitigation explanation. This is not complete, but it should be OK for now.

### DIFF
--- a/ihatemoney/tests/import_test.py
+++ b/ihatemoney/tests/import_test.py
@@ -596,6 +596,38 @@ class ExportTestCase(IhatemoneyTestCase):
                 set(line.split(",")), set(received_lines[i].strip("\r").split(","))
             )
 
+    def test_export_escape_formulae(self):
+        self.post_project("raclette", default_currency="EUR")
+
+        # add participants
+        self.client.post("/raclette/members/add", data={"name": "zorglub"})
+
+        # create bills
+        self.client.post(
+            "/raclette/add",
+            data={
+                "date": "2016-12-31",
+                "what": "=COS(36)",
+                "payer": 1,
+                "payed_for": [1],
+                "amount": "10.0",
+                "original_currency": "EUR",
+            },
+        )
+
+        # generate csv export of bills
+        resp = self.client.get("/raclette/export/bills.csv")
+        expected = [
+            "date,what,amount,currency,payer_name,payer_weight,owers",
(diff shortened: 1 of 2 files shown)


Glandos:
Fixes https://www.huntr.dev/bounties/5c63f9ed-3b0a-4774-b89a-0aa0a2187ec0/

iamshooter99:
@Glandos  The Severity of this report is not low you can see here it mentioned https://nvd.nist.gov/vuln/detail/CVE-2022-22121  Base Score: 8.0 HIGHVector:  CVSS:3.1/AV:N/AC:L/PR:L/UI:R/S:U/C:H/I:H/A:H

Glandos:
OK, I can't pretend I have the skills to argue with NVD, but here is the thing in this project:
- All data are public within a project
- All data are displayed in the HTML table, so a formula will be obviously here
- Exporting a bill list as a CSV needs a lot of clicks in our application, because it is by far not the main use case

Finally, and most importantly, as based on reports in https://www.contextis.com/us/blog/comma-separated-vulnerabilities or http://georgemauer.net/2017/10/07/csv-injection.html it means that the person who exports the data must use an insecure software to trigger this. [CSV RFC](https://datatracker.ietf.org/doc/html/rfc4180) is clear on that: no formulae in this format. We can't take responsibility for bugs in other software. If a JSON parser decide to parse values starting with "=" using `eval()` it will be a nightmare for everyone.

Side note: don't send your [comment](https://github.com/spiral-project/ihatemoney/pull/1025#issuecomment-1126860199) [twice](https://github.com/spiral-project/ihatemoney/commit/042b33aeb2e87ad652e1e0478ce1b2b6bfcb9210#commitcomment-73676271). It sends two notifications. It also send the signal that you are looking for a recognition of something I can't provide. Maybe it's not case, I'm just explaining what this behavior creates.

Thanks for the report!

iamshooter99:
ok, Thank you. 
However, I'm not expecting any recognition from you. I'm learning every day new to explore as much as possible and to secure our Open Source community. 
I explained everything to show the highest impact of this vulnerability based on functionality existence. Now some might consider it as far not the main use case.

Anyways, I'm not expecting any recognition from you.
Happy Hacking.

PR #1028, opened 2022-05-28
fix CSRF on logout (aka /exit)

Glandos:
See https://huntr.dev/bounties/a3045614-1125-4901-bb7a-9d51be4beeed/

### DIFF
--- a/ihatemoney/run.py
+++ b/ihatemoney/run.py
@@ -8,6 +8,7 @@ from flask_babel import Babel, format_currency
 from flask_mail import Mail
 from flask_migrate import Migrate, stamp, upgrade
 from flask_talisman import Talisman
+from flask_wtf.csrf import generate_csrf
 from jinja2 import pass_context
 from markupsafe import Markup
 import pytz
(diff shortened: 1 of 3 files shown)


Glandos:
The fact that there is no CSRF token of `/exit` seemed very harmless to me. However, I dared fighting the challenge to add it in a `GET` request. It wasn't able to find something really relevant in the documentation, so I built my own solution.
The main bogging thing is the `if … WTF_CSRF_ENABLED` that was needed for tests. I don't like it, but I wasn't able to find something better.
@zorun since you implemented the last CSRF hardening, do you have a strong opinion on this?

zorun:
Wait, do we really change the state with a GET request? We're not supposed to do this.

@Glandos did you try just switching this to a POST request?

Glandos:
This is a right move… But is there a nice way to make a `POST` from the menu entry?

zorun:
You need to use a form with just the CSRF token and a button, and then tweak CSS if you want the button to look like a link.

zorun:
I thought we already had something like this, but apparently not...

Glandos:
[**@zorun**](https://github.com/zorun) commented on [29 mai 2022, 16:04 UTC+2](https://github.com/spiral-project/ihatemoney/pull/1028#issuecomment-1140456470 "2022-05-29T14:04:41Z - Replied by Github Reply Comments"):
> I thought we already had something like this, but apparently not...

Hehe, exactly, we are missing a nice confirmation screen.

So I think I will make a new PR, with a "Do you confirm logout" dialog that will have a form. And we can get rid of this edge case by having a normal CSRF-protected form.

Glandos:
Superseeded by #1040

PR #1033, opened 2022-06-23
Create robots.txt

scout249:
Disallow search engine indexing the site by creating robot.txt file


Glandos:
The whole site is nearly inaccessible without login.
If you want your own instance not to be indexed, then it's up to each administrator. But I think this is not a good default. Moreover, `robots.txt` can even be at the wrong place, if the instance is running in a subdirectory.

PR #1040, opened 2022-07-12
Fix CSRF on logout

Glandos:
Superseed #1028

### DIFF
--- a/ihatemoney/tests/budget_test.py
+++ b/ihatemoney/tests/budget_test.py
@@ -79,7 +79,7 @@ class BudgetTestCase(IhatemoneyTestCase):
             url_start = outbox[0].body.find("You can log in using this link: ") + 32
             url_end = outbox[0].body.find(".\n", url_start)
             url = outbox[0].body[url_start:url_end]
-        self.client.get("/exit")
+        self.client.post("/exit")
         # Test that we got a valid token
         resp = self.client.get(url, follow_redirects=True)
         self.assertIn(
@@ -87,7 +87,7 @@ class BudgetTestCase(IhatemoneyTestCase):
             resp.data.decode("utf-8"),
         )
         # Test empty and invalid tokens
-        self.client.get("/exit")
+        self.client.post("/exit")
         # Use another project_id
         parsed_url = urlparse(url)
         resp = self.client.get(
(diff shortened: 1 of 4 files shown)


ihatemoney/web.py
```diff
@@ -534,11 +536,23 @@ def export_project(file, format):
...
 
-@main.route("/exit")
+@main.route("/exit", methods=["GET", "POST"])
 def exit():
-    # delete the session
-    session.clear()
-    return redirect(url_for(".home"))
+    # We must test it manually, because otherwise, it creates a project "exit"
+    if request.method == "GET":
+        abort(405)
```
zorun, line 543: Why not just remove GET from the supported methods for this route?

zorun:
Thanks, the new approach looks good!

Just to clarify, in the end you didn't implement a confirmation dialog? That's totally fine for me, I think logging out needs no confirmation. But you should maybe remove "confirm" in the name of the form (`ConfirmLogoutForm`) to make that clear.

ihatemoney/web.py
Glandos: I tried this, but then, entering `/exit` will redirect to the project creation page with the project id `exit`. I bet this is unexpected, and prefer to cover this weird use case with custom handling.

Glandos:
> Thanks, the new approach looks good!
> 
> Just to clarify, in the end you didn't implement a confirmation dialog? That's totally fine for me, I think logging out needs no confirmation. But you should maybe remove "confirm" in the name of the form (`ConfirmLogoutForm`) to make that clear.

You're right: no confirmation, since this is far easier to code, and the UX is smoother. And you're right, I let the old name, it's meaningless now, so I renamed it.

ihatemoney/tests/budget_test.py
```diff
@@ -498,8 +498,12 @@ def test_authentication(self):
             self.assertIn("raclette", session)
             self.assertTrue(session["raclette"])
 
-            # logout should wipe the session out
+            # logout should work with POST only
             c.get("/exit")
```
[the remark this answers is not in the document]
Glandos: Exactly. I messed up the test here, so I changed it. Thanks!

ihatemoney/web.py
zorun: Ah, annoying indeed...

zorun (approved):
Everything good for me, thanks!

PR #1045, opened 2022-07-17
Created OpenAPI specification

autumnull:
The specification can be viewed in the editor at https://editor.swagger.io/

I have done my best to make the specification compliant with the actual behavior of the api, testing using the `ihatemoney.org` instance.

This specification can also be used to generate better documentation in the `readthedocs.io` page, like this: https://docs.gotosocial.org/en/latest/api/swagger/

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -1,0 +1,724 @@
+openapi: "3.0.3"
+servers:
+- url: "https://ihatemoney.org/api"
+  description: "The public I Hate Money test instance"
+info:
+  description: "All of what’s possible to do with the website is also possible via a web API. This document explains how the API is organized and how you can query it.\n\nThe main supported data format is JSON. When using POST or PUT, you can either pass data encoded in JSON or in `application/x-www-form-urlencoded` format.\n\nYou can access three different things: projects, members and bills. You can also get the balance for a project.\n\nThe examples here are using curl, feel free to use whatever you want to do the same thing, curl is not a requirement."
+  version: "5.2.0"
+  title: "I Hate Money"
+  license:
+    name: "I Hate Money license"
+    url: "https://github.com/spiral-project/ihatemoney/blob/master/LICENSE"
+externalDocs:
+  description: "I Hate Money documentation"
+  url: "http://swagger.io"
+security:
+- basic: []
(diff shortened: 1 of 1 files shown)


Glandos:
Thanks for this hard work. Did you do this manually? In this case: well done!
The OpenAPI/Swagger definition is not *really* a goal for IHateMoney itself, but it's very important nonetheless because of the ecosystem (Cospend/MoneyBuster/PayForMe).
So I'll be delighted to integrate this work. However, I have a question. How to ensure that API spec is up to date with the code?
I've done some search about existing libraries that generate OpenAPI definitions:
- https://github.com/flasgger/flasgger Seems old, but example with Flask-RESTful. Need YAML in docstring
- https://github.com/marshmallow-code/flask-smorest Lightweight and maintained. Need marshmallow definitions as code
- https://github.com/luolingchun/flask-openapi3 Maintained but replace Flask-RESTful, but with support for Blueprints.
- https://github.com/jmcarp/flask-apispec Lightweight and somewhat maintained. Need marshmallow code, but using decorators
- https://github.com/danleonard-nj/swagger-gen Very young, but work with "simple" decorator.

Do you have any experience/opinion on these?

autumnull:
I wrote this specification manually based on the REST API specification on the readthedocs page, and the behavior of the instance hosted at `ihatemoney.org`. 

Unfortunately I have no experience/opinion on those libraries. It does seem like a good idea to keep the spec up to date with the code though, and manually rewriting the spec every time would be a chore. Hopefully this initial contribution will help get the ball rolling on that.

Glandos:
@autumnull did you have any reason to delete your fork?
If it is just based on our slow reaction time, it's sad, but don't worry, we won't throw your work away!

autumnull:
Oh woops I actually completely forgot why I had a fork of this repo so i just deleted it while i was cleaning up my repositories -- entirely an accident! Hope the file changes are still useful :heart:

almet:
Hi, I'm not sure what to do with this. It seems that it would be better to use an automated generation of the API rather than having this done manually. If so, we should probably close this and lead to using one of these libraries.

Would that be okay for you?

autumnull:
yep no worries :)

PR #1058, opened 2022-08-28
Integrate flask-limiter

Glandos:
For the moment, it's just the integration of flask-limiter instead of the old and hand-made LoginThrottler.

Open discussion: since we are about to use limiter in more places than admin login, I modified the main layout to display limit breached. But… should we displayed the content block in this case? Or just optionnally, based on some variable?

### DIFF
--- a/ihatemoney/forms.py
+++ b/ihatemoney/forms.py
@@ -295,7 +295,9 @@ class AuthenticationForm(FlaskForm):
 
 
 class AdminAuthenticationForm(FlaskForm):
-    admin_password = PasswordField(_("Admin password"), validators=[DataRequired()])
+    admin_password = PasswordField(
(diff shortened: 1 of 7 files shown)


Glandos:
By the way, Flask-Limiter is Python>3.6

almet:
We probably shouldn't display the content block if we reached the max limit.

Should we merge this?

Glandos:
Here you are!

Yes, I also agree that the content block shouldn't be displayed. At first, I didn't implemented a message in the template, so I relied on the content block to explain why it was blocked.
But now… we could just not show it.

By the way, I will first merge #1083 first, to be sure tests are OK on 3.7+

almet:
Crap, I merged without wanting to. Hope things will be okay.

Glandos:
It's Python: better ask for forgiveness than for permission!

Joke aside, we can revert it if needed. Or better: improve it!

PR #1061, opened 2022-09-09
Captcha value should be case insensitive on both side (form and l10n values)

### LINKED ISSUE #1060: Creating new project on hosted site fails on captcha
Sorry in advance if this bug tracker is not meant for issues with the hosted site.

I tried to create a new project but the captcha validation always fails. I am 99.9% sure I am human and answering it correctly:

Bildschirmfoto_2022-09-09_09-19-07

I tried in multiple browsers but always got the verification fail.

Glandos:
Fixes #1060


kannes:
Thank you!

PR #1072, opened 2022-10-03
Updated the readme

saroj-regmi:
Added the contributer image and the link to their profile using contrib.rocks to give them some more respect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,14 @@ You can also [donate some
 money](https://liberapay.com/IHateMoney/donate). All funds will be used
 to maintain the [hosted version](https://ihatemoney.org).
 
+Join the Contributers already contributing to this Repository.
+
+<a href="https://github.com/spiral-project/ihatemoney/graphs/contributors">
+  <img src="https://contrib.rocks/image?repo=spiral-project/ihatemoney" />
+</a>
+
+Made with [contrib.rocks](https://contrib.rocks).
+
 ## Translation status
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,6 +37,14 @@ You can also [donate some
 money](https://liberapay.com/IHateMoney/donate). All funds will be used
 to maintain the [hosted version](https://ihatemoney.org).
 
+Join the Contributers already contributing to this Repository.
```
almet, line 40: Typo: I believe it should be contribut**o**rs (no capital C, and o instead of e). Repository should also not be capitalized.

Something like "Join the other contributors" would work.

README.md
```diff
@@ -37,6 +37,14 @@ You can also [donate some
...
+
+<a href="https://github.com/spiral-project/ihatemoney/graphs/contributors">
+  <img src="https://contrib.rocks/image?repo=spiral-project/ihatemoney" />
+</a>
```
almet, line 44: We're using Markdown instead of HTML here, so it would be better to use it here as well :-)

README.md
```diff
@@ -37,6 +37,14 @@ You can also [donate some
...
+
+Made with [contrib.rocks](https://contrib.rocks).
```
almet, line 46: I don't think we need to name the tool here.

almet:
Thanks! Why not, I added a few remarks.

saroj-regmi:
@almet I have updated the pull request, would you mind verifying it. Thank you

almet (approved):

PR #1083, opened 2022-10-23
Update Python

Glandos:
- [X] Drop Python 3.6
- [x] Use Python 3.11
- [x] Increase version

### DIFF
--- a/.github/workflows/test-docs.yml
+++ b/.github/workflows/test-docs.yml
@@ -36,7 +36,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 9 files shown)


Glandos:
@almet @Natim @zorun Do you have any strong opinion about this?

Natim:
I think we could drop support to 3.7 too.

Natim (approved):

Glandos:
For me… Yes, we can. But since all our dependencies still supports 3.7, it's nearly effortless. For now!

almet:
:+1:

Issue #654, opened 2020-07-17
Split a project by year / archive old bills


zorun:
I have used the same project for 5 years, and it is now too big to be usable: it has 1200 bills and it's all really slow (both for the server and for the browser). I had already improved IHateMoney's performance when I was having performance issues with "just" 350 bills (#161) but it's no longer enough. And now I understand that small projects fit better in the philosophy of IHateMoney.

A nice way to solve this situation would be to "split" a project by year: given a (big) project, create many smaller projects by grouping bills by year.  That is, from a project named `myproject`, it would create:

- `myproject-2015`
- `myproject-2017`
- `myproject-2018`
- etc

This way, the project for the current year would become much smaller.

It could also be seen as a way to archive old bills, so that you only keep bills for the current year in the main project.

Another use-case is to set different weights to participants depending on the year: at the end of 2019, you could "archive" all bills from 2019 in a separate project, and then decide which weights should be applied to 2019 (for instance based on your total income of the year). You would then settle the 2019 project based on these weights, and import it as an initial debt in the 2020 project.

In any case, I'm not sure it's a very common situation, so I'll leave this ticket open to see if somebody else has the same needs.

zorun:
I have a quick & dirty implementation of splitting by year, implemented as a command in `manage.py`. I will publish it at some point.

Glandos:
I use the same project to balance bills in my house. But once in a while, we do the math and money transfer, and clear everything. Accounting history is done in another software, so my project is never big enough.

Don't you think pagination (#480) can improve this?

zorun:
Now the main issue for me is really to have different weight for each year, exactly like you (based on average income for the year).

Pagination does help for performance (this is why I developed it ;) ), but the "Statistics" page still takes 15 seconds to load with 1200 bills.

I ended up splitting the history of bills by year (using the quick & dirty command I was mentioning above). And now, we just create a new project every year.

zorun:
To be more accurate: we add bills during the year, and then, at the end of the year (or even a bit later), we change the weights of the participants according to their average income during the year.

This works very nicely, and it would be much more difficult to do with per-bill weights (which I think would be too complex anyway)

Glandos:
This would be "solved" by fixing #471. However, for things like this, I would **much** prefer having options in the project settings to do that: Splitting a project according to some criteria, etc. Having a subproject feature seems overkill in this situation.

PR #1092, opened 2022-11-13
Archive and Filter by Year Features

Huda11Baig:
Resolves spiral-project/ihatemoney#654

The added code implements a feature where a user can archive bills by pressing the added archive icon. The unarchived and archived bills are displayed on separate pages which can be navigated by a button placed next to the Add Bill button.

Additionally, a text field has been added under unarchived bills that allow users to enter a year. All the bills are then filtered by the year specified. Non-number or non-existent years generate an empty list.


ihatemoney/templates/list_bills_archived.html
```diff
@@ -0,0 +1,180 @@
+{% extends "sidebar_table_layout.html" %}
```
almet, line 1: It seems to be a copy/paste from the more generic list_bills if I'm not mistaken. What about refactoring these two so it's easier to make edits to the way bills are rendered in general?

ihatemoney/web.py
```diff
@@ -646,6 +646,33 @@ def list_bills():
     )
 
 
+@main.route("/<project_id>/archived")
```
almet, line 649: I also feel that this is a lot of copy/pasting. You probably want to remove some code that's not needed here (no need to support the addition of new bills), we probably should also remove the possibility to add a bill when we're browing the archives.

almet:
Thanks for your work on this. I've added a few comments and questions, don't hesitate to answer and discuss these :-)

Issue #1077, opened 2022-10-11
Feature Request: Add a confirmation on expense deletion


chmac:
I just deleted a duplicated expense in our budget, and if I'm not mistaken, it was instantly removed without any confirmation dialog. As deletion is a destructive operation, I'd be in favour of asking the user to confirm this action before carrying it out. This is the web UI which I'm referring to.

Even a simple `globalThis.confirm()` would be good enough from my perspective.

Side note, would you be open to a PR which adds this option? If yes, it would be great to get some pointers on where this code lives, if there's any translations to consider, and any suggestions on what message to use in the dialog. I'd go with "Are you sure?" as a first step otherwise.

Glandos:
I think a confirmation can be useful. Until we implemented an Undo feature, which I would much prefer.
Anyway, in the meantime, there is already that kind of confirmation in the project deletion workflow. The form is rendered here: https://github.com/spiral-project/ihatemoney/blob/master/ihatemoney/templates/edit_project.html#L36
And the little JS snippet to make a double click button is right above: https://github.com/spiral-project/ihatemoney/blob/master/ihatemoney/templates/edit_project.html#L5

You can try to make a confirmation based on this.

chmac:
Thanks for the detailed pointers. I'm not sufficiently familiar with the codebase to attempt this kind of change I'm afraid. I was pretty confident I could find a spot to add the correct `globalThis.prompt()`, but that is a lot simpler.

I'm happy to leave this ticket open in case somebody else picks up on it and wants to submit a PR? Maybe adding the "good first issue" label would make sense?

FlowingCloudRTL:
Hello! is it ok for me to work on this issue?

chmac:
@FlowingCloudRTL It seems like this is an issue which is open to any contributor, so I'd imagine that you're welcome to work on it. But just to be clear, I'm not a member of the project, so I don't know what the policies are for merging pull requests and so on.

Glandos:
Fixed by #1096

PR #1096, opened 2022-12-06
Feature Request: Add a confirmation on expense deletion #1077

cbrosnan:
Completed feature request #1077 to use the double-click confirmation button that is also found in edit_project.html. This will prevent the accidental deletion of expenses (a destructive operation) with no warning.

### DIFF
--- a/ihatemoney/static/css/main.css
+++ b/ihatemoney/static/css/main.css
@@ -310,6 +310,11 @@ footer .footer-left {
   background: url("../images/delete.png") no-repeat right;
 }
 
+.bill-actions > form > .confirm.btn-danger {
+  outline-color: #c82333;
+  box-shadow: none;
+}
+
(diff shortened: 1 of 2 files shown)


almet:
This looks good to me. I feel we could refactor some code with the exact same thing happening with user deletion (just a few lines up in the JS). What do you think?

almet:
Also, see #1099

cbrosnan:
I followed the JS convention that @Glandos pointed out in the original feature request (#1077). It is the same as the confirmation that is found in the project deletion workflow. I personally thought that the design looked better and drew better attention to the expense deletion than the user deletion confirmation. What are your thoughts?

almet:
I thought originally that we could do some refactoring, but you're right :  let's merge this :-) Thanks

PR #1103, opened 2022-12-10
Test the CI (do not merge)

almet:
So it seems that the CI is not working at the moment. More specifically, docker test doesn't work : it cannot connect to the remote and then fails.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# I hate money
+# I Hate Money
 
 [![GitHub Actions Status](https://github.com/spiral-project/ihatemoney/actions/workflows/test-docs.yml/badge.svg)](https://github.com/spiral-project/ihatemoney/actions/workflows/test-docs.yml)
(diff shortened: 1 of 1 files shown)


almet:
I'm not sure what's happening. It works locally when I run the action alone, but doesn't remotely.

Here's the code https://github.com/spiral-project/ihatemoney/commit/ec0a3aa07050a6933d239e64eb7631a64e84d1ff

zorun:
This part of the CI seems to be working again now.

zorun:
Damn, it happened to me today, but it's kind of random.

I believe it's because we don't wait for Ihatemoney to be ready before running the `wget` test.

Simple fix: we could just sleep for a few seconds.

Proper fix: we would need to tell docker-compose about healthchecks and add a dependency between the two containers.

@youegraillot any thoughts?

zorun:
Now fixed by #1139 (hopefully)

PR #1104, opened 2022-12-10
Bugfix #1080: "Deactivated users reappeared"

### LINKED ISSUE #1080: Deactivated users reappeared
I created a project on https://ihatemoney.org/ for my house. Two flatmates left, I balanced their accounts to 0€ and deactivated them. Several expenses later, they reappeared with a positive balance, even though they are still deactivated. How is that possible and how to prevent it?
Merci d'avance.

chestnutFan:
Fixed #1080 by adding the deactivated member check. Corresponding test functions were also finished.<br>
Details:<br>1. Added the deactivated member check for editing/deleting bills. 
Compare the payer&owers list (POL) of the bill with the active member list (AML). If the relative complement of AML in POL (POL \ AML) is not empty, then the bill contains deactivated members and should not be edited/deleted.<br>2. Added two corresponding integration test functions.<br>
Test:<br>Passed all the tests.

### DIFF
--- a/ihatemoney/web.py
+++ b/ihatemoney/web.py
@@ -806,6 +806,10 @@ def delete_bill(bill_id):
     if not bill:
         return redirect(url_for(".list_bills"))
 
+    # Check if the bill contains deactivated member. If yes, stop deleting.
+    if bill.involves_deactivated_members:
+        return redirect(url_for(".list_bills"))
+
     db.session.delete(bill)
     db.session.commit()
     flash(_("The bill has been deleted"))
@@ -819,6 +823,10 @@ def edit_bill(bill_id):
     bill = Bill.query.get(g.project, bill_id)
     if not bill:
         raise NotFound()
+
+    # Check if the bill contains deactivated member. If yes, stop editing.
+    if bill.involves_deactivated_members:
+        return redirect(url_for(".list_bills"))
 
     form = get_billform_for(g.project, set_default=False)
 
--- a/ihatemoney/models.py
+++ b/ihatemoney/models.py
@@ -752,6 +752,22 @@ class Bill(db.Model):
         else:
             return 0
 
+    @property
+    def involves_deactivated_members(self):
+        """Check whether the bill contains deactivated member.
+        Return:
+        True if it contains deactivated member,
+        False if not.
+        """
+        owers_id = [int(m.id) for m in self.owers]
+        bill_members = owers_id + [self.payer_id]
+        deactivated_members_count = (
+            Person.query.filter(Person.id.in_(bill_members))
+            .filter(Person.activated.is_(False))
+            .count()
+        )
+        return deactivated_member_count != 0
+
     def __str__(self):
         return self.what
 
--- a/ihatemoney/templates/list_bills.html
+++ b/ihatemoney/templates/list_bills.html
@@ -148,10 +148,22 @@
                     </span>
                 </td>
                 <td class="bill-actions d-flex align-items-center">
-                    <a class="edit" href="{{ url_for(".edit_bill", bill_id=bill.id) }}" title="{{ _("edit") }}">{{ _('edit') }}</a>
+                    <a class="edit" href="{{ url_for(".edit_bill", bill_id=bill.id) }}" data-toggle="tooltip"
+                        {% if bill.involves_deactivated_members %}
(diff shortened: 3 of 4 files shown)


ihatemoney/web.py
```diff
@@ -760,6 +760,19 @@ def delete_bill(bill_id):
     if not bill:
         return redirect(url_for(".list_bills"))
 
+    # Prevent deleting if the bill involves deactivated user.
+    active_member_id = [m.id for m in g.project.active_members]
+    owers_id = [int(m.id) for m in bill.owers]
+    deactivated_members = set([bill.payer_id] + owers_id).difference(
+        set(active_member_id)
+    )
```
almet, line 768: I believe it's possible to ask the model for deactivated users, rather than building it yourself here. You need to look for (`activated=False`).

ihatemoney/web.py
```diff
@@ -776,6 +789,19 @@ def edit_bill(bill_id):
...
+    if len(deactivated_members):
+        flash(
+            _("Deactivated users involved. This bill cannot be edited."),
+            category="warning",
+        )
```
almet, line 802: I think having this message shown on a tooltip next to the delete or edit button of the bill would be better, don't you think?

almet (changes requested):
Thanks for your work on this. It's leading the right way, but needs some changes before integration.

Some feedback : 

1. At all times, avoid copy-pasting code. Here the code to get the deactivated member list is in two different places. If we needed this code, it should probably go in the model instead ;
2. I believe we don't actually need this code because the model can already tel us what are the deactivated members. Try to do this instead of building the list on your own
3. In terms of UX, I believe it would be better to have a tooltip as I explained. I'm happy to change my mind on this if you have good arguments :-)

Anyway, thanks for taking the time to do this :+1:. Let me know when you've made some edits.

ihatemoney/web.py
chestnutFan: Thanks for the suggestion! Will change it.

ihatemoney/web.py
chestnutFan: Good idea. A tooltip makes more sense.

chestnutFan: Hi Almet. I looked into the tooltip thing and found it kind of challenging to me and would take some time to finish. Would you like to keep the flashed message for now?

I'll keep working on it and launch a new pull request once I'm done.

chestnutFan:
Hi Almet! Thanks for reviewing my code. I rewrote the check function using the activated attribute, and moved it to the model.

ihatemoney/web.py
almet: I prefer to merge the pull request once everything is ready :-) Take the time you need and don't hesitate to ask for help. What's blocking you?

chestnutFan: Thanks bro. Just some small issues. Could be done soon.

chestnutFan: Tooltips have been added.

chestnutFan:
Hi Almet! Tooltips have been added.

almet:
This is a bit old now, but I wanted to continue working on it, as I'm currently in the process of tidying a bit this repository.

I've rebased this on top of the latest `main` branch, and it can be merged. But... I'm actually starting to think we're fixing something that is not a problem: if you delete or edit a bill that was involving a deactivated user, I believe we should make it reappear, since its balance is now changed.

Making it so that it's not possible doesn't actually seem right to me, as there are cases where we want to actually change this, for instance if a mistake was made.

So, I'm planning on closing this and #1080 at the same time, unless somebody points me why we should do this.

Cheers!

chestnutFan:
> This is a bit old now, but I wanted to continue working on it, as I'm currently in the process of tidying a bit this repository.
> 
> I've rebased this on top of the latest `main` branch, and it can be merged. But... I'm actually starting to think we're fixing something that is not a problem: if you delete or edit a bill that was involving a deactivated user, I believe we should make it reappear, since its balance is now changed.
> 
> Making it so that it's not possible doesn't actually seem right to me, as there are cases where we want to actually change this, for instance if a mistake was made.
> 
> So, I'm planning on closing this and #1080 at the same time, unless somebody points me why we should do this.
> 
> Cheers!

Sounds reasonable to me. Glad to see this one finally been nailed :)

PR #1106, opened 2022-12-10
Updated checks for validate_name() in MemberForm()

mzhongqi:
The database allows users to deactivate an account with a non-zero value, and create a new user with the same name, reactivating the previous user will allow two users of the same name. This change assures that new user names can not be the same as deactivated users with associated bills (Users that are not deleted from deactivation).


almet:
Thanks for this, it seems good to me, but lacks a test to ensure that what you're doing works the way you intend. Are you okay to write such a test? Thanks!

mzhongqi:
> Thanks for this, it seems good to me, but lacks a test to ensure that what you're doing works the way you intend. Are you okay to write such a test? Thanks!

almet:
Probably a manipulation error?

mzhongqi:
> Probably a manipulation error?

Yes that was a manipulation error, sorry! I have just added two tests for this particular change at #1111.

zorun:
Thanks for the patch, I added your new tests from #1111 so that we have everything in one place.

Two tests are failing:

- your new test `test_add_duplicate_user`: the last call should be a POST, not a GET
- the existing test `test_membership`: I believe this test is wrong, it uses `/<project_id>/members/add` to reactivate the user, but it should use `/<project_id>/members/<member_id>/reactivate` instead

@mzhongqi can you fix both tests?

PR #1107, opened 2022-12-11
New Click Command

adammentzer:
Added Click command to display the number of projects. Options exist to specify for projects within the bounds of the specified number of bills (inclusive) and/or less than x days old.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -1,0 +1,192 @@
+#!/usr/bin/env python
+
+import datetime
+import getpass
+import os
+import random
+import sys
+
+import click
+from flask.cli import FlaskGroup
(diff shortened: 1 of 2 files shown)


almet:
Hi, and thanks for your changes. It seems however than I can't merge them directly, because you're proposing to use the files in a different location than the one they are already in.

You need to use git, checkout the project locally and then edit the files and propose a change. We've done some documentation about that, it's [here](https://ihatemoney.readthedocs.io/en/latest/contributing.html).

If it's simpler for you, don't hesitate to close this pull request and start a new one. If you're lost, let me know :-)

Issue #978, opened 2022-01-14
Docker latest and master: deleting a project in admin-dashboard


AlexanderSch90:
See #977 
If i delete a project from the admin dashboard: I also get an error... 
```
Method Not Allowed
The method is not allowed for the requested URL.
```

almet:
This seems to be related to secure cookies being enabled by default on docker. Closing as I suspect this is solved @AlexanderSch90? Let us know if not :-)

SchoofLux:
I’m still experiencing the same issue. I cannot delete any project from the Admin dashboard.
I’m using docker which works perfectly otherwise.
Problem occurs with SESSION_COOKIE_SECURE set to True or False. No difference.

PR #1108, opened 2022-12-11
Fix project deletion in the dashboard. Fixes #1094

### LINKED ISSUE #1094: Deleting a project in admin-dashboard
I’m still experiencing the same issue as #978. I cannot delete any project from the Admin dashboard.
I’m using docker which works perfectly otherwise.
Problem occurs with SESSION_COOKIE_SECURE set to True or False. No difference.

almet:
This was broken due to changes introduced to make project deletion more secure. Rather than doing some complicated if branches, I decided this dashboard stuff is probably better handled with separate routes instead.

So I've reintroduced a way to delete the projects without specifying the project code (otherwise it's not possible for admins to do it).

### DIFF
--- a/ihatemoney/static/css/main.css
+++ b/ihatemoney/static/css/main.css
@@ -14,18 +14,22 @@ body {
 .navbar-brand span {
   color: #abe128;
 }
+
 .navbar h1 {
   font-size: 1rem;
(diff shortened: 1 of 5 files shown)


Glandos:
This will fix #978

PR #1110, opened 2022-12-11
Solved the issue deactivated user's bill cannot be edited

MelodyZhangYiqun:
Solved #1080 that deactivated user's bill can still be edited.
Before, this system allows to deactivate user with non-zero bill value. However, the bill can still be edited by other active member.
After, the deactivated user's bill cannot be edited with a prompt noticing the users.
Test has been made with this fix.


almet:
Superseded by #1104, thanks for the contribution and sorry for letting this bitrot for a bit.

PR #1113, opened 2022-12-12
Bill_type added to bill object as an attribute

FlowingCloudRTL:
Bill type is added as an attribute of Bill. Currently the bill types that are available to be used is hardcoded in the project model, where 
`bill_types = [
        ("Expense", "Expense"),
        ("Reimbursment", "Reimbursment"),
        ("Refund", "Refund"),
        ("Transfer", "Transfer"),
        ("Payment", "Payment"),
    ]`
One can access the bill_type attribute the same way as any other attribute of the bill object, and all functionalities related to bill has been updated to be compatible with the new attribute


Glandos:
Superseeded by #1116

PR #1114, opened 2022-12-13
New Click Command

adammentzer:
Added Click command to display the number of projects. Options exist to specify for projects within the bounds of the specified number of bills (inclusive) and/or less than x days old.

### DIFF
--- a/ihatemoney/manage.py
+++ b/ihatemoney/manage.py
@@ -4,6 +4,7 @@ import getpass
 import os
 import random
 import sys
+import datetime
 
 import click
 from flask.cli import FlaskGroup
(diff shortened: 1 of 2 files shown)


ihatemoney/manage.py
```diff
@@ -94,5 +95,97 @@ def delete_project(project_name):
...
+):
+    """Displays the number of projects. Options exist to specify for \
+        projects within the bounds of the specified number of bills (inclusive) and/or less than x days old. """
+    if bills_low and bills_high:
+
+        # Make sure low < high
+        if bills_high < bills_low:
+            temp = bills_high
+            bills_high = bills_low
+            bills_low = temp
```
Glandos, line 137:
```suggestion
        if bills_high < bills_low:
            (bills_low, bill_high) = bills_high, bills_low
```

ihatemoney/manage.py
```diff
@@ -94,5 +95,97 @@ def delete_project(project_name):
...
+
+        if num_days:
+            projects = [
+                pr
+                for pr in Project.query.all()
+                if pr.get_bills().count() >= bills_low
+                and pr.get_bills().count() <= bills_high
```
Glandos, line 144:
```suggestion
                if  bills_low <= pr.get_bills().count() <= bills_high
```

ihatemoney/manage.py
```diff
@@ -94,5 +95,97 @@ def delete_project(project_name):
...
+        else:
+            projects = Project.query.all()
+
+        click.echo(len(projects))
+        if emails:
+            list_emails = ", ".join(set([pr.contact_email for pr in projects]))
+            click.echo(list_emails)
+        if names:
+            proj_names = [pr.name for pr in projects]
+            click.echo(proj_names)
```
Glandos, line 187: This is duplicated code from line 155-161.

Glandos (changes requested):
I've added some small Pythonic changes.

But globally, it would better to have a list of predicates to filter the list of project, and execute this list comprehension `project = []` only once.
I bet this is possible with SQLAlchemy, but don't have the time to find it, so, it will be fine for me with Python lambdas :)

Glandos:
Also, maybe a default value of 0 for `bills_low` should be a better default, instead of having to chose both low and high.

almet:
Thanks, this has been done in #1115

PR #1132, opened 2023-01-31
Fix several issues with translations

zorun:


zorun:
@Glandos I don't remember how this works, does Weblate handle updating and compiling translation files?

Both were broken (updating translations because of f-strings, and compiling translations because of errors in the spanish translation), which surprises me.

Glandos:
Yes, as you can see in the last [PR from Weblate](https://github.com/spiral-project/ihatemoney/pull/1084/files). So don't change `.po`/`.mo` unless it is really necessary, otherwise Weblate will probably need a conflict resolution :)

Glandos:
But of course, `messages.pot` is not extracted automatically by Weblate, so it can (and must) be committed.

zorun:
Ok, currently `make update-translations` regenerates both `messages.pot` AND all `messages.po`. If that's not the right way to do it nowadays, we should change that and update the contribution documentation.

I have only kept the fix for Spanish because it's really an error.

PR #1133, opened 2023-01-31
Allow bills with an amount of zero

zorun:
Bills with an amount of zero may be useful to remember that a transaction happened on a specific date, while the amount doesn't matter.

I use that with per-year projects when a reimbursement happens in year N but is relative to year N-1: I record it with an amount of zero in the project of year N, and with the real amount in the project of year N-1.

Besides, it's already possible to create such bills: while "0" is refused, "0.0" is accepted.  There are no visible issues with this kind of bills.

### DIFF
--- a/ihatemoney/forms.py
+++ b/ihatemoney/forms.py
@@ -387,9 +387,7 @@ class BillForm(FlaskForm):
         self.payed_for.data = self.payed_for.default
 
     def validate_amount(self, field):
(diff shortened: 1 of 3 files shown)


Glandos (approved):

zorun:
Thanks for the historical perspective :D

I had to rebase, I will merge when the CI completes.

PR #1135, opened 2023-01-31
tests: Avoid plain "assert" and use unittest helper functions

zorun:


azmeuk:
Out of curiosity, what motivates such a decision (as pytest can display legible error messages out of plain `assert`)?

zorun:
Consistency! We had a mix of both styles (but mostly unittest-style)

PR #1141, opened 2023-02-07
[CI] Build mutli-architecture docker image

### LINKED ISSUE #1066: Docker ARM build
Hello. Would it be possible to publish and maintain an arm build of the Docker Hub image, so that it can be ran from a Raspberry Pi, etc.?

greedybro:
I built an arm/v7 image locally with buildx and run it on my raspberry pi 3 successfully (https://hub.docker.com/r/greedybro/ihatemoney/tags)

Since the project CI is already using buildx via `docker/build-push-action@v2`, it should only require to had a list of architectures to build for.

The list should cover most of user needs with `armv7` for rapsberry 3 and before, and `arm64` for Mac with Apple silicon and raspberry 4 (and probabilly future models)

Closes  #1066

### DIFF
--- a/.github/workflows/dockerhub.yml
+++ b/.github/workflows/dockerhub.yml
@@ -61,6 +61,7 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           cache-from: type=local,src=/tmp/.buildx-cache
           cache-to: type=local,dest=/tmp/.buildx-cache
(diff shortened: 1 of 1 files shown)


rriski:
Would be great to get this in! I'm also currently building my own images for ARM.

Glandos:
Well, I'm not a Docker person, but this seems the right way :)

PR #1149, opened 2023-02-25
Migrate to pytest

### LINKED ISSUE #1148: Migrate to pytest
Currently, we are using pytest for **launching** tests, but they are all written with [Flask-Testing](https://github.com/jarus/flask-testing) and unittest.

Since Flask-Testing is nearly dead, and Flask is moving fastly, we should move away from it, and use another way to write test. As far as I can tell, the [base documentation for testing Flask](https://flask.palletsprojects.com/en/2.2.x/testing/) can easily use fixtures of pytest.

In this issue, I want to discuss what we are currently using, and what could be done to move to pytest.

However, there might be other library I'm not aware of.

Glandos:
Fix #1148

Well for now, this is just experiment, but showing good signs of success.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -134,7 +134,7 @@ To run the tests:
 
     make test
 
-Tests can be edited in `ihatemoney/tests/tests.py`. If some test cases
+Tests can be edited in `ihatemoney/tests/`. If some test cases
 fail because of your changes, first check whether your code correctly
 handle these cases. If you are confident that your code is correct and
 that the test cases simply need to be updated to match your changes,
@@ -143,6 +143,18 @@ update the test cases and send them as part of your pull request.
 If you are introducing a new feature, you need to either add tests to
 existing classes, or add a new class (if your new feature is
 significantly different from existing code).
+
+#### Old tests and new tests
+Historically, tests were written with unittest and Flask-Testing.
+Because the latter was dying, we removed it and switch as much as
+possible to pytest fixtures. However, a lot of tests were already written,
+and are still using class-based tests, albeit without setUp()/teardown().
+
+If you write new tests, please **try to write them with pytest in mind**:
+- standalone function (not in class)
+- plain asserts
+
+For further reading, go to [pytest project documentation](https://docs.pytest.org)
 
 ### Formatting code
 
--- a/ihatemoney/tests/budget_test.py
+++ b/ihatemoney/tests/budget_test.py
@@ -1374,6 +1374,7 @@ class BudgetTestCase(IhatemoneyTestCase):
         member = models.Person.query.filter(models.Person.id == 1).one_or_none()
         self.assertEqual(member, None)
 
+    @pytest.mark.usefixtures("app_ctx")
     def test_currency_switch(self):
         # A project should be editable
         self.post_project("raclette")
@@ -1500,6 +1501,7 @@ class BudgetTestCase(IhatemoneyTestCase):
         self.assertIn('<p class="alert alert-danger">', resp.data.decode("utf-8"))
         self.assertEqual(self.get_project("raclette").default_currency, "USD")
 
+    @pytest.mark.usefixtures("app_ctx")
     def test_currency_switch_to_bill_currency(self):
         # Default currency is 'XXX', but we should start from a project with a currency
         self.post_project("raclette", default_currency="USD")
@@ -1534,6 +1536,7 @@ class BudgetTestCase(IhatemoneyTestCase):
         bill = project.get_bills().first()
         self.assertEqual(bill.converted_amount, bill.amount)
 
+    @pytest.mark.usefixtures("app_ctx")
     def test_currency_switch_to_no_currency(self):
         # Default currency is 'XXX', but we should start from a project with a currency
         self.post_project("raclette", default_currency="USD")
--- a/ihatemoney/tests/budget_test.py
+++ b/ihatemoney/tests/budget_test.py
@@ -119,6 +119,7 @@ class BudgetTestCase(IhatemoneyTestCase):
         resp = self.client.get("/raclette/join/token.invalid", follow_redirects=True)
         self.assertIn("Provided token is invalid", resp.data.decode("utf-8"))
 
+    @pytest.mark.usefixtures("app_ctx")
     def test_multiple_join(self):
         """Test that joining multiple times a project
         doesn't add it multiple times in the session"""
--- a/ihatemoney/tests/common/ihatemoney_testcase.py
+++ b/ihatemoney/tests/common/ihatemoney_testcase.py
@@ -1,104 +1,10 @@
 import os
 import unittest
-from unittest.mock import MagicMock
 
-from flask_testing import TestCase
 import pytest
 from werkzeug.security import generate_password_hash
 
 from ihatemoney import models
-from ihatemoney.currency_convertor import CurrencyConverter
-from ihatemoney.run import create_app, db
-
-
-
-class BaseTestCaseOld(TestCase):
-    SECRET_KEY = "TEST SESSION"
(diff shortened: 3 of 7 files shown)


Glandos:
Here it is!
Flask-Testing is now removed, we will be able to upgrade everything else! And write more `pytest`-esque tests.

Glandos:
Ah, we need an explanation/discussion/debate. Good!
**Disclaimer**: I like the pytest pythonic approach of writing tests. This is purely subjective, so I think it's good to say it before reading.

**fixtures are reusable `setUp()`/`teardown()`**. You just write some code in a function (the `setUp`), then `yield` (the value, or nothing), then some code (the `teardown`). And you can write small functions, they all contains, in one go, their `setUp/teardown` phases. This is a huge win for me.

Now, the main disadvantage for me are the following implicit behavior:
- Globally available fixtures must be in a file called `conftest.py`. It took me an hour to find it.
- Using fixtures can be automatic, based on the test function argument names. E.g., a function `test_my_client(app, client)` will try to use fixtures called `app` and `client`. This can be confusing.

Honestly, for now, I took the lazy approach: don't modify every existing tests, since they use a class and `self` everywhere. So I declared some small fixtures (to make them a bit independent), and use them wherever needed.

**MagickMock and patch**: as far as I know, nothing like this exist in pytest. They have a module for monkeypatching, but the one from unittest is far more useful in my opinion, so I'll stick to that. Since `unittest` is part of standard library, we are free to use when we want to.

zorun:
Yeah, I don't like implicit fixtures. If we do move to pytest, we should document/enforce that we only use explicit fixtures (like the code you propose here)

I don't understand why you need to wrap some test code with a `with self.app.app_context():`, can you explain?

docs/contributing.md
```diff
@@ -144,6 +144,18 @@ If you are introducing a new feature, you need to either add tests to
 existing classes, or add a new class (if your new feature is
 significantly different from existing code).
 
+#### Old tests and new tests
+Historically, tests were written with unittest and Flask-Testing.
+Because the latter was dying, we removed it and switch as much as
+possible to pytest fixtures. However, a lot of tests were already written,
+and are still using class-based tests, albeit without setUp()/teardown().
+
+If you write new tests, please **try to write them with pytest in mind**:
```
almet, line 153: Seems to be missing a newline here as well.

docs/contributing.md
```diff
@@ -144,6 +144,18 @@ If you are introducing a new feature, you need to either add tests to
 existing classes, or add a new class (if your new feature is
 significantly different from existing code).
 
+#### Old tests and new tests
+Historically, tests were written with unittest and Flask-Testing.
+Because the latter was dying, we removed it and switch as much as
+possible to pytest fixtures. However, a lot of tests were already written,
+and are still using class-based tests, albeit without setUp()/teardown().
```
almet, line 151: put the method names within backticks for formatting.

docs/contributing.md
```diff
@@ -144,6 +144,18 @@ If you are introducing a new feature, you need to either add tests to
...
+
+If you write new tests, please **try to write them with pytest in mind**:
+- standalone function (not in class)
```
almet, line 154: Maybe a sentence would be better, like "tests are done in standalone functions rather than in class and class methods".

docs/contributing.md
```diff
@@ -144,6 +144,18 @@ If you are introducing a new feature, you need to either add tests to
...
+- plain asserts
```
almet, line 155: Use plain `assert` rather tha `this.assertTrue`, for instance.

docs/contributing.md
```diff
@@ -144,6 +144,18 @@ If you are introducing a new feature, you need to either add tests to
 existing classes, or add a new class (if your new feature is
 significantly different from existing code).
 
+#### Old tests and new tests
+Historically, tests were written with unittest and Flask-Testing.
```
almet, line 148: Rather than starting by explaining how it was in the past, it might be better to explain how it's supposed to be now, and then explain why we have legacy code around. Maybe something like :

> Tests are now using [pytest](https://docs.pytest.org/). However, [...].

ihatemoney/tests/budget_test.py
```diff
@@ -119,6 +119,7 @@ def test_invite(self):
         resp = self.client.get("/raclette/join/token.invalid", follow_redirects=True)
         self.assertIn("Provided token is invalid", resp.data.decode("utf-8"))
 
+    @pytest.mark.usefixtures("app_ctx")
```
almet, line 122: Is there a way to avoid repeating this line every so often?

ihatemoney/tests/conftest.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    This fixture add both app_context AND request context for ease of use.
+    If you only need an app_context locally, use `with self.app.app_context():`
+    in your code.
+    """
+    with app.app_context(), app.test_request_context():
+        yield
+
+
+@pytest.fixture
```
almet, line 45: I'm not sure if it's possible with pytest, but any chance this could be in the same file as the tests?

almet (approved):
Hey, looks good to me! I'm glad to see the tests are moving a bit. I guess they could be refactored a lot and reworked to be simpler now.

I approve the changes. I've a few comments about the docs, because I feel it would be clearer this way. Happy to go use pytest now rather than the old unittest way.

ihatemoney/tests/conftest.py
azmeuk: `conftest.py` is the place for shared fixtures, but pytest allows to define fixtures in test files :ok_hand:

ihatemoney/tests/budget_test.py
azmeuk: I would recommend having a look at [pytest-flask](https://pypi.org/project/pytest-flask/). I think this is not documented, but this allows among other things to access the flask app context directly in the tests (without `with app.app_context(): ...`

ihatemoney/tests/common/ihatemoney_testcase.py
```diff
@@ -1,45 +1,21 @@
...
 from werkzeug.security import generate_password_hash
 
 from ihatemoney import models
-from ihatemoney.currency_convertor import CurrencyConverter
-from ihatemoney.run import create_app, db
 
 
-class BaseTestCase(TestCase):
+@pytest.mark.usefixtures("client", "converter")
+class BaseTestCase(unittest.TestCase):
```
azmeuk, line 11: Inherinting from `unittest.TestCase` is not required with pytest.

https://docs.pytest.org/en/7.1.x/getting-started.html#group-multiple-tests-in-a-class

azmeuk:
Any reason why sticking to `self.assert*` unittest assertions? Nowadays pytest assertions produce legible error messages, are easier to write and easier to read, though I know not everybody shares this opinion in the python community. Is this a unit test style choice in IHM or do you plan to move on in a second step?

almet:
No, there are no reason. If we are to decide to go the pytest way, I believe it's just a migration we have to do, but which is not done yet.

@zorun, you seem to be a against using pytest. Would that be a big problem for you? Is it mainly a problem with fixtures?

On my side, I don't have a strong opinion about this. I like pytest and have been using it in other projects without trouble. I would feel more confortable with it, but I can also keep things as they currently are.

zorun:
I'm not strongly opposed to pytest, I just don't like implicit fixtures. If we can avoid them, then let's move to pytest.

The other thing I find really strange is the need for the app_context sometimes, but @azmeuk mentioned that pytest-flask should solve this point.

More work for @Glandos I guess :D

azmeuk:
> More work for @Glandos I guess :D

I can gladly help with the pytest migration if ever you don't have a lot of time to spend @Glandos.

zorun:
Superseded by #1213, thanks @Glandos for the initial work!

PR #1179, opened 2023-07-11
Prepare next release

zorun:


Glandos:
I should have merged #1149 earlier. But you know how time flies…

Here is my thought: make the release, if something break, we'll fix it. Oh, wait, you already made it :tada:

zorun:
Don't worry, we always need a bugfix release shortly after a major release :)

I would like us to move to a much more frequent release cycle. It could even be for every feature / bunch of bugfixes.

PR #1199, opened 2023-07-27
Drop python 3.7

azmeuk:
[Python 3.7 reached end of life on 27 Jun 2023](https://devguide.python.org/versions/). This PR removes tests for python 3.7, updates the documentation to reflect the new minimum python version. I also applied [pyupgrade](https://github.com/asottile/pyupgrade) to automatically update the codebase to a more recent syntax: `find ihatemoney conf ihatemoney -name "*.py" -exec pyupgrade --py38-plus "{}" +`.

### DIFF
--- a/.github/workflows/test-docs.yml
+++ b/.github/workflows/test-docs.yml
@@ -36,25 +36,25 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.7, 3.8, 3.9, "3.10", "3.11"]
+        python-version: [3.8, 3.9, "3.10", "3.11"]
         dependencies: [normal]
         database: [sqlite]
-        # Test other databases only with one version of Python (Debian buster has 3.7)
+        # Test other databases only with one version of Python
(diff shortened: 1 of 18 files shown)


zorun:
As usual, NACK for me, at least for as long as Debian buster is supported, and probably even more. That's at least June 2024: https://wiki.debian.org/LTS

Even then, I would need good arguments why supporting Python 3.7 is preventing us from doing something.

People don't necessarily update their server OS very often, and we have to think of downstream packaging such as in Yunohost.

I'll leave this open in case @Glandos @almet want to chime in.

azmeuk:
> As usual, NACK for me, at least for as long as Debian buster is supported, and probably even more.

This is a good point. I did not realized that some supported debian would ship some eol python.

almet:
I share the concerns with @zorun on this. I'm closing, hope it's okay for you @azmeuk

azmeuk:
No problem I totally understand :ok_hand:

PR #1202, opened 2023-07-28
Update security docs for the new feed token

zorun:


almet:
Thanks, I completely forgot about this doc :-)

PR #1203, opened 2023-07-28
unit test assertion fixes

azmeuk:
Playing with the pytest branch I found that there were occurences of `self.assertTrue(200, resp.status_code)` style assertions in the tests. This really looks like someone intended to write `self.assertEqual` instead, as testing that any number is True will always pass :see_no_evil:

### DIFF
--- a/ihatemoney/tests/api_test.py
+++ b/ihatemoney/tests/api_test.py
@@ -94,7 +94,7 @@ class APITestCase(IhatemoneyTestCase):
             },
         )
 
-        self.assertTrue(400, resp.status_code)
+        self.assertEqual(400, resp.status_code)
         self.assertEqual(
             '{"contact_email": ["Invalid email address."]}\n', resp.data.decode("utf-8")
         )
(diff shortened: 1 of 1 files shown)


ihatemoney/tests/api_test.py
```diff
@@ -577,19 +577,19 @@ def test_bills_with_calculation(self):
     def test_currencies(self):
         # check /currencies for list of supported currencies
         resp = self.client.get("/api/currencies")
-        self.assertTrue(201, resp.status_code)
+        self.assertEquals(200, resp.status_code)
```
azmeuk, line 580: I supposed that `200` is really what's wanted here, instead of `200`, as this is not an object creation.

almet:
Haha, you're right, and I'm probably the one who introduced this a few years back. Thanks for the changes.

ihatemoney/tests/api_test.py
almet: Yes, it's supposed to be 200.

PR #1204, opened 2023-07-28
Require private code to edit a project settings, document better the invite options

zorun:
This is something we had documented in our [security documentation](https://ihatemoney.readthedocs.io/en/latest/security.html#giving-access-to-a-project), but we didn't actually do it...

As mentioned in this document, this has good security properties: you can invite somebody with an invitation link, and they will be able to access the project but not change the private code (because they don't know the current private code).

This new check also applies to all other settings (email address, history settings, currency), which is desirable.  Only somebody with knowledge of the private code can now change these settings.

Finally, document these security properties in the invitation page.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -34,9 +34,9 @@ the token (of course, you need to authenticate):
     $ curl --basic -u demo:demo https://ihatemoney.org/api/projects/demo/token
     {"token": "WyJ0ZXN0Il0.Rt04fNMmxp9YslCRq8hB6jE9s1Q"}
 
-Make sure to store this token securely: it allows full access to the
+Make sure to store this token securely: it allows almost full access to the
 project. For instance, use it to obtain information about the project
-(replace PROJECT_TOKEN with the actual token):
+(replace `PROJECT_TOKEN` with the actual token):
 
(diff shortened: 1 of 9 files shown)


zorun:
The tests need fixing, but I want to start the discussion.

Any opinion about it @almet @Glandos ? To be honest, I thought we were already doing it, I'm the one who wrote that in our documentation...

almet:
That makes sense to me, go ahead with this :-)

zorun:
Ok :)

The good thing I had not anticipated: it also restricts changing the settings through the API if you just have a token (which is the same as the invite token).  So, you also need the current private code to change settings:

```
$ curl --oauth2-bearer 'WyJmb28iXQ....'  -X PUT http://localhost:5000/api/projects/foo -d 'name=newfoo&id=newfoo&password=newbar&contact_email=foo@example.com'
{
    "current_password": [
        "This field is required."
    ]
}
```

I need to update tests, add new tests that check this, and update the API doc.

zorun:
Crap, you can still delete a project through the API with just an auth token, so it's not strictly equivalent :(

zorun:
OK, I'm going to merge this, and leave the project deletion issue for #1206. This is already an improvement.

PR #1213, opened 2023-08-12
pytest migration

### LINKED ISSUE #1148: Migrate to pytest

azmeuk:
supersedes #1149 
closes #1148

This PR basically starts from 1149 and:

- adds pytest-flask for test readability
- replace setUp/tearDown with pytest fixtures
- rename test classes to use the pytest convention
- use pytest assertions

I suppose more could be done, for instance by moving away from the test class structure to a flat test files structure, and reorganize fixtures too, but maybe this is a good enough first step?

### DIFF
--- a/ihatemoney/tests/main_test.py
+++ b/ihatemoney/tests/main_test.py
@@ -1,9 +1,9 @@
 import os
 import smtplib
 import socket
-import unittest
 from unittest.mock import MagicMock, patch
 
+import pytest
 from sqlalchemy import orm
 from werkzeug.security import check_password_hash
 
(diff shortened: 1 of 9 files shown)


almet:
Wow, this is huge work. Thanks a lot for this :-)

ihatemoney/tests/main_test.py
```diff
@@ -368,11 +364,11 @@ def test_api_project_creation_does_not_need_captcha(self):
                 "contact_email": "raclette@notmyidea.org",
             },
         )
-        self.assertTrue(resp.status, 201)
```
TheTiz, line 371: Is this assert right at the origin ? assertTrue take only one parameter, second is an optional msg.

Did we want to assert resp.status == 201 here or something else ?

azmeuk: Good catch. This is fixed.

almet (approved):
I skimmed quickly trough the changes and they look good to me.

Waiting for a second pair of eyes though.

zorun:
@Glandos can you review this?

zorun:
It's been one-and-a-half month so I'm going to merge this (before it gets out-of-date because of unrelated changes)

azmeuk:
Thank you!

PR #1215, opened 2023-08-12
speed up unit tests

### LINKED ISSUE #1214: unit tests are slow
Testing the whole test suite takes 59s for the 118 tests on my computer (which is very capable).

With [pytest-xdist](https://pypi.org/project/pytest-xdist/) and 8 CPUs this goes down to 17s (`pytest -n auto`)
 
`pytest --durations 10` gives us those stats:
```
2.08s call     ihatemoney/tests/api_test.py::TestAPI::test_project
1.99s call     ihatemoney/tests/api_test.py::TestAPI::test_bills_with_calculation
1.97s call     ihatemoney/tests/history_test.py::TestHistory::test_clear_ip_records
1.88s call     ihatemoney/tests/api_test.py::TestAPI::test_member
1.77s call     ihatemoney/tests/api_test.py::TestAPI::test_bills
1.62s call     ihatemoney/tests/api_test.py::TestAPI::test_currencies
1.31s call     ihatemoney/tests/budget_test.py::TestBudget::test_edit_project
1.08s call     ihatemoney/tests/history_test.py::TestHistory::test_disable_clear_no_new_records
1.02s call     ihatemoney/tests/history_test.py::TestHistory::test_project_privacy_edit2
1.00s call     ihatemoney/tests/history_test.py::TestHistory::test_project_privacy_edit
```

Looking a bit more closely with [pytest-profiling](https://pypi.org/project/pytest-profiling/) and [snakeviz](https://pypi.org/project/snakeviz/), I can see that the longest function call is [werkzeug.security.check_password_hash](https://werkzeug.palletsprojects.com/en/2.3.x/utils/#werkzeug.security.generate_password_hash) from the `need_auth` decorator:

https://github.com/spiral-project/ihatemoney/blob/0187932365b8e13505d341d784db6294696bba17/ihatemoney/api/common.py#L28

It seems to be [a known issue](https://stackoverflow.com/questions/45375935/werkzeug-security-check-password-hash-takes-too-long), and one way to mitigate this is to generate lower security password hashes.

By using `method='pbkdf2:sha1:1000'` in `generate_password_hash` tests run in 24s (instead of 59s) or 8s with xdist (instead of 17s).

I suggest to use a faster and less secure hash algorithm to run unit tests, but still use the default algorithm in other environment. I think a simple solution would be to add two configuration parameters `PASSWORD_HASH_METHOD` and `PASSWORD_HASH_SALT` that would be passed as arguments in `generate_password_hash`. By default those vars would be unset, except in the unit test environment where they would have convenient values.

I would volunteer to provide a PR for this.

What do you think?

azmeuk:
Adds two configuration parameters that are passed to generate_password_hash:

- PASSWORD_HASH_METHOD
- PASSWORD_HASH_SALT_LENGTH

The unit tests use high-speed low-security values and gain 50% speed.

fixes #1214

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ This document describes changes between each past release.
 
 ## 6.1.1 (unreleased)
 
-- Nothing changed yet.
+- Speed up unit tests (#1214)
 
 
 ## 6.1.0 (2023-07-29)
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -188,3 +188,12 @@ project](https://pythonhosted.org/flask-mail/#configuring-flask-mail)
 -   **MAIL_USERNAME** : default **None**
 -   **MAIL_PASSWORD** : default **None**
 -   **DEFAULT_MAIL_SENDER** : default **None**
+
+## Configuring password hashes
+
+The werkzeug [generate_password_hash](https://werkzeug.palletsprojects.com/utils/#werkzeug.security.generate_password_hash)
+is used to create password hashes. By default the default werkzeug values
+are used, however you can customize those values with:
+
+-  **PASSWORD_HASH_METHOD** : default **None**
+-  **PASSWORD_HASH_SALT_LENGTH** : default **None**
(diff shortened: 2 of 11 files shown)


almet:
This is neat. Well done. Thanks!

PR #1221, opened 2023-08-28
Simplifies adding a bill with keyboard only

Jojo144:
This PR proposes three small changes to simplify adding a bill with only keyboard:

1. Makes the default focus on the "Add a new bill" button:
Capture d’écran de 2023-08-28 23-23-50

2. In the "Add a new bill" modal, makes the default focus on the first field:
Capture d’écran de 2023-08-28 23-26-03

3. Allows to escape the "Add a new bill" modal with the "Esc" key. (Was this intentionally deactivated?)

### DIFF
--- a/ihatemoney/templates/list_bills.html
+++ b/ihatemoney/templates/list_bills.html
@@ -11,6 +11,11 @@
 {% block js %}
     {% if add_bill %} $('#new-bill > a').click(); {% endif %}
 
+    // focus on first field when adding a bill
+    $("#bill-form").on('shown.bs.modal', function(){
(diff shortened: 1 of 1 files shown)


zorun:
Thanks, it looks good to me.

I just have one UX remark: isn't it a bit annoying to focus a date field by default? I guess it will display the date picker as soon as the modal opens?

Jojo144:
You're right. The datepicker did not display but I think it is more intuitive to focus on "What". I changed this:
Capture d’écran de 2023-08-30 10-34-35 resized

zorun:
Thanks!

PR #1222, opened 2023-08-30
Default owers when adding a bill

Jojo144:
# English

In short. Would you find nice to:
1. either remember the owers when adding several bills
2. or allow to define default owers in the settings of a project
?

(This PR implements 1.)

# Français

J'utilise régulièrement IHM dans un projet où je sélectionne toujours les mêmes "Pour qui ?" quand je saisi une facture. C'est un peu long de désélectionner à chaque fois.

Est-ce que vous seriez intéressés par une des propositions suivantes :
1. se rappeler, d'une facture sur l'autre, des derniers "Pour qui ?" sélectionnés (c'est actuellement le cas avec le "Qui a payé ?")
2. ou permettre de définir des "Pour qui ?" par défaut dans les paramètres du projet
?

Ma préférence va à 1 (interface plus légère). Et vous ?

(Cette PR implémente 1.)

### DIFF
--- a/ihatemoney/web.py
+++ b/ihatemoney/web.py
@@ -645,7 +645,7 @@ def list_bills():
     bill_form = get_billform_for(g.project)
     # Used for CSRF validation
     csrf_form = EmptyForm()
-    # set the last selected payer as default choice if exists
+    # set the last selected payer and last selected owers as default choice if they exist
     if "last_selected_payer_per_project" in session:
         if g.project.id in session["last_selected_payer_per_project"]:
             bill_form.payer.data = session["last_selected_payer_per_project"][
@@ -655,6 +655,11 @@ def list_bills():
     else:
         if "last_selected_payer" in session:
             bill_form.payer.data = session["last_selected_payer"]
+    if (
+        "last_selected_payed_for" in session
+        and g.project.id in session["last_selected_payed_for"]
+    ):
+        bill_form.payed_for.data = session["last_selected_payed_for"][g.project.id]
 
     # Each item will be a (weight_sum, Bill) tuple.
     # TODO: improve this awkward result using column_property:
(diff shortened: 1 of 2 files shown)


ihatemoney/web.py
```diff
@@ -647,7 +647,11 @@ def list_bills():
     csrf_form = EmptyForm()
     # set the last selected payer as default choice if exists
     if "last_selected_payer" in session:
-        bill_form.payer.data = session["last_selected_payer"]
+        if g.project.id in session["last_selected_payer"]:
+            bill_form.payer.data = session["last_selected_payer"][g.project.id]
```
Jojo144, line 651: *Question for Flask experts:*
When does the session expires? I changed the representation of `session["last_selected_payer"]`, does the session automatically expires during an update of IHM?

zorun: No, the session is kept even when you update IHM, so you always need to account for the old format and convert it. This has been a frequent source of heisenbugs in the past.

In general, can you separate the two changes into separate commits/PR? One which adds the new "payed_for", and one which fixes the behaviour of payer.

zorun:
I like the idea, and the "simple" approach with the session looks good to me.

As said above, please separate the new feature and the fix.

You can look at adding a test (including for the fix, to avoid a regression). You can find a similar test here I believe: test_session_projects_migration_to_list

ihatemoney/web.py
Jojo144: > In general, can you separate the two changes into separate commits/PR?

I'll do that.

Jojo144: I updated the PR. The bug fix is now in #1224 and the new feature in this PR.

zorun:
Woah, all tests fail with this error: `sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) duplicate column name: weight`

This looks totally unrelated to your change... Can you try to rebase?

zorun:
Sorry, you need to rebase because of the tests re-organisation + the flask-wtf fix

zorun:
Can you fix the conflict?

almet:
Perfect, thanks.

PR #1223, opened 2023-09-03
Add more details for bills in the history

Jojo144:
This PR add more details for bills in the history for each operation: insert, update, delete.

The idea is to be able to easily know what was a bill at some time (which can be difficult when it has been modified several times).

Capture d’écran de 2023-09-03 10-04-43

The PR also shorten long URL with an ellipsis `...` to avoid big overflows.

### DIFF
--- a/ihatemoney/history.py
+++ b/ihatemoney/history.py
@@ -83,9 +83,25 @@ def get_history(project, human_readable_names=True):
                 "time": version.transaction.issued_at,
                 "operation_type": version.operation_type,
                 "object_type": object_type,
+                "bill_details": None,
                 "object_desc": object_str,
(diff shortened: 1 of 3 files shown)


zorun:
The principle and implementation looks good to me, thanks!

Can you just change the strings to make translation easier:

- we use "change" elsewhere, not "modification"
- it's better to have two complete strings: "Details of the bill (before the change)" and "Details of the bill". Otherwise, the translators may be missing the context.

Jojo144:
Done!

zorun:
Thanks!

(Note: the test are failing, but this is unrelated, I fixed it in #1233 )

PR #1224, opened 2023-09-07
[Fix] Remember the last selected payer for each project

Jojo144:
Previously the last selected payer was lost when switching of project.

The structure of the variable `last_selected_payer` is changed (now it's a dictionary) then it has been renamed to `last_selected_payer_per_project` to avoid session bugs.

### DIFF
--- a/ihatemoney/tests/budget_test.py
+++ b/ihatemoney/tests/budget_test.py
@@ -2050,3 +2050,53 @@ class TestBudget(IhatemoneyTestCase):
 
         resp = self.client.get(f"/raclette/feed/{token}.xml")
         assert resp.status_code == 404
+
+    def test_remember_payer_per_project(self):
+        """
+        Tests that the last payer is remembered for each project
+        """
+        self.post_project("raclette")
+        self.client.post("/raclette/members/add", data={"name": "zorglub"})
+        self.client.post("/raclette/members/add", data={"name": "fred"})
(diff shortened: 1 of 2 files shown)


almet:
Hey, thanks for the changes.

Could you make some edits so that it's not lost for all users when we release the new version? You could do that by checking if `last_selected_payer` exists, and if it does use it to populate the new dict.

Also, could you add some tests about this, to ensure it's working the way it's intended? Thanks :-)

Jojo144:
> Could you make some edits so that it's not lost for all users when we release the new version? You could do that by checking if last_selected_payer exists, and if it does use it to populate the new dict.

@almet Are you sure of that? 99.9 % won't remark that the default user is lost during the update and it will produce dead code...

I have a look at the tests.

almet:
We can mark the code for deletion later with a comment, it's not an issue.

Thanks :-)

zorun:
I also don't think we need compatibility on this one, especially because it was already lost when switching project. But now it's done...

@Jojo144 you also need to rebase that one, sorry. Can you also adapt your test to use plain asserts?

azmeuk:
In the meantime #1213 has been merged. Can you use pytest assertions in unit tests?

Jojo144:
Done. Thanks for your comment.

zorun:
Thanks!

Issue #1232, opened 2023-09-28
Change the behaviour of the currency convertor


almet:
We currently use an external API to get conversion rates, and it went away. I think we should not rely on external services anymore to do these kind of tasks.

My proposal is to : 

- Let the project owner select a default currency (as we currently do, but without a list) on project creation, or via the project settings ;
- When creating bills, users can specify the currency and a conversion rate of their choosing. We could eventually select the last conversion rate that was recorded if that makes things easier.

Would that be okay for you? (cc @Glandos as git blame points to you on this code / last changes).

Cheers :-)

zorun:
I agree we should stop relying on an external service.

Your solution is probably the most reasonable but is still a bit complex. Let me try to widen the discussion by proposing more radical alternatives:

- Remove support for currencies. I'm only half-joking here, it's arguably the most complex and fragile part of Ihatemoney
- Keep currencies on bills, but only perform conversion when settling a project. This boils down to having a single manual conversion rate at the project level.
- Keep currencies on bills, but don't do any conversion. This probably means we have to show multiple balances and settling plans (one for each currency), which makes the UI more difficult to design

In any case, let's keep in mind the original use-case: handling bills in multiple currencies in a single project. For reference, here are the original discussion and pull request:

- #512
- #541

almet:
I actually would advocate for removing currencies all together. I'm not using this feature so I might be too optimistic, but I actually believe it's part of the features that we should have rejected at first to keep the codebase simple.

dylanmorroll:
As a British person living in The Netherlands with a Swiss partner I find the multiple currencies super useful! So I would definitely advocate for keeping it in haha. Found this issue cause my ihatemoney project has fully stopped working.

Although yeah I generally do not really like the automatic currency conversion and would actually prefer just having the currencies separate. I generally have some of each currency in my Revolut/Wise/whatever at the same time, so sometimes settle up in other currencies. Or, when I do decide to settle it all up in euros or something, I just check the current exchange rate and take that, so I don't really use the rate that ihatemoney is using anyway.

I guess the downside here is settling the bill could be a bit clunky. e.g. if I owed my partner 80 CHF and she owed me 100 EUR, settling the bill wouldn't be able to be reduced to a single transaction. But still, the maximum amount of different transactions you would have to do is one per currency. I think generally the number of different currencies would be 2 or 3, so it would still be fairly simple maths.

[...]

PR #1240, opened 2023-10-03
Currency hotfixes

zorun:
Hardcode list of currencies, and temporarily disable some currency operations to prevent crashes.

Here is what is disabled:

- setting or changing the default currency on an existing project
- adding or editing a bill with a currency that differs from the default currency of the project

Adjust and skip tests as necessary.

The real fix will come after #1232  is discussed


ihatemoney/tests/api_test.py
```diff
@@ -615,6 +617,7 @@ def test_bills_with_calculation(self):
             )
             self.assertStatus(400, req)
 
+    @pytest.mark.skip(reason="Currency conversion is broken")
```
almet, line 620: Yay, we benefit from new pytest markers :-)

zorun:
I'd like to release version 6.1.1 as soon as this is merged (which means really soon)

ihatemoney/tests/budget_test.py
```diff
@@ -1703,7 +1706,7 @@ def test_rss_feed(self):
         Tests that the RSS feed output content is expected.
         """
         with fake_time("2023-07-25 12:00:00"):
-            self.post_project("raclette")
+            self.post_project("raclette", default_currency="EUR")
```
almet, line 1709: Not sure why we need to specify the currency here?

almet (commented):
Hi, thanks for doing this.

What about adding a test scenario which reproduces the problems we have currently? So we can ensure this is fixed by the changes you're doing here?

If it's too long to do, I think we should merge this and release a bugfix release, but I think it could be a nice addition.

ihatemoney/tests/main_test.py
```diff
@@ -385,9 +385,9 @@ def test_only_one_instance(self):
         assert one == two
 
     def test_get_currencies(self):
-        assert set(self.converter.get_currencies()) == set(
-            ["USD", "EUR", "CAD", "PLN", CurrencyConverter.no_currency]
-        )
+        currencies = self.converter.get_currencies()
+        for currency in ["USD", "EUR", "CAD", "PLN", CurrencyConverter.no_currency]:
```
almet, line 389: I'm a bit confused with `CurrencyConverter.no_currency` here. Should it be `XXX`?

zorun: Well yes, it's the same thing.

ihatemoney/tests/budget_test.py
almet: Okay, thanks for the heads up.

zorun:
You mean simulating the exception that would be raised by the API call? It's a good idea, that would have allowed us to catch this way earlier. But I'm not motivated enough to do this now...

ihatemoney/tests/main_test.py
almet: [Heh, yes it is](https://github.com/spiral-project/ihatemoney/pull/1240/files#diff-9534886c8df967ba5f34f0c976eb6790ec62b2e39103822885bff66ad95a2d4cR19)

almet:
> You mean simulating the exception that would be raised by the API call? It's a good idea, that would have allowed us to catch this way earlier. But I'm not motivated enough to do this now...

I understand, especially as it's a quick fix for an emergency release. Let's move on then, I won't have the time to add it either.

Thanks again for doing that!

zorun:
Thanks for the review! I will make the release now.

Issue #1238, opened 2023-10-02
Do not version the compiled catalogs


azmeuk:
I just noticed `*.mo` files are versioned in the git repository.
While they must be included in python packages, they do not need to be in the git repository. They make the repository bigger, and as git retrieves all the commit when cloning, with enough time this can become noticeable. They can also induce unnecessary conflicts.

Catalog can be compiled just in time when tests are launched (via a pytest fixture [for example](https://github.com/yaal-coop/canaille/blob/main/tests/conftest.py#L14-L21)) or when the python packages are built (via setuptools/hatch/poetry/etc. hooks).

What do you think? If this OK with you I can submit a PR.

zorun:
I agree it's not ideal, but it's intended. Several years ago we converged on this behaviour, but I couldn't find the discussion. I believe it was because we sometimes forgot to rebuild translations for releases, and maybe there was an issue with weblate.

I'm completely open to revisit this, your help would be appreciated! Things to keep in mind:

- pypi release (semi-automated, see contributing doc)
- docker publication (automated in CI)
- local development
- tests (but I think we don't use translations in tests)
- possible interactions with weblate?

almet:
Hi @azmeuk, and thanks for your feedback on this.

I'm also open to revisit this, as the current behavior sometimes leads to malfunctions. Don't hesitate to submit a pull request, I'm +1 for it.

azmeuk:
As I wrote in #1243 it might be better to move away from setuptools someday, and replacement tools would allow us to run some custom code at build time (such as catalog compilation). I have not seen this with setuptools+setup.cfg but I think we could do this with setuptools+setup.py

Anyways, I know that at least [hatch](https://github.com/pypa/hatch) and [poetry](https://python-poetry.org/) would allow this. We [moved to hatch in wtforms recently](https://github.com/wtforms/wtforms/pull/798) and [catalog compilation at build time works](https://github.com/wtforms/wtforms/blob/master/hatch_build.py), and there is no issue with weblate so far. As tox builds the packages, unit tests run with tox will have compiled catalogs. For people who would like to use pytest outside a tox environment, we can write an auto-used pytest fixture as [we did in canaille](https://github.com/yaal-coop/canaille/blob/main/tests/conftest.py#L14-L21).

I would suggest if/when #1243 is merged to move to hatch. What do you think?

almet:
[...]

PR #1242, opened 2023-10-07
fix: password generation command line crash

azmeuk:
Fix a password generation command line crash introduced by #1214 

fixes #1241

PR #1243, opened 2023-10-07
Migrate from setup.cfg to pyproject.toml

azmeuk:
I was having a look at #1238 and I realized that setuptools+setup.cfg is not really convenient for build callbacks (such as: catalog compilation).
I was thinking about using hatch instead of setuptools to build packages, like [we did recently in wtforms](https://github.com/wtforms/wtforms/pull/798), as it is lightweight and provides build callbacks, and setuptools is deprecated anyways (since it is built on distutils, that have [just been deleted in python 3.12](https://docs.python.org/3/whatsnew/3.12.html?highlight=distutils)).

Anyways, if we use hatch in the future (or flit, or poetry, or whatever), then pyproject.toml is required, and if we keep using setuptools, pyproject.toml won't harm :)

We can have the discussion about the build tool on #1238


setup.cfg
```diff
@@ -1,82 +0,0 @@
...
-
-[options.entry_points]
-flask.commands =
-    generate_password_hash = ihatemoney.manage:password_hash
-    generate-config = ihatemoney.manage:generate_config
-
-console_scripts =
-    ihatemoney = ihatemoney.manage:cli
-
-paste.app_factory =
```
azmeuk, line 81: Is [paste deployment](https://docs.pylonsproject.org/projects/pastedeploy/en/latest/) still used?

pyproject.toml
```diff
@@ -0,0 +1,88 @@
...
+]
+
+[project.urls]
+"Source Code" = "https://github.com/spiral-project/ihatemoney"
+"Main instance" = "https://ihatemoney.org/"
+"Mobile app" = "https://ihatemoney.org/mobile"
+"Documentation" = "https://ihatemoney.readthedocs.io/en/latest/"
+
+[project.scripts]
+ihatemoney = "ihatemoney.manage:cli"
```
azmeuk, line 85: This makes the `ihatemoney` command available.

ihatemoney/manage.py
```diff
@@ -32,14 +32,14 @@ def runserver(ctx):
     ctx.forward(run)
 
 
-@click.command(name="generate_password_hash")
+@cli.command(name="generate_password_hash")
```
azmeuk, line 35: This was needed to provide `ihatemoney generate_password_hash`

pyproject.toml
```diff
@@ -0,0 +1,88 @@
...
+    "Sphinx>=7.0.1,<8",
+    "docutils==0.20.1",
+    "myst-parser>=2,<3",
+]
+
+[project.urls]
+"Source Code" = "https://github.com/spiral-project/ihatemoney"
+"Main instance" = "https://ihatemoney.org/"
+"Mobile app" = "https://ihatemoney.org/mobile"
+"Documentation" = "https://ihatemoney.readthedocs.io/en/latest/"
```
azmeuk, line 82: `[project.urls]` allows for more urls to be displayed on the PyPI page.

setup.cfg
almet: I've checked and it seems that all the documented deployment options and configuration files are relying on the wsgi module, so we probably don't need to maintain this.

zorun:
Thanks for the work @azmeuk 

@almet Could you please let time for other people to review changes before merging them extra-fast? In that case, I would have like to:

- first add more tests to catch the silent password hash generation crash
- then merge #1242 fixing the crash
- then release a bugfix release 6.1.2
- then merge more impacting changes that have a risk to introduce regression. Migrating to pyproject.toml does have such a risk (small but anyway)

almet:
@zorun I'm sorry I didn't let you more time to review this. I understand what you mean and why it could have been simpler to merge it later on.

I believe these scenarii will probably happen more often than before, as more people are working on the repository than usual. We might want to setup some rules for merging, releasing, etc.

I see different ways to solve this, I've open #1245 to discuss that.

PR #1248, opened 2023-10-11
support for wtforms 3.1

azmeuk:
supersedes #1247


almet (approved):
Thanks. I had to find and read https://github.com/wtforms/wtforms/pull/739/files to understand why this was needed.

I'm okay with the changes :-)

azmeuk:
I will add this to the wtforms changelog so it is clearer.

PR #1250, opened 2023-10-19
invite email translation now happens on the fly

### LINKED ISSUE #869: Allow translation of invitation emails

afriedman412:
Fixes #869 

Invite emails were hard coded and could not be translated using the existing code base. This change runs the (already rendered) email body through a Jinja `{% trans %}` tag.

There's a new function to do this (`utils. translate_template_text`) and it can be generalized to other emails if need be!


.gitignore
```diff
@@ -18,3 +18,5 @@ ihatemoney/budget.db
 .python-version
 .coverage*
 prof
+*.ipynb
+*/.ipynb_checkpoints/
```
almet, line 22: I believe these should be better in your global `~/.gitignore` file maybe?

ihatemoney/templates/invitation_mail.j2
```diff
@@ -1,3 +1,5 @@
+
```
almet, line 1: I don't believe we need the extra spaces here. Any reason you added them?

almet (changes requested):
I don't understand why we need a level of indirection here. Adding `{% trans %}` blocks in the templates should be working.

If it's not, we need to figure out why :-)

almet:
Anyway, thanks for taking a look at this =D

ihatemoney/templates/invitation_mail.j2
afriedman412: fixed

.gitignore
afriedman412: didnt know that was possible!  fixed

afriedman412:
> I don't understand why we need a level of indirection here. Adding `{% trans %}` blocks in the templates should be working.
> 
> If it's not, we need to figure out why :-)

aight well you know what they say about the perfect and the good

ihatemoney/templates/translation_template.j2
```diff
@@ -0,0 +1,3 @@
+{% trans %}
+{{ untranslated_text }}
+{% endtrans %}
```
almet, line 3: ditto

almet:
> aight well you know what they say about the perfect and the good

Have you tried to understand what's going on here? If it's not working then we need to understand why and potentially open a bug upstream :)

afriedman412:
> > aight well you know what they say about the perfect and the good
> 
> Have you tried to understand what's going on here? If it's not working then we need to understand why and potentially open a bug upstream :)

iuno man you wanted to "Allow translation of invitation emails"

this allows translation of invitation emails (and any other emails you don't want to have to translate manually into other languages)

take it or leave it

zorun:
> take it or leave it

Thanks for the proposed change, but that's not the way we work here. And in general I believe it's not the right approach to contribute to open source projects.

About the change itself: have you verified it actually works?  That is: 1) that you were able to extract the text to translate into the translation catalog, and 2) that a translation of this text in the catalog correctly shows up. It seems to me that the indirection will prevent the extraction from working.

almet:
> iuno man you wanted to "Allow translation of invitation emails"
> 
> this allows translation of invitation emails (and any other emails you don't want to have to translate manually into other languages)
> 
> take it or leave it

Hi @afriedman412. I'm sorry to see your comments and understand your frustration. I can only hope you understand ours. As @zorun mentioned, the "Take it or leave it" is not how we chose to work on this project. We are open to the conversation, but as maintainers have the final say into what's merged or not, and code quality is one of our concerns.

So, let me try to be more precise in my feedback: I'm surprised we have to do an indirection here in the code, and I don't understand why we have to do it. It seems that something weird is happening, and might be related to how we configured Jinja, or how we call it. 

I also want to add that I'm only criticizing here the code you wrote, and not you. Again : thanks for your contribution here, and I hope you're well. Don't hesitate to ask if you need some help.

PS : Thanks @zorun for the clarifications and help here 👍

afriedman412:
> > iuno man you wanted to "Allow translation of invitation emails"
> > this allows translation of invitation emails (and any other emails you don't want to have to translate manually into other languages)
> > take it or leave it
> 
> Hi @afriedman412. I'm sorry to see your comments and understand your frustration. I can only hope you understand ours. As @zorun mentioned, the "Take it or leave it" is not how we chose to work on this project. We are open to the conversation, but as maintainers have the final say into what's merged or not, and code quality is one of our concerns.
> 
> So, let me try to be more precise in my feedback: I'm surprised we have to do an indirection here in the code, and I don't understand why we have to do it. It seems that something weird is happening, and might be related to how we configured Jinja, or how we call it.
> 
> I also want to add that I'm only criticizing here the code you wrote, and not you. Again : thanks for your contribution here, and I hope you're well. Don't hesitate to ask if you need some help.
> 
> PS : Thanks @zorun for the clarifications and help here 👍

hi -- thanks for revisiting this! it looks like this is going to take a bit more hands-on translation work than I expected, so I'll pass on it for now.

PR #1244, opened 2023-10-07
update to flask and werkzeug 2.3

azmeuk:
Everything looks fine, there was just a test to edit except for python 3.7

PR #1256, opened 2023-11-11
Prepare 6.2.0 release

zorun:
I would like to release 6.1.2, mostly because of the fix for `generate_password_hash`.

I will add an entry about #1244 once it's fixed.

According to our new [repository rules](https://ihatemoney.readthedocs.io/en/latest/contributing.html#repository-rules) it needs another maintainer approval or a delay of one week.


CHANGELOG.md
```diff
@@ -4,8 +4,10 @@ This document describes changes between each past release.
 
 ## 6.1.2 (unreleased)
```
almet, line 5: Moving from `setup.cfg` to `pyproject.toml` should probably be considered a breaking change. I agree with a previous comment of yours (where you told it would be better to not do things in this order and issue the quick fix first), but it's also probably too late now.

To me, it's at least a new feature (so updating the minor, to 6.2.0), but probably a breaking change, so we should update the major (7.0.0)

zorun:
I agree we are in an annoying situation, with both a "bugfix" and a "feature" release in sight.

I suggest an alternative route:

- let's plan a 6.2.0 release with existing packaging changes + python 3.12 support, and maybe also #1258 (ping @azmeuk , does it make sense to release all packaging changes together?)
- but in the meantime, let's release a backport 6.1.2 release with just the bugfix from #1242. We've done this in the past, we still have the `stable-4.1` branch as an example.

@almet what do you think of this plan? If you agree, I can do the backport release this week-end or next week.

azmeuk:
> let's plan a 6.2.0 release with existing packaging changes + python 3.12 support, and maybe also https://github.com/spiral-project/ihatemoney/pull/1258 (ping @azmeuk , does it make sense to release all packaging changes together?)

Maybe this worth waiting on #1258 as this is also a big change? I will try to move this forward in the weekend

> but in the meantime, let's release a backport 6.1.2 release with just the bugfix from https://github.com/spiral-project/ihatemoney/pull/1242. We've done this in the past, we still have the stable-4.1 branch as an example.

Sounds reasonable to me.

almet:
That sounds like a plan :-)

1. Release a backport release
2. Wait for #1258 and then issue a new minor release (according to semver, as it's rectrocompatible with the previous packaging tools, a minor is okay).

zorun:
Great :) I've just released 6.1.2 and backported changelog and CI changes in #1266

I'll update this merge request accordingly.

almet:
Merged as it's just a changelog entry.

PR #1257, opened 2023-11-11
Actually update to flask and werkzeug 2.3

almet:
The update was previously made by @azmeuk but
lost in a bad merge (see #1244 for details).


zorun:
Thanks @almet !

Note that this will get us a security fix for https://github.com/pallets/werkzeug/security/advisories/GHSA-hrfv-mqp8-q5rw so it's even better.

PR #1260, opened 2023-11-11
Swapping Out 'Fred' for 'Jeanne' in Docs/API – Personal Peacekeeping 😊

almet:
Swapping ‘Fred’ for ‘Jeanne’ across our docs and API.

Here’s the lowdown - 'Fred' and I ... we've got history. It's the kind of backstory that doesn't really gel with my day-to-day code jams. Every time I bump into 'Fred' in our project, it’s like a weird little jingle that gets stuck in my head – not the fun kind. So, I’m trading it out for 'Jeanne,' which is just a way chiller vibe for me.

But no worries, I’ve double-checked this swap, making sure it’s all smooth:
- All the 'Fred' references in the docs have waved goodbye, and 'Jeanne' has warmly taken their spot.
- If 'Fred' had a cameo in any comments, tests, or examples, 'Jeanne' is now nailing the audition instead.
- Gave the whole thing a once-over (then a twice-over) to catch any sneaky 'Fred' stragglers.

This isn’t about flipping tables – it’s all about keeping the code kitchen a place where we’re all comfy cooking up our projects. I’m sure you’ll find 'Jeanne' to be a pretty seamless switcheroo.

Ping me if things look off or you just wanna chat about it. Thanks for being the kind of crew that's cool with these little tweaks. It means a lot. 🌟

Issue #1259, opened 2023-11-11
Renaming a member via the Rest-API reset optional arguments (activated & weight)


Turtle6665:
Hello, 

When a member name is updated without filling the optional argument, it resets them to a weight of one and a deactivation. So in short, any optional argument not given will be reset.
For example, with this code:

```
#if we give the activated field & the weight along side the name
curl --basic -u demo:demo -X PUT https://ihatemoney.org/api/projects/demo/members/36807 -d 'name=yeaaaaah&activated=true&weight=2'
#> {"id": 36807, "name": "yeaaaaah", "weight": 2.0, "activated": true}

#if we only give the name
curl --basic -u demo:demo -X PUT https://ihatemoney.org/api/projects/demo/members/36807 -d 'name=yeaaaaah'
#> {"id": 36807, "name": "yeaaaaah", "weight": 1.0, "activated": false}
```

What I assumed by reading the [API documentation](https://ihatemoney.readthedocs.io/en/latest/api.html) is that those values are optional and, when not given, will not change their value. I therefore assume it's a bug, but it may be the expected behaviour ?

Have a nice day

almet:
Hi, thanks for opening this issue. I think you're correct : this is a bug. We should look in the database before erasing the values with the defaults.

https://github.com/spiral-project/ihatemoney/blob/master/ihatemoney/api/common.py#L100-L110

zorun:
It's a general issue with the API, similar to #1205

That being said, that's probably how a "REST-compliant API" is supposed to behave for PUT requests with optional arguments. PUT is supposed to completely replace the resource, so it would theoretically make sense to reset optional values to their default. But this is usually not what we want, especially when we add new optional arguments that some clients may not know about yet.

So, I believe the "clean" REST way should be to implement PATCH instead.

almet:
I agree with you on the semantics, but that's not how it's implemented now. I believe the problem in this issue and #1205 is a bug.

As I mentioned in #1261, WTForms considers that [no value for a Boolean field means it should be set to False](https://github.com/wtforms/wtforms/issues/402).

PR #1261, opened 2023-11-12
Fix API for member edition

almet:
Fix for #1259. To merge after #1260.

- [x] Add a test showcasing the problem
- [ ] Find and implement a fix

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -133,7 +133,7 @@ You can get all the members with a `GET` on
     [{"weight": 1, "activated": true, "id": 31, "name": "Arnaud"},
      {"weight": 1, "activated": true, "id": 32, "name": "Alexis"},
      {"weight": 1, "activated": true, "id": 33, "name": "Olivier"},
-     {"weight": 1, "activated": true, "id": 34, "name": "Fred"}]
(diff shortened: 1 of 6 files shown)


almet:
I can run the failing test with

> .venv/bin/pytest -W ignore::DeprecationWarning ihatemoney/tests/api_test.py -k test_member_edition_keep_data -s

almet:
From what I understand:

- When no data is provided, the form falls back to its default values, rather than using the already-present data
- There is an extra `form.fill(member)` method on the form, which job is to explicitely do this. [It's used in the web.py views](https://github.com/spiral-project/ihatemoney/blob/0fa1a0d4f80b70d85429885bce3fbc699fb5da0f/ihatemoney/web.py#L757:L757)

almet:
This is actually manyfold.

- The solution for the weight was to pass the obj to the Form,
- For the activation, it was trickier: WTForms considers that [no value for a Boolean field means it should be set to False](https://github.com/wtforms/wtforms/issues/402), so I've been looking at the raw data to see if its present before using the data.

almet:
Thanks for the work on this.

I'm doing some winter cleanup on the repo, and closing this as it had no activity since more than a year. Feel free to reopen if you feel like it!

PR #1270, opened 2023-11-20
Fix APPLICATION_ROOT example in docs

aristocrates:
If the application is hosted on the path `/somestring`, static asset paths need to have prefix `/somestring/static/` but setting `APPLICATION_ROOT` to "somestring" will result in a relative path which will give a 4XX. Using "/somestring" fixes this.

e.g. static assets should look like this in source:

```html
<link rel="stylesheet" type="text/css" href="/somestring/static/css/main.css">
```

but without the leading slash they have `href="somestring/static/css/main.css"` which results in a network request for `{DOMAIN}/somestring/{PROJECT_NAME}/somestring/static/css/main.css`

The test suite includes the leading slash in `test_prefixed`:

https://github.com/spiral-project/ihatemoney/blob/76e8b3baf0b52d098c1287f01ebf30cdeeb6ecc5/ihatemoney/tests/main_test.py#L76-L77

---

Also fix default value, which is "/" and not the empty string:

https://github.com/spiral-project/ihatemoney/blob/76e8b3baf0b52d098c1287f01ebf30cdeeb6ecc5/ihatemoney/default_settings.py#L13

https://github.com/spiral-project/ihatemoney/blob/76e8b3baf0b52d098c1287f01ebf30cdeeb6ecc5/Dockerfile#L29

https://github.com/spiral-project/ihatemoney/blob/76e8b3baf0b52d098c1287f01ebf30cdeeb6ecc5/docker-compose.yml#L38

https://github.com/spiral-project/ihatemoney/blob/76e8b3baf0b52d098c1287f01ebf30cdeeb6ecc5/ihatemoney/conf-templates/ihatemoney.cfg.j2#L50-L51

https://github.com/spiral-project/ihatemoney/blob/76e8b3baf0b52d098c1287f01ebf30cdeeb6ecc5/ihatemoney/tests/main_test.py#L72

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -127,7 +127,7 @@ ADMIN_PASSWORD needs to be set.
 
 ## APPLICATION_ROOT
 
-If empty, ihatemoney will be served at domain root (e.g:
+By default, ihatemoney will be served at domain root (e.g:
 *http://domain.tld*), if set to `"/somestring"`, it will be served from a
 "folder" (e.g: *http://domain.tld/somestring*).
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -128,10 +128,10 @@ ADMIN_PASSWORD needs to be set.
 ## APPLICATION_ROOT
 
 If empty, ihatemoney will be served at domain root (e.g:
-*http://domain.tld*), if set to `"somestring"`, it will be served from a
+*http://domain.tld*), if set to `"/somestring"`, it will be served from a
 "folder" (e.g: *http://domain.tld/somestring*).
 
--   **Default value:** `""` (empty string)
+-   **Default value:** `"/"`
 
 ## BABEL_DEFAULT_TIMEZONE
 


docs/configuration.md
```diff
@@ -128,10 +128,10 @@ ADMIN_PASSWORD needs to be set.
 ## APPLICATION_ROOT
 
 If empty, ihatemoney will be served at domain root (e.g:
```
aristocrates, line 130: Perhaps this should be changed to "by default" since the default value is "/"

```suggestion
By default, ihatemoney will be served at domain root (e.g:
```

almet:
Thanks, that looks good to me.

almet (approved):

aristocrates:
Thanks, I also added the `s/If empty/By default/` change

zorun:
Thanks for catching the doc issue.

From what I remember from last time I looked at this, the default value from Flask was an empty string. We merely force this to be `/` in all our example configs, see 3788b6f5e75a1077f662edd660650877e9cd52e6 . But it seems I remember wrong: https://flask.palletsprojects.com/en/2.3.x/config/#APPLICATION_ROOT

By the way, this is all a hack, normally we are supposed to get this variable from the web server / runtime environment, and not define it ourselves. But anyway :)

In any case, this looks good to me!

PR #1271, opened 2023-11-20
Ajout d'une BD explicative en anglais.

TheodoreCouz:
--FR--
J'ai créé et ajouté une BD explicative du fonctionnement du site en version anglaise. Cette BD est basée sur la BD orginiale disponible sur le site (uniquement en version française).
J'ai modifié le code HTML pour que la BD en version anglaise soit disponible par défaut lorsque la langue selectionnée est différente du français.

--EN--
I've created and added a comic strip explaining how the site works in English. This comic is based on the original comic available on the site (French version only).
I've modified the HTML code so that the English version is available by default when the selected language is different from French.


almet:
Hi @TheodoreCouz, thanks for doing this, it's appreciated. That being said, it seems that you submited way too much files in this pull request (I'm not sure where they're coming from?).

If you need some help to set things up, don't hesitate to ask here for help :-)

TheodoreCouz:
It seems to be coming from Python libraries. I will resolve the issue soon. Thank you for bringing it to my attention!

TheodoreCouz:
Update: the issue with the added files is fixed.

almet:
Hi, thanks for your work on this. I've reworked it and added changes in #1273.

Some of the pictures are a big cropped, which is not ideal, but it works.
I'm closing this in favor of the other pull request. Don't hesitate to discuss there.

PR #1273, opened 2023-11-23
English showcase

almet:
Some work to integrate what @ThedoreCouz has done in #1271


zorun:
Hey, thanks for your work @TheodoreCouz !

Here is a review on the current version (not sure what comes from you or from @almet ):

- I'm not sure it makes sense to have a configuration variable for this. What happens if people add "sv" to this variable? There's no images in Swedish so it won't work.
- I would suggest always displaying the English version when we don't have images in the current language/ That would even encourage people to translate the comics (but of course if we end up with 1 MB of images for each language, then it becomes a packaging issue)
- minor nitpick: there are French labels such as "Fermer", "Suivant", "Précédent"
- before merging, don't forget to squash the huge commits with the mistakenly commited libs

almet:
Thanks for the feedback.

> I'm not sure it makes sense to have a configuration variable for this. What happens if people add "sv" to this variable? There's no images in Swedish so it won't work.

The setting was actually intended, so that if your language is in there, then the button is displayed, and fallsback to english if your language isn't there.

But we could also always display it.

> minor nitpick: there are French labels such as "Fermer", "Suivant", "Précédent"

Ah, I missed these ones. I'll add them to the translation.

> before merging, don't forget to squash the huge commits with the mistakenly commited libs

:-)

PR #1275, opened 2023-11-23
Update security support status

zorun:
I just discovered this file, and it's out-of-date

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -4,9 +4,9 @@
 
 | Version | Supported          |
 | ------- | ------------------ |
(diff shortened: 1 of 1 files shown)


almet:
Thanks. I believe we should either be on the liberachat IRC, or ask people to send us emails instead though. What do you think?

zorun:
Agreed about an email address. Is there email available on the ihatemoney.org domain? Can you create a security@ email alias?

almet:
> Can you create a security@ email alias?

Done :-)

PR #1276, opened 2023-12-04
translation blocks added to invitation emails

stuinman:
I added the translation blocks that were discussed in earlier posts for https://github.com/spiral-project/ihatemoney/issues/869. I referred to this site for my changes:
https://svn.python.org/projects/external/Jinja-1.1/docs/build/templatei18n.html

Let me know if there is anything more I can do/fix!


almet:
Thanks! I'm actually unsure this is the whole story. Behind this we need to call `make update-translations` and when I do it (after rebasing #1279) I don't see the translated strings in the regenerated messages.

Have you tried it works as intended?

PR #1278, opened 2023-12-07
Update installation.md

rabelosr20:
Updated installation.md to include instructions on how to setup a virtual environment in Windows


almet:
Hi, thanks for your additions.

Unfortunately, we don't support windows installation per se. The recommended way to install on windows is to use WSL https://learn.microsoft.com/en-us/windows/wsl/install

I'm closing this, don't hesitate to reopen if you want to include some information about WSL instead.

PR #1288, opened 2024-02-04
 Fixing the issue of Chinese character garbled encoding in CSV export.

qqAys:
**Fixing the issue of garbled characters (Chinese characters) when opening the CSV export in Microsoft Excel.**
Using version 6.1.3 and opening CSV export files with Chinese content by default in Microsoft Excel in a Chinese environment results in garbled characters due to encoding issues, as shown in the image (the top part is the current version export, and the bottom part is the export after fixing).
Snipaste_2024-02-04_14-31-15

### DIFF
--- a/ihatemoney/tests/import_test.py
+++ b/ihatemoney/tests/import_test.py
@@ -423,7 +423,7 @@ class TestExport(IhatemoneyTestCase):
             '2016-12-31,red wine,XXX,200.0,jeanne,1.0,"zorglub, tata"',
             '2016-12-31,fromage à raclette,10.0,XXX,zorglub,2.0,"zorglub, jeanne, tata, pépé"',
         ]
-        received_lines = resp.data.decode("utf-8").split("\n")
+        received_lines = resp.data.decode("utf-8-sig").split("\n")
 
         for i, line in enumerate(expected):
             assert set(line.split(",")) == set(received_lines[i].strip("\r").split(","))
@@ -457,7 +457,7 @@ class TestExport(IhatemoneyTestCase):
             "55.34,XXX,jeanne,tata",
             "127.33,XXX,jeanne,zorglub",
         ]
-        received_lines = resp.data.decode("utf-8").split("\n")
+        received_lines = resp.data.decode("utf-8-sig").split("\n")
 
         for i, line in enumerate(expected):
             assert set(line.split(",")) == set(received_lines[i].strip("\r").split(","))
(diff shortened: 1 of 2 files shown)


almet:
Thanks. This seems to be failing the tests though… We might need to change the way the .csv files are loaded.

qqAys:
Happy Chinese New Year! Thank you for bringing this to my attention. I've made the necessary adjustments to address the test failures. Regarding the loading of the .csv files, I've implemented some changes to ensure proper handling. Please let me know if you encounter any further issues or if there's anything else I can assist with.

zorun:
utf-8-sig seems to add a BOM sequence at the start of the CSV file.

Is that standard, and can that break other clients? It feels like Excel is the problem here, it should correctly handle UTF-8 encoded files.

qqAys:
> utf-8-sig seems to add a BOM sequence at the start of the CSV file.
> 
> Is that standard, and can that break other clients? It feels like Excel is the problem here, it should correctly handle UTF-8 encoded files.

Thanks for your comment! Indeed, I found that changing the encoding of the CSV file from UTF-8 to UTF-8-SIG can solve the problem of garbled Chinese characters. UTF-8-SIG adds a special identifier at the beginning of the file, helping software to correctly parse the file and avoid garbled characters.

Regarding the issue you mentioned about other clients possibly experiencing problems, it depends mainly on how those clients are implemented. Most modern software can handle UTF-8 files with a BOM correctly, but older versions or software with specific settings might struggle with the BOM. Therefore, using UTF-8-SIG requires consideration of compatibility with different clients.

As for whether Excel should be able to handle UTF-8 encoded files correctly, that is indeed an important question. Modern versions of Excel can handle UTF-8 encoded files well, but there may still be some issues. This might require further improvement and optimization from software vendors.

In conclusion, using UTF-8-SIG is an effective solution, but it's important to balance and adjust based on the actual situation to ensure the best compatibility and user experience. Thanks again for your comment and suggestion!

almet:
> In practice, you can generate either RFC 4180 compliant CSV files, or you can generate Microsoft compatible CSV files.
> 
> — [Stack Overflow](https://softwareengineering.stackexchange.com/questions/372692/should-utf-8-csv-files-contain-a-bom-byte-order-mark)

I'd rather not add a BOM for all our CSV files. If we want to be able to generate Excel-Compatible CSV files, then probably we should offer the option to the user to chose between the two file formats.

To me, adding a BOM to all the generated `.csv` files looks like it could break other readers following RFC 4180 to break, which probably we don't want either.

almet:
Closing this issue for now as I believe we don't want to merge it as-is. As I offered, feel free to reopen with a different approach proposing multiple download formats.

Also: thanks for the delay on all this, the project has been unmaintained for the last year or so, and we're getting back at it :-)

PR #1116, opened 2022-12-13
Bill Types added to Bill , pay feature added for Settle page, Statistics Page adjusted to be compatible with new types

FlowingCloudRTL:
# Description
To tackle #137, I took some ideas from the discussion under the issue and made some design changes that are different from the previous PR that was done: 
- Bill Types are added in the Bill model and forms to differentiate refund/reimbursement from expenses of the project. 
   Currently, only three types of bills are hardcoded in as the present for all projects: Expense, Reimbursement, and Transfer. Expense is the normal spending, Reimbursement bill serves to settle the debt within the project, and Transfer is any bills between members that should not be tracked as an expense. 
image
image  
- Settle button is added to directly add a Reimbursement bill to clear out the debt associated with another participant with one click rather than going through another form for better UX. 
image
- To maintain the Statistics page's functionality, Monthly Expense would only track expense type bills, and when participant A reimburses participant B, A's paid amount would increase and B's paid amount would decrease by the reimbursed amount to maintain the relationship between paid, expense, and balance. (balances = Paid - Spent)
image

Fixes #137

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [✓] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?
Most tests associated with the Bill model are modified to be compatible with the new bill type attribute. New tests and validations are added for bill type, the pay button on the settle page, and statistics calculations. 
All tests pass except the broken CI/CD in history_test.py 

**Test Configuration**:
* Firmware version: Windows 10
* SDK: Python 3.8.10

# Checklist:

- [✓] My code follows the style guidelines of this project
- [✓] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [✓] My changes generate no new warnings
- [✓] I have added tests that prove my fix is effective or that my feature works
- [✓] New and existing unit tests pass locally with my changes
- [✓] Any dependent changes have been merged and published in downstream modules

To-Do:
Documentation changes, inline documentation (ideally will be done in the next few days) 
Allow custom types to be added and removed in the project setting
Add more statistics page for different bill types

PR #1290, opened 2024-03-02
Adding bill types and automatic settling between people

### LINKED ISSUE #137: Add the ability to pay from the "settle" page

TomRoussel:
# Summary
This PR is a continuation of #1116. The goal of this PR is to finish the work of the original PR and fix #137. The overall design changes proposed in #1116 are mostly kept, though I'll detail the changes below. This PR adds a button to the "Settle" page to automatically let users signify they have settled their debt with other people. This generates a new bill in the overview page that will bring their debt with that person to 0. 

Normally this would pollute the statistics of that project, so a bill_type column is added to the database. There are 2 types, "Reimbursement" and "Expense", only the latter is used to calculate the total cost of the project. 

# Changes since the original PR

 - Brought the FlowingCloudRTL branch up to speed with the current master
 - Worked in the minor suggested changes by @Glandos 
 - Changed the "bill_type" column to use an Enum (as requested by @Glandos). This enum is also used by the bill form validator
 - Removed the "Transfer" bill type. Personally I don't see a use for it and it seems confusing to the user. In the original PR, there didn't seem to be any interest in keeping it around
 - Added an api test to check if bill_type is validated correctly

# Open questions

With these changes, all bill creation requests to the API require that a "bill_type" is provided. This will break compatibility with apps using the API. Should we make the "bill_type" field an optional parameter? I think we can safely make the default type be "Expense".


almet:
Hey, thanks for your work on this.

> With these changes, all bill creation requests to the API require that a "bill_type" is provided. This will break compatibility with apps using the API. Should we make the "bill_type" field an optional parameter? I think we can safely make the default type be "Expense".

+1 for this being the default behaviour.

almet (commented):
Hey, thanks for this.

I wonder how [cospend](https://github.com/julien-nc/cospend-nc/) does it, we probably should have a look at their API and try to mimic it, so we can stay compatible.

ihatemoney/migrations/versions/7a9b38559992_new_bill_type_attribute_added.py
```diff
@@ -0,0 +1,24 @@
...
+down_revision = "927ed575acbd"
+
+from alembic import op
+import sqlalchemy as sa
+from ihatemoney.models import BillType
+
+
+def upgrade():
+    op.add_column("bill", sa.Column("bill_type", sa.Enum(BillType)))
+    op.add_column("bill_version", sa.Column("bill_type", sa.UnicodeText()))
```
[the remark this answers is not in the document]
TomRoussel: Right! I've added "expense" as the default type here.

ihatemoney/migrations/versions/7a9b38559992_new_bill_type_attribute_added.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def downgrade():
+    pass
```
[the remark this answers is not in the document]
TomRoussel: Of course, to be honest I'm not familiar with alembic, so I kind of glanced over this :) I've added a downgrade rule now

ihatemoney/tests/api_test.py
```diff
@@ -990,8 +1007,44 @@ def test_amount_too_high(self):
...
+                "what": "fromage",
+                "payer": "1",
+                "payed_for": ["1"],
+                "bill_type": "wrong_bill_type",
+                "amount": "50",
+            },
+            headers=self.get_auth("raclette")
+        )
+
+        self.assertStatus(400, req)
```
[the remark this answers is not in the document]
TomRoussel: I don't think this needs changing. I'm testing here if an unknown bill type is being rejected by the form validation code. 

I'll add another test that checks if no bill_type is passed to the api request, the bill is still accepted as an expense.

TomRoussel:
Okay, I'll make bill_type an optional parameter to the bill adding API then and get back to you when that's ready!

TomRoussel:
Done, bill_type is now "Expense" by default, and this default is now applied to all existing bills as well.

ihatemoney/templates/list_bills.html
```diff
@@ -117,6 +117,7 @@ <h3 class="modal-title">{{ _('Add a bill') }}</h3>
         <table id="bill_table" class="col table table-striped table-hover table-responsive-sm">
             <thead>
                 <tr><th>{{ _("When?") }}
+                </th><th>{{ _("Bill Type") }}
```
almet, line 120: Rather than adding a column, what about adding an info next to the name, or just use the "for what?" field?

TomRoussel: I actually just removed the bill type column now. The fact that it's a settlement is indeed clear from the "for what" field. Now it looks like this:
image

ihatemoney/web.py
```diff
@@ -848,6 +850,32 @@ def settle_bill():
...
 
 
+@main.route("/<project_id>/settle/<amount>/<int:ower_id>/<int:payer_id>")
+def settle(amount, ower_id, payer_id):
+    # FIXME: Test this bill belongs to this project !
+    # form = get_billform_for(g.project, set_default=False)
+    # form.bill_type = ("Refund", "Refund")
+    # form.amount = amount
+    # form.payer = ower
+    # form.payedfor = payer
```
almet, line 860: Are these comments still useful? 🤔

ihatemoney/web.py
```diff
@@ -848,6 +850,32 @@ def settle_bill():
...
+
+    new_reinbursement = Bill(
+        amount=float(amount),
+        date=datetime.datetime.today(),
+        owers=[Person.query.get(payer_id)],
+        payer_id=ower_id,
+        project_default_currency=g.project.default_currency,
+        bill_type="Reimbursement",
```
almet, line 868: Maybe use `BillType.REIMBURSEMENT` here

ihatemoney/web.py
```diff
@@ -848,6 +850,32 @@ def settle_bill():
...
+        what="Settlement",
```
almet, line 869: We might want this to be translated?

almet:
Thanks for the follow up on this. On my end, it seems good. I took another glance at your changes and added a few remarks. We're almost there!

ihatemoney/web.py
TomRoussel: I agree! If I understand correctly, to make this string translatable I just need to wrap it with `babel.gettext`, or `_`?

ihatemoney/web.py
TomRoussel: They don't seem useful, I'll remove them!

ihatemoney/web.py
TomRoussel: Alright, done!

almet:
Tests are currently failing on postgres, because it tries to use a unknown type:

```
psycopg2.errors.UndefinedObject: type "billtype" does not exist
E       LINE 1: ALTER TABLE bill ADD COLUMN bill_type billtype DEFAULT 'EXPE...
```

(The tests run otherwise properly when using sqlite)

TomRoussel:
Oh weird, I'll have a look this weekend to see why it's doing that.

TomRoussel:
Turns out alembic does not handle the generation of new enums correctly for postgres databases. [See this issue](https://github.com/sqlalchemy/alembic/issues/278). 

I fixed it by manually creating the sqlalchemy enum in the upgrade step of the migration, as someone suggested in the issue I linked. In that issue someone also introduced a library that handles enum migrations in alembic automatically: [alemibc-postgresql-enum](https://pypi.org/project/alembic-postgresql-enum/]), but I didn't want to add a new dependency for this small issue :).

Hopefully the tests are green now! It seemed to work on a postgres database I spun up locally.

almet:
Perfect, thanks :-)

almet:
Thanks for the work on this 👍

almet:
It was a pleasure to work with you :-)

TomRoussel:
Same here!

ihatemoney/tests/budget_test.py
```diff
@@ -706,10 +712,23 @@ def test_manage_bills(self):
...
+            "/raclette/add",
+            data={
+                "date": "2011-08-10",
+                "what": "Transfer",
+                "payer": members_ids[1],
+                "payed_for": members_ids[0],
+                "bill_type": "Transfer",
+                "amount": "500",
+            },
+        )
```
zorun, line 731: This looks like a leftover of the "Transfer" type

ihatemoney/tests/budget_test.py
```diff
@@ -764,12 +786,179 @@ def test_manage_bills(self):
...
+        # test transfer bill (should not affect anything whatsoever)
+        self.client.post(
+            "/random/add",
+            data={
+                "date": "2022-10-10",
+                "what": "Transfer of 500 to fred",
+                "payer": members_ids[0], #zorglub
+                "payed_for": members_ids[1], #fred  
+                "bill_type": "Transfer",
```
zorun, line 872: Same here

ihatemoney/web.py
```diff
@@ -848,6 +850,27 @@ def settle_bill():
     return render_template("settle_bills.html", bills=bills, current_view="settle_bill")
 
 
+@main.route("/<project_id>/settle/<amount>/<int:ower_id>/<int:payer_id>")
+def settle(amount, ower_id, payer_id):
+    # FIXME: Test this bill belongs to this project !
```
zorun, line 855: This looks like it should really be fixed! (and a test added for good measure)

ihatemoney/tests/import_test.py
```diff
@@ -536,7 +560,8 @@ def test_export_with_currencies(self):
             },
             {
                 "date": "2016-12-31",
-                "what": "fromage \xe0 raclette",
+                "what": "\xe0 raclette",
+                "bill_type": "Expense",
```
zorun, line 564: I take it you don't like fromage? ;)

ihatemoney/web.py
```diff
@@ -848,6 +850,27 @@ def settle_bill():
     return render_template("settle_bills.html", bills=bills, current_view="settle_bill")
 
 
+@main.route("/<project_id>/settle/<amount>/<int:ower_id>/<int:payer_id>")
```
zorun, line 853: As far as I can tell, there is no test for this new route. We should have one.

zorun (commented):
Thanks for finishing this long-awaited feature!

I'm a bit late to the party, I have added some comments, they can be addressed in a new pull request.

By the way, black is also complaining about formatting (see the CI output)

TomRoussel:
Sure, I'll go through those comments and submit a new PR! I'll probably have time to have a look in the weekend, though no promises.

You just use python-black on the whole codebase? Then I'll make sure to run it along with the suggested changes.

PR #1292, opened 2024-03-12
Changing any settings is prevented when project has existing currency

### LINKED ISSUE #1291: Changing any settings is prevented when project has existing currency
When a project as a default currency different from "XXX", changing the project settings will not be possible. An error is raised with the message : "Cannot change project currency because currency conversion is broken". The fact is that even if we don't want to change the currency, the message still shows up (same for the API). 

This comes from this if statement : 
https://github.com/spiral-project/ihatemoney/blob/2bb535070aae0f708856f727156c89ce857b85bb/ihatemoney/forms.py#L196-L204

For me, a better if statement could be (although I didn't test it): 
```{python}
if ( 
     project is not None 
     and field.data != project.default_currency
     and project.has_bills() 
 ):
```

I know this was a hot fix waiting for improvement with the #1232 but checking for the current project default currency makes more sens considering the error message. This bug is annoying as to change the other settings, we have to set the currency to "no currency" as it's the only condition that doesn't raise this error.

Turtle6665:
This PR fixes #1291. 

# Summary of the issue
It's a small fix to a small annoying bug : when a project as some bills, a default currency (different than "XXX") and you want to change the project settings an error message is shown : "Cannot change project currency because currency conversion is broken". The only way to change the project settings is then to remove the currency (set to "XXX"). Therefore, the error message is not constant with the behavior.

# The new behavior
The default currency field is check against the current project default currency. If the project has bills and the new input default currency doesn't match the current one, the error "Cannot change project currency because currency conversion is broken" is shown. 
I have tested the behavior. This is true for the API as well as the GUI input.

### DIFF
--- a/ihatemoney/forms.py
+++ b/ihatemoney/forms.py
@@ -195,7 +195,7 @@ class EditProjectForm(FlaskForm):
             raise ValidationError(msg)
         if (
             project is not None
-            and field.data != CurrencyConverter.no_currency
+            and field.data != project.default_currency
             and project.has_bills()
         ):
             msg = _(


Turtle6665:
One of the automatic test has failed (Test & Docs / build (3.12, normal, sqlite)). After reviewing the logs, I'm having no clue to which action should I do. There seams to be looking for a non-existing module: 'distutils'. This doesn't seam to be related to my change though.

Turtle6665:
Well, after a quick search on internet, I found out that `distutils` package is apparently removed in python version 3.12 (thanks to this [Stack Overflow answer](https://stackoverflow.com/a/76691103)).

It cites the [python 3.12 change log](https://docs.python.org/3.12/whatsnew/3.12.html) :
 > [gh-95299](https://github.com/python/cpython/issues/95299): Do not pre-install setuptools in virtual environments created with [venv](https://docs.python.org/3.12/library/venv.html#module-venv). This means that distutils, setuptools, pkg_resources, and easy_install will no longer available by default; to access these run `pip install setuptools` in the [activated](https://docs.python.org/3.12/library/venv.html#venv-explanation) virtual environment.

So a `pip install setuptools` should probably be added to the GitHub action for the python 3.12. Would this be better done on a different PR as it's a different issue @almet ? BTW, I'm not confident enough to successfully adapt the GitHub action, so I will let that to someone else to do.

almet:
Hey, thanks for the research on this. I find this particularly weird, because this seem already covered by #1258.

Including distutils in the dependencies might be the answer for the interim, even if I'm not sure why this is failing atm (with the fix already present in the released babel package)

zorun:
I just released 6.1.5 with this fix

PR #1294, opened 2024-03-23
fix: 'Bill Type: Invalid Choice: could not coerce' error

### LINKED ISSUE #1293: Bill Type: Invalid Choice: could not coerce.
After #1290 I receive "Bill Type: Invalid Choice: could not coerce" when attempting to add new bills. Changing the bill type doesn't help.

The request that is going out to `POST /<project>/add`:

```
csrf_token=...
date=2024-03-22
bill_type=BillType.EXPENSE
what=test
payer=10
amount=1
payed_for=1
payed_for=2
submit=Submit
```

I believe the issue could be in how the BillType enum is coerced by WTForms, see https://stackoverflow.com/a/49376245. BillType doesn't have a `coerce` classmethod: https://github.com/spiral-project/ihatemoney/blob/4af4c10b1fe7781207a7634823e200df40146935/ihatemoney/models.py#L54-L60.

I'm running the ihatemoney/ihatemoney:master docker image. Here are my pip dependencies:

```
/ # pip3 list
Package              Version      Editable project location
-------------------- ------------ -------------------------
alembic              1.13.1
aniso8601            9.0.1
Babel                2.14.0
blinker              1.7.0
cachetools           4.2.4
certifi              2024.2.2
charset-normalizer   3.3.2
click                8.1.7
commonmark           0.9.1
debts                0.5
Deprecated           1.2.14
dnspython            2.6.1
email_validator      2.1.1
Flask                2.3.3
flask-babel          3.1.0
Flask-Cors           3.0.10
Flask-Limiter        2.9.2
Flask-Mail           0.9.1
Flask-Migrate        4.0.7
Flask-RESTful        0.3.10
Flask-SQLAlchemy     2.5.1
flask-talisman       1.1.0
Flask-WTF            1.2.1
greenlet             3.0.3
gunicorn             21.2.0
idna                 3.6
ihatemoney           6.2.0.dev0   /src
importlib_resources  6.3.2
itsdangerous         2.1.2
Jinja2               3.1.3
limits               3.10.1
Mako                 1.3.2
MarkupSafe           2.1.5
ordered-set          4.1.0
packaging            24.0
pip                  23.0.1
psycopg2-binary      2.9.9
Pygments             2.17.2
PyMySQL              1.0.3
pypng                0.20220715.0
python-dateutil      2.9.0.post0
pytz                 2024.1
qrcode               7.4.2
requests             2.31.0
rich                 12.6.0
setuptools           65.5.1
six                  1.16.0
SQLAlchemy           1.4.52
SQLAlchemy-Continuum 1.4.1
SQLAlchemy-Utils     0.41.1
typing_extensions    4.10.0
urllib3              2.2.1
Werkzeug             2.3.8
wheel                0.43.0
wrapt                1.16.0
WTForms              3.1.2
``` 

Screenshot from 2024-03-22 23-19-44

rriski:
Error introduced in #1290. Fixes #1293. WTForms needs to be bumped to >=2.3.2
as it includes a fix to `SelectField` which is required for this change to work.

See:
  - https://wtforms.readthedocs.io/en/3.1.x/changes/#version-2-3-2
  - https://github.com/wtforms/wtforms/pull/598


ihatemoney/models.py
```diff
@@ -51,14 +51,25 @@
     ],
 )
 
+
 class BillType(Enum):
```
rriski, line 55: There is a `FormEnum` helper here https://github.com/spiral-project/ihatemoney/blob/4af4c10b1fe7781207a7634823e200df40146935/ihatemoney/utils.py#L324-L344 but we can't use it since it expects int values for the extending class.

ihatemoney/models.py
```diff
@@ -51,14 +51,25 @@
...
+        return self.value
+
     @classmethod
     def choices(cls):
         return [(choice, choice.value) for choice in cls]
 
+    @classmethod
+    def coerce(cls, item):
+        try:
+            return item if isinstance(item, BillType) else BillType[item.upper()]
```
rriski, line 69: Need `item.upper()` because the API expects that `bill_type` is passed as the enum value (i.e. "Expense" or "Reimbursement") but `item` is "EXPENSE" or "REIMBURSEMENT" when its passed as string to `coerce`.

ihatemoney/models.py
TomRoussel: Indeed I also found this while adding the enum, but I wanted to have an Enum with string values to easily show a "nice" string to the user.

TomRoussel:
I've been playing around with the changes in this PR. To me it seems like `BillType.coerce` in unnecessary. We can just use the enum constructor for this. `BillType("Expense")` correctly returns `BillType.Expense`.

I don't really understand why, but I think the key change that fixes the error in the form is adding `BillType.__str__`?

TomRoussel:
I quickly pushed my change to explain what I mean: See [this commit](https://github.com/TomRoussel/ihatemoney/commit/22e00a77fd7122d46d08974326cff36b46561ed3).

When I run the server with this change, I also don't get the error message.

TomRoussel:
Okay, I figured out what's going on. The WTForms SelectField sends the different choices to the browser as strings ([documented here](https://wtforms.readthedocs.io/en/2.3.x/fields/#wtforms.fields.SelectField). These are specified by `BillType.choices()` as a list of (value, label) tuples. The value should be the actual enum instance, and label would then be what the user sees. Since WTForms encodes everything as a string, it also sends this value to the  browser as `str(value)`. For an ordinary enum, this becomes `"BillType.EXPENSE"` for example. When the browser then sends the post request, it uses that string, but that's not understood by the coerce function.

That's why adding an explicit `BillType.__str__` fixes the problem. It's actually the equivalent as modifying choices as follows:

```python
    @classmethod
    def choices(cls):
        return [(choice.value, choice.value) for choice in cls]
```

rriski:
> Okay, I figured out what's going on. The WTForms SelectField sends the different choices to the browser as strings ([documented here](https://wtforms.readthedocs.io/en/2.3.x/fields/#wtforms.fields.SelectField). These are specified by `BillType.choices()` as a list of (value, label) tuples. The value should be the actual enum instance, and label would then be what the user sees. Since WTForms encodes everything as a string, it also sends this value to the browser as `str(value)`. For an ordinary enum, this becomes `"BillType.EXPENSE"` for example. When the browser then sends the post request, it uses that string, but that's not understood by the coerce function.
> 
> That's why adding an explicit `BillType.__str__` fixes the problem. It's actually the equivalent as modifying choices as follows:
> 
> ```python
>     @classmethod
>     def choices(cls):
>         return [(choice.value, choice.value) for choice in cls]
> ```

Thanks, I have updated the PR to use: 

```python
@classmethod
    def choices(cls):
        return [(choice.value, choice.value) for choice in cls]
```

TomRoussel:
That looks good to me! Ah nice, you also caught the duplicate tests, I just saw those today.

zorun:
Thanks for the fix.

Our CI was failing, which means that unformatted code went through in the last PR. I have fixed the issue in #1296 and hopefully reorganized the CI to avoid this in the future.

It does mean that you have to rebase and possibly fix conflicts, sorry about that. But we should then get a clearer view of the CI.

zorun:
The testcases with the lower versions for dependency are failing with HTTP 400 errors. It means that the approach doesn't work with older version of some libraries (here, most probably wtforms)

Does your solution rely on a undocumented behaviour in wtforms? If so, it's preferable to find another solution that will be less fragile over time.

If there is really no other workable approach, you can try bumping the minimum version of wtforms.

TomRoussel:
Very strange, as far as I can tell, this solution is not using anything exotic from wtforms. All api calls are following the documented way of working for 2.3.x of wtforms, which should correspond to the minimal version. What I think is strange is that this test is breaking now, but the minimal version checks in #1290 were fine.

I can try replicating the problem when I have time later this week.

TomRoussel:
I found the culprit I think. I believe the problem is related to a bug fixed in a PR of WTForms [here](https://github.com/wtforms/wtforms/pull/598). [The changelog for 2.3.2](https://wtforms.readthedocs.io/en/3.1.x/changes/#version-2-3-2) mentions that they fixed a bug in the `SelectField.choices` field. Using WTForms 2.3.1 fails, but 2.3.2 succeeds. Reading the PR description, I'm really not sure why our implementation here doesn't work. The only reason I can think of is that their validation logic for 2.3.1 somehow breaks if the choices tuples are identical values?

Is see 2 solutions to this problem, either we do a minor version bump for WTForms from 2.3.1 --> 2.3.2, or we use the following workaround which also seems to do the trick:
```python
class BillType(Enum):
    EXPENSE = "Expense"
    REIMBURSEMENT = "Reimbursement"

    def __str__(self):
        return self.value

    @classmethod
    def choices(cls):
        return [(choice, choice.value) for choice in cls]
```
Using a BillType instance as the first member of the choice tuple seems to not trigger the validation bug, while the __str__ method makes sure flasks sends a sensible value to the browser which can then be coerced back to the enum.

By the way, sorry @rriski that you got involved in this weird can of worms from my implementation!

zorun:
Thanks for looking at it!

If it's a bug in wtforms, then let's use the simplest solution: @rriski , can you bump the wtforms dependency to >=2.3.3 ?

rriski:
> Thanks for looking at it!
> 
> If it's a bug in wtforms, then let's use the simplest solution: @rriski , can you bump the wtforms dependency to >=2.3.3 ?

Sure, updated PR now. Thanks @TomRoussel for figuring out the issue.

Issue #1299, opened 2024-03-29
Find a consistent name for the new "reimbursement" / "settlement" / "transfer" bill type


zorun:
The new feature in #1290 uses a mix of "reimbursement" and "settlement" terminology. At one point in the development, it was called "transfer".

We should find a good and consistent name and stick to it. Keep in mind that there are two use-cases:

1) "I just gave 20 € to Bob to help cover my past and future expenses, and I am creating a bill to account for that transfer of money"
2) "I want to settle my debt to other members. So I look at the "Settle" page, I make one or more bank transfers of the right amount, and then I use the "Settle" buttons to automatically create the corresponding bills"

Regarding terminology, I'm not a native English speaker, but from what I understand:

- "reimbursement" / "reimburse" : used to cover for expenses after the fact, e.g. if I buy a sandwich in a business trip, my employer will reimburse me. I think this use-case does not really fit Ihatemoney since it's not a shared expense.
- "settlement" / "settle", e.g. "settle a debt". I owe money to somebody, and I give them back the right amount of money to settle this debt. This is definitely use-case 2
- "transfer": give money to somebody, i.e. decrease a debt I have towards this person or increase a debt this person has towards me. This is definitely use-case 1

I would go with "transfer" for the bill type, since this is what is visible when creating a bill, but keep "settlement" everywhere in the "Settle" page. In the statistics page, I'm thinking of adding two new columns "transferred" (total amount of transfers you made to somebody else) and "received" (total amount of transfers other people have made to you).

Any thought? @TomRoussel @almet

TomRoussel:
It's good to think about keeping our terminology simple and consistent!

I'm also not a native English speaker, so I can't really say if how I interpret the terminology makes sense or not. I interpret reimbursement as paying back what you owe to someone, so to me it sounds like use-case 1, but specifically to cover past expenses. The way you word it here, "transfer" does sound more general semantically, so that might be the better way to go. I'm more convinced of using transfer, as giving someone some money in advance to cover expenses later is a sensible usecase, and reimbursement is a weird way to phrase that.

In any case, I think it makes the most sense to keeping it to only 2 bill types until there is a convincing other use-case. The terminology in the Settle page is pretty clear I think, so I don't see a reason to change that. 

I don't think I would use those extra columns in the statistics page, but maybe others will. In general having some more stats is just fun, so I don't see a downside in adding it.

zorun:
[...]

PR #1300, opened 2024-03-29
Add transferred/received amounts in statistics page

zorun:
ihatemoney-stats-transfers

### DIFF
--- a/ihatemoney/models.py
+++ b/ihatemoney/models.py
@@ -113,42 +113,57 @@ class Project(db.Model):
 
     @property
     def full_balance(self):
-        """Returns a triple of dicts:
+        """Returns a tuple of dicts:
 
-        - dict mapping each member to its balance
+        - dict mapping each member to its overall balance
 
-        - dict mapping each member to how much he/she should pay others
-          (i.e. how much he/she benefited from bills)
+        - dict mapping each member to its expenses (i.e. how much he/she
+          benefited from all bills, whoever actually paid)
 
-        - dict mapping each member to how much he/she should be paid by
-          others (i.e. how much he/she has paid for bills)
+        - dict mapping each member to how much he/she has paid for bills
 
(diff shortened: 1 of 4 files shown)


zorun:
Names are temporary, see discussion in #1299

I have tried several orders for the column, this one seems clearer to me, but I'm intestered in feedback.

The negative signs are intended to show that all columns should sum to the balance.

zorun:
Some ideas for improving the rendering after showing this to somebody:

- there should be two clearly separated sets of columns:
  - **me and the group**: how much did I spend for the group, how much did the group spent for me
    - one column with "payment/contribution" (currently called "Paid" in my screenshot above)
    - one column with "expenses involving me" (currently called "Expenses" in my screenshot above)
  - **money transfers**: individual transfers of money between participants to restore the balance of the group (partially or completely)
    - "transfers"
    - "received transfers"
 - the two sets should be easy to understand, with good names and maybe an icon
 - if it's clear enough, there might not be any need for negative signs anywhere

almet:
Hi, thanks for the proposed changes. I'm not sure you need a review, or do you want to keep on adding things? :thinking:

It could be improved, but at the same time it feels like it could just already add some values to the stats page.

almet:
I've just rebased this on top of the latest `master` branch and updated the naming to follow our discussion in #1299

almet:
(I plan to merge if/when tests are green)

almet:
The tests should be passing now, **but** it seems that they change the way the balance is working.

Before these changes, a reimbursement to somebody would remove the amount from how much somebody has paid, and now it's not the case anymore.

I'm for merging this, as it seems to be the "good" behavior, but that might be a surprise for some clients.

Because it's changing the API we're exposing, we probably should issue a major release with the next release.

As this is changing some code you did @zorun, it would be great if you can review the changes I did.
Thanks!

PR #1302, opened 2024-03-31
Requested billtype changes

TomRoussel:
This PR contains the changes requested after PR #1290 was merged. It contains only pretty minor changes.
Specifically this PR contains the following:
 - Some tests contained leftover references to the "Transfer" type, these were removed. Also one test specifically checking the "Transfer" type has been completely removed.
   - The actual terminology might change depending on the discussion in #1299 
 - The `web.settle` route contained a "FIXME" comment. Looking into it, I think it was previously erroneously copied from `edit_bill` in web.py:811. It doesn't need to check if the bill belongs to the requested project as this function creates that bill itself.
 - In one of the tests, "fromage" was removed. This erasure has been undone, and the vibes of the codebase have been improved.
 - A test was requested for the `web.settle` route, however this route is already being tested in `budget_test.py:1309`, `test_settle_button`. In this test, the settle route is called and checked. So no new test was added.

@zorun, I think that was everything you asked back in #1290? Let me know if there is anything I missed!

### DIFF
--- a/ihatemoney/tests/budget_test.py
+++ b/ihatemoney/tests/budget_test.py
@@ -716,19 +716,7 @@ class TestBudget(IhatemoneyTestCase):
                 "amount": "17",
             },
         )
-
-        # transfer bill should not affect balances at all
-        self.client.post(
-            "/raclette/add",
-            data={
-                "date": "2011-08-10",
-                "what": "Transfer",
-                "payer": members_ids[1],
-                "payed_for": members_ids[0],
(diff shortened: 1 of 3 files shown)


zorun:
Thanks, it all looks good! Indeed, I overlooked the test checking the new endpoint.

FYI, we should also convert this new endpoint to POST to avoid XSS issues, but let's move that to another thread (I have started doing the conversion)

PR #1305, opened 2024-04-16
speed up unit tests

azmeuk:
With this PR the jinja templates are compiled once per test session instead of once per test, by using the jinja cache system and a pytest fixture.

On my machine, tests run in 19.50s instead of 27,84s.

https://jinja.palletsprojects.com/en/3.1.x/api/#jinja2.FileSystemBytecodeCache


almet:
That's very nice :-) Thanks!

zorun:
Nice improvement, thanks!

PR #1308, opened 2024-04-26
Added a filter option to the Bills screen

### LINKED ISSUE #1155: Add filter option to the Bills screen
Bills screen there is no option to filter the bill like date and who paid its easy-to-get details individually.

charlieezh:
Closes #1155
Hi, I just implemented a new feature as requested in the issue #1155 that we can filter the bill screen based on start date and end date. I made changes in three files - added a filter form in the list_bills.html template, and added corresponding functions in the model.py and web.py. I hope it would be helpful!

### DIFF
--- a/ihatemoney/models.py
+++ b/ihatemoney/models.py
@@ -264,6 +264,12 @@ class Project(db.Model):
             .order_by(Bill.id.desc())
         )
 
+    @staticmethod
+    def filter(query, start_date, end_date):
+        return (
+            query.filter(Bill.date >= start_date, Bill.date <= end_date)
+        )
+
     def get_bill_weights(self):
         """
         Return all bills for this project, along with the sum of weight for each bill.
@@ -284,6 +290,9 @@ class Project(db.Model):
     def get_bill_weights_ordered(self):
         """Ordered version of get_bill_weights"""
         return self.order_bills(self.get_bill_weights())
+
+    def get_filtered_bill_weights_ordered(self, start_date, end_date):
+        return self.filter(self.get_bill_weights_ordered(), start_date, end_date)
 
     def get_member_bills(self, member_id):
         """Return the list of bills related to a specific member"""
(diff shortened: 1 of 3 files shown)


zorun:
Hi, thanks for the proposal. Is this something you need yourself?

Could you share a screenshot of how this new feature looks like in the interface? As you may have seen, UI/UX considerations are important for the project (see https://github.com/spiral-project/ihatemoney?tab=readme-ov-file#current-direction-as-of-2024)

charlieezh:
> Hi, thanks for the proposal. Is this something you need yourself?
> 
> Could you share a screenshot of how this new feature looks like in the interface? As you may have seen, UI/UX considerations are important for the project (see https://github.com/spiral-project/ihatemoney?tab=readme-ov-file#current-direction-as-of-2024)

Hi, this feature is submitted as an issue in #1155 by another user and I implemented it. I also think this is a very useful and fundamental feature since we can sort out the transaction based on a period of time and see how much is being spent within that time. Below is a screenshot of the filter feature in the interface (note that we can always improve the look later using css)
image
After setting the start date to 2024-04-15 and the end date to 2024-04-25, only transactions between this time show up on the screen and the dates input remained in the form, indicating the effectiveness of the functionality. (Note that each members amount did not change on the left, only the transactions shown up have changed.)
image

almet:
Thanks for this work. I wonder if we should only display it when we have a certain number of bills (maybe when we have more than a page). What do you think?

charlieezh:
> Thanks for this work. I wonder if we should only display it when we have a certain number of bills (maybe when we have more than a page). What do you think?

Personally I would like this option to be consistent to be on the screen no matter how many bills there are, since sometimes people do want to filter and sort out their bills even though there are not many of them.

almet:
That seems okay to me. I was suggesting this because this can "clutter" the UI, by showing too many options, especially for new projects. Maybe we could hide this behind a "filter bills" link, or something similar?

charlieezh:
> Maybe we could hide this behind a "filter bills

Yea that sounds good to me!

sergiors:
+1

Jojo144:
I agree with the button.
It would also be nice to filter by the name/payer/ower (but especially the name).
@charlieezh Do you want any help to work on this?

PR #1313, opened 2024-05-15
feat(tags): Add tags on bills (WIP)

almet:
Tags can now be added in the description of a bill, using a hashtag symbol (`#tagname`), and it will be stored in a specific `tag` table.

(This is still a work in progress, it's not finished. Comments welcome on the approach)


almet:
A few remarks:

- I'm not even sure we should add a table for this. Maybe having an encoded list would be enough.
- I would need to expose this with the REST API
- Migrations would need to be run before adding these changes. It seems they haven't been created on latest `main`branch

almet:
This now works properly. Tags are displayed on the main bill list, and monthly statistics have been updated.

Here are a few screen captures:

image

image

image

Glandos:
My little experience on UI told me that we could have a little bit of CSS to show that it's a tag, e.g. using a rounded border and colored background. Like from [Material design](https://m3.material.io/foundations/interaction/states/overview#ed7c501f-5cf9-432a-8453-dea5dfaa7f25):
image

zorun:
Nice work! I remember trying to convince you a few months ago that tags were more useful than simple categories, happy to see the result ;)

I haven't looked at the code, but I have a few questions UX-wise:

- what happens when there are lots of tags? Does it generate 100 columns in the statistics page?
- how does the list of bills look like on mobile with the new column? I remember it's quite tight already
- is this feature "discoverable", i.e. how can I know that I can enter tags in the "What" field?
- is there some auto-completion of tags to reuse existing ones, instead of creating slightly different tags? like `#animal` vs `#Animal` vs `#animals`

The statistics page is complex, we would like to show stats per month and per tag, but this is really a lot of data. Maybe something dynamic with javascript? I imagine showing just one amount per month, and then being able to click on tags to enable/disable them individually, and that would update the amount display per month.

Another nice visualisation for stats would be a breakdown: 20% of expenses for food, 10% for transport... Of course the total can be more than 100% if some bills have several tags, so I'm not sure how to present that meaningfully. We would need some user stories to guide the UX.

PR #1323, opened 2024-08-10
UI/navigation changes to in "Edit this participant"

singeltary:
Hi,
I updated the form for editing participants. I added a cancel/back button, and changed the button text to reflect that it's an edit vs add. I believe this makes the app a little easier to navigate, and a little less confusing. 
Thank you


almet:
Hi, thanks for taking the time to open this. If that's possible for you, could you please attach some screenshots of the changes, so I can better understand what they look like?

I'm wondering if this form is used in another way? Would need to run the changes locally to see :-)

almet:
Thanks!

PR #1331, opened 2024-10-28
DRAFT: refactor currency converter - discussion

gr4viton:
It is just a draft to discuss the approach

I will not polish this if this is not a "approved" direction - with enough simplicity in mind. 

## tldr

refactor currency converter
- to allow multiple exchange rate getters
- defaulting to user defined csv file
- or if not defined hard coded exchange rates

based on discussion in - https://github.com/spiral-project/ihatemoney/issues/1232

### DIFF
--- a/ihatemoney/currency_convertor.py
+++ b/ihatemoney/currency_convertor.py
@@ -1,9 +1,16 @@
+import csv
 import traceback
 import warnings
+from abc import ABC, abstractmethod
+from decimal import Decimal
+from typing import Dict, Optional
 
 from cachetools import TTLCache, cached
 import requests
 
+
+NO_CURRENCY = "XXX"
+ExchangeRates = Dict[str, float]
 
 class Singleton(type):
     _instances = {}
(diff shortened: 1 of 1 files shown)


ihatemoney/currency_convertor.py
```diff
@@ -14,205 +21,93 @@ def __call__(cls, *args, **kwargs):
...
-            "XPD",
-            "XPF",
-            "XPT",
-            "YER",
-            "ZAR",
-            "ZMW",
-            "ZWL",
-        ]
+    def get_currencies(self, with_no_currency: bool=True) -> list:
+        currencies = list(self.get_rates.keys())
```
gr4viton, line 104: Or was the list of currencies hard-coded to have faster loading or something else?

ihatemoney/currency_convertor.py
```diff
@@ -14,205 +21,93 @@ def __call__(cls, *args, **kwargs):
...
+                f"Exchange rate getter failed - {traceback.format_exc(limit=0).strip()}"
+            )
+
+    @abstractmethod
+    def _get_rates(self) -> Optional[ExchangeRates]:
+        """Actual implementation of the exchange rate getter."""
+        raise NotImplementedError
+
+
+class ApiExchangeRate(ExchangeRateGetter):
```
almet, line 45: I believe we should ditch this way of retrieving the rates using an external API all together.

ihatemoney/currency_convertor.py
```diff
@@ -14,205 +21,93 @@ def __call__(cls, *args, **kwargs):
...
     api_url = "https://api.exchangerate.host/latest?base=USD"
 
+    def _get_rates(self) -> Optional[ExchangeRates]:
+        return requests.get(self.api_url).json()["rates"]  # TODO not working currently probably
+
+
+class UserExchangeRate(ExchangeRateGetter):
+    user_csv_file = "path/to/file.csv"
```
almet, line 53: I don't believe using a CSV file would be useful here, instead, we could propose the user to define the rates themselves for the project, in the settings.

ihatemoney/currency_convertor.py
```diff
@@ -14,205 +21,93 @@ def __call__(cls, *args, **kwargs):
...
+            rate = float(row[1])
+            # TODO add validation and exception handling for typos
+            rates[from_currency] = rate
+        return rates
+
+
+class HardCodedExchangeRate(ExchangeRateGetter):
+
+    def _get_rates(self) -> Optional[dict]:
+        return {"USD": 1.0}  # TODO fill in more
```
almet, line 83: We could have the default values defined here, for sure.

ihatemoney/currency_convertor.py
almet: I believe we could have a set of predefined rates here, but it's really not a requirement. If we want to have some of them, we could just put the most used ones here (like USD, EUR, and a few others). We don't need the full list here though :-)

ihatemoney/currency_convertor.py
gr4viton: Understandable. I saw the comment mentioning other apps are also favoring the user-defined rates - and I get the simplicty is the key and not worth maintaining part of the app which would be used just by small partition of its users (providing API keys) .

Will remove.

ihatemoney/currency_convertor.py
gr4viton: Understandable. Makes sense. Will try to incorporate the settings :)

almet:
@gr4viton hey let me know if you need any help on this :-)

almet:
FWIW, I plan on removing support for multi currencies altogether with https://github.com/spiral-project/ihatemoney/pull/1362

PR #1333, opened 2024-11-15
feat: preprend "I Hate Money" to the page title

BoboTiG:
In browser tabs, I like to see the name of the website. I used the same format as it's done in the feed: https://github.com/spiral-project/ihatemoney/blob/6582e2c0c3b5bd84516715d287e54e26147d09e1/ihatemoney/templates/project_feed.xml#L7
So here is a tiny change, and it's also OK  to close the PR if such a modification is not welcome :)

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -173,6 +173,14 @@ URL you want.
 -   **Default value:** `""` (empty string)
 -   **Production value:** The URL of your chosing.
 
+## SITE_NAME
+
+It is possible to change the name of the site to something at your liking.
+
+-   **Default value:** `"I Hate Money"` (empty string)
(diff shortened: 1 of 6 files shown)


almet:
Hi, thanks for taking the time to propose this.

I feel that "I Hate Money" in addition to "Account manager" might be too much. Maybe we should decide on one of them?

Also, I'm thinking we might want to introduce a setting for this, so it's possible to change the name of the app, if people want to "rebrand" it somehow.

What do you think?

almet:
I've updated your PR with the changes I proposed, thanks!

BoboTiG:
Thanks to you!

PR #1334, opened 2024-11-29
I added Azerbaijani Translation

ferid333:
I added Azerbaijani Translation


ferid333:
What is the reason for the failure ?

ferid333:
@almet There is no conflict

almet:
Hi, @ferid333, it's now merged :-) Congrats on the translation, and thanks again for waiting a bit on this one!

PR #1348, opened 2024-12-20
Remove flake8

almet:
Remove it from the deps as it's not used anymore.


Glandos (approved):
Well, since you migrated to `uv`, I see no issue with that.

Issue #1063, opened 2022-09-11
Dropping Python 3.6 support


Glandos:
So, it seems that every library developer is dropping support for Python 3.6:
- Flask
- Werkzeug
- Flask-limiter

The latter is the most important to me, because it allows to close some security issues regarding spamming.

On the other hand, some Linux distribution still use Python 3.6 as the main supported (?) version. Mainly Centos 7 and OpenSUSE Leap. Centos is [going away in 2024](https://endoflife.date/centos), and OpenSUSE Leap 15.4 will [fade in late 2023](https://endoflife.date/opensuse)

I don't think it's a good idea to support Python3.6 until then, but… we have a difficult choice:
- Keep Python 3.6 and give the most of security update for IHM users. Flask-limiter is not an option though, since recent version ([that are needed for what we do](https://github.com/alisaifee/flask-limiter/issues/350)) don't support it
- Move to (at least) Python 3.7, and drop support for some users that don't want to (or can't) upgrade their Python.

I really don't know the Python landscape in our users, so it's a difficult choice. For example, what is the state of the main instance in ihatemoney.org?

almet:
Thanks for bringing this up. I believe if this is blocking development, dropping support for older python version is a good idea. There are ways to install newer versions of python on these systems anyway.

About the hosted instance, it's able to run recent versions of python, so we don't have to worry.

Of course, if anybody here depends on older systems which aren't updated wrt python, let us know!

almet:
Closing, we don't support python 3.6 anymore.

PR #1349, opened 2024-12-20
Drop python 3.8

almet:
Security support has ended in October.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8, 3.9, "3.10", "3.11", "3.12"]
+        python-version: [3.9, "3.10", "3.11", "3.12"]
         dependencies: [normal]
         database: [sqlite]
         # Test other databases with only a few versions of Python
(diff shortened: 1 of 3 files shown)


Glandos (commented):
Is there any other reason than ending security support?

As I mentionned in https://github.com/spiral-project/ihatemoney/issues/1063 some people don't have the choice to use a newer Python.
E.g. SuSE Linux Enterprise 15 uses 3.6 in it's base install. There are official package for 3.12 though, but sometimes, you don't have the choice.

Otherwise, Debian oldstable is using 3.9. Oldoldstable is using 3.7, but it's very old, some maybe dropping 3.8 is reasonable though.

almet:
> Is there any other reason than ending security support?

Some packages started dropping support for python 3.8 (see the failing dependabot PRs), making us unable to upgrade their version, which might cause some security problems at some point (maybe).

Unless there is a specific person not able to install ihm on one specific system, I believe it's reasonable to drop python 3.8.

Glandos (approved):

almet:
Leaving some time for @zorun to weight in. Will merge next week unless somebody objects :-)

almet:
Now merged in the `main` branch (not sure why github marks it as closed rather than merged)

PR #1358, opened 2024-12-25
Update installation.md with uv dependency

Jojo144:
Probably not perfect but better than nothing :-)

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -78,6 +78,15 @@ Thanks again!
 (setup-dev-environment)=
 ## Set up a dev environment
 
+### Requirements
+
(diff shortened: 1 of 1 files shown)


almet:
This documentation is about the dependencies of IHM as a project. `uv` is only required if you want to install and work on ihatemoney, so I believe this should belong to https://github.com/spiral-project/ihatemoney/blob/master/docs/contributing.md#set-up-a-dev-environment instead, adding a requirement section.

Jojo144:
Updated!
I had hesitated :-)

almet (approved):
This looks perfect, thanks a lot @jojo144

Issue #1324, opened 2024-08-20
Internal Server Error with History


Jojo144:
On my production server, history works well for new small projects. But for old projects I get "Internal Server Error" when I go on the history page.
I didn't succeed in reproducing the bug on a small example in a dev environment.

After some investigations, I realized that `detailed_version.payer` and `detailed_version.owers` returns None here:

https://github.com/spiral-project/ihatemoney/blob/710aee97115e501eb548b9e8b4a9713037af0793/ihatemoney/history.py#L96-L101

So if I replace `describe_version(detailed_version.payer))` by `str(Person.query.get(detailed_version.payer_id, project))` it works for the payer. I don't know how to do the same trick for owers.

* Any idea why the `payer` method returns None here?

* Any idea why the object `detailed_version` does not have `owers` attribute?

Is it linked to the fact that this relationship is manually defined?

https://github.com/spiral-project/ihatemoney/blob/710aee97115e501eb548b9e8b4a9713037af0793/ihatemoney/models.py#L644-L649

Jojo144:
Is it link to this?
https://stackoverflow.com/questions/42457115/sqlalchemy-get-returns-none

I don't understand the answer really...

zorun:
Thanks for the investigation. I don't understand the history code in details, sorry... Potentially, we manipulate old version of objects, and I don't know how it interacts with schema changes and migrations.

If I had to guess, I see two possibilities:

- either the history code does not handle migrations, so we need to take into account that old objects do not conform to the current database model (what a mess!)
- either your old projects did not correctly get migrations applied

I checked who wrote that code, and it's you :smile: #1223 But the complexity comes from previous code, of course.

Jojo144:
I spent a couple of hours to investigate further but I didn't find anything convincing.
I propose #1359 as workaround.

almet:
I'm closing this as we have a workaround for it now. Please reopen if you have another error.

PR #1359, opened 2024-12-25
Workaround to avoid history bug #1324

Jojo144:
This is a poor workaround to bug #1324.
It does not solve the issue but at least it avoid the "Internal server error", which is nice.

Now the bills for which payer or ower are not well versioned seem to:
image

If anyone has the same issue as me, I found a trick to make members well versioned: deactivate and reactivate them. The bills recorded afterward are well versioned.

### DIFF
--- a/ihatemoney/history.py
+++ b/ihatemoney/history.py
@@ -38,7 +38,7 @@ def history_sort_key(history_item_dict):
 
 def describe_version(version_obj):
     """Use the base model str() function to describe a version object"""
-    if version_obj is None:
+    if not version_obj:
         return ""
     else:
         return parent_class(type(version_obj)).__str__(version_obj)
--- a/ihatemoney/history.py
+++ b/ihatemoney/history.py
@@ -38,7 +38,10 @@ def history_sort_key(history_item_dict):
 
 def describe_version(version_obj):
     """Use the base model str() function to describe a version object"""
-    return parent_class(type(version_obj)).__str__(version_obj)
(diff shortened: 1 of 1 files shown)


ihatemoney/history.py
```diff
@@ -38,7 +38,10 @@ def history_sort_key(history_item_dict):
 
 def describe_version(version_obj):
     """Use the base model str() function to describe a version object"""
-    return parent_class(type(version_obj)).__str__(version_obj)
+    if version_obj is None:
```
almet, line 41:
```suggestion
    if not version_obj:
```

almet:
Thanks :+1: Added a suggestion, other than that it seems fine by me

almet (approved):

ihatemoney/history.py
Jojo144: It's a few characters shorter but less clear I feel. Anyway, I don't mind a lot...

almet: Let me explain why I added this comment (sorry if this felt a bit short). This is not testing the exact same thing, so it's not only a matter of preference.

Given the fact that we're not using types in the project (we could), testing for Falsey values allows to consider both `None` and `False`. This is a general thing I try to do in these situations. Here it doesn't really matter to be honest, but that's my natural tendency.

Feel free to update your code or not, and we can merge :-)

Jojo144: Interesting, thanks.

PR #1361, opened 2024-12-25
Move the bill type field in the More options section

Jojo144:
The "Bill type" field shouldn't be better placed in the "More options" section? What do you think?

I feel that it makes the UI clearer:
image
image

(works also in the /add and /edit pages)


almet:
That seem clearer to me as well, thanks.

almet (approved):

PR #1362, opened 2024-12-26
Remove multi currencies support

almet:
For more information about the rationale, this was [discussed](https://github.com/spiral-project/ihatemoney/issues/1232#issuecomment-2081517453) in our bugtracker.

Actually, this also removes the notion of currency, and I'm not sure it should go away, as it might be beneficial to keep it in order to display the currency of the project (to add a € or $ sign, for instance)


Glandos:
I don't think removing support for it is a good idea. I much prefer having a manual rate forms, where the project admin must enter all rates from currently used currencies.

almet:
> I don't think removing support for it is a good idea. I much prefer having a manual rate forms, where the project admin must enter all rates from currently used currencies.

Thanks for your input, but I fear it's coming a bit late. As you can see in https://github.com/spiral-project/ihatemoney/issues/1232#issuecomment-2081517453 we discussed it this summer and decided to remove support unless somebody would like to contribute such a change, as nobody on the maintainer team wants to do it.

It seems counter-intuitive to remove a feature, but I really believe it's the direction we should take, unless somebody wants to weight in and do it.

almet:
> unless somebody wants to weight in and do it.

And actually, I believe it might be too late for this as well, since we've been waiting for 6 months already.

Glandos:
Yes, I know. I'm just sad, but I don't have enough time to work on this :(

almet:
> Yes, I know. I'm just sad, but I don't have enough time to work on this :(

I agree with you :-) The goal is to reduce the maintenance burder as much as possible, and if somebody wants to add the feature back, it'll be welcome!

PR #1367, opened 2025-01-28
Eliminate duplicate “Settle” button

mavit:


almet:
Thanks, that looks like a failed conflict resolution on a recent merge.

PR #1374, opened 2025-04-03
430 add ability to search filter expenses

### LINKED ISSUE #430: Add the ability to search / filter expenses

MilouH-TUE:
Closes #430 

I created this pull request in collaboration with @c5laungani . This merge request adds filter options for the bills list as requested in issue 430. We used the previous attempts for this issue as our starting point. Just before opening this pull request, we also observed that there was another issue (1155), that is a duplicate of the filtering issue. In a recent pull request for that issue, the feedback was to add tests. So in this pull request we tried to cover all feedback from previous attempts.

Screenshots:
filter_closed
filter_shown
filter_applied

Demo video:
https://github.com/user-attachments/assets/40465943-3819-4158-b642-b3fc5899c065

Please let us know if there's anything we should look into further!

### DIFF
--- a/ihatemoney/templates/list_bills.html
+++ b/ihatemoney/templates/list_bills.html
@@ -126,32 +126,34 @@
             <div class="form-row w-100">
               <div class="col-md-2 mb-2">
                 <label for="search" class="sr-only">{{ _("Search") }}</label>
-                <input type="text" class="form-control form-control-sm w-100" id="search" name="search" placeholder="{{ _('Search in For what?') }}" value="{{ search_query }}">
+                <input type="text" class="form-control form-control-sm w-100" id="search" name="search" placeholder="{{ _('Search in For what?') }}" value="{{ search }}">
               </div>
               <div class="col-md-2 mb-2">
                 <label for="payer" class="sr-only">{{ _("Payer") }}</label>
                 <select class="form-control form-control-sm w-100" id="payer" name="payer">
-                  <option value="">{{ _('Any payer') }}</option>
-                  {% for member in g.project.active_members %}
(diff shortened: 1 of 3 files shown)


Jojo144:
Thanks for your contribution! And sorry for the very late answer.
Here are my first comments:
- [ ] it should nice to have a "for whom" field, it is useful -> maybe too complicated in the UI for the usefulness
- [ ] we should be accent insensitive (useful in french) -> seems not so simple to do
- [ ] does "Clear filters" should really dismiss the panel? I would prefer not. -> maybe it's good like this, it does a real reset so in case of bug it is usefull
- [x] I think that the "Filter" button should be normal size instead of btn-sm

Jojo144:
Also:
- [x] we should accept comma in "min" and "max" fields and replace it by a dot
- [x] the links are wrong when there are several pages (it clean the filters)

I tested on a project with lot of bills, it will be very handy!

Jojo144:
For me this is ready for reviews

JocelynDelalande:
I tried to carefully test it in various cases… works great !

I fixed minor UX issues, see my commits.

I think that the top-bar over the list is kind-a clumsy, but this is clearly another topic. Same for accent-insensitive search, as I agree with @Jojo144 : seems non-trivial.

I'm excited we finally get filtering implemented ! Thanks @MilouH-TUE and @c5laungani to put it on the workbench again 🫶 ! And @Jojo144 for further testing and improving.

I hope this is the right one !

@Jojo144 I think you could squash all PR commits into one single commit (GH offers to do that with merge button), as our log is kinda-noisy :-)

Jojo144 (approved):

Jojo144:
\o/

Issue #1380, opened 2025-06-01
Use participant weight when displaying their share for a bill (or hide it)


greedybro:
Hi !

This is a minor issue, but the information displaying "X$ each" is misleading when using different weights for several participant.

I have a shared project with 2 participants. One is 1,15 and the other 0,85 but the bill is still displaying half for each of the participant in the list (see screen).

<img width="226" alt="Image" src="https://github.com/user-attachments/assets/b2014fa1-19b1-4336-8580-9db6d8799810" />

I believe the code is here, just checking if there is more than one participant and dividing `bill / participant`.
https://github.com/spiral-project/ihatemoney/blob/56bee933468a6f8fe912d8b8e91849fad41fec6d/ihatemoney/templates/list_bills.html#L3

It would be a nice feature to have a way to find out each shares for the bill, although I realize that the UI may not be able to handle displaying the correct share for each participant, especially if they are numerous.

Maybe it should be considered to hide this info when the weight are different than `1` for at least one participant, as it doesn't reflect the correct shares anyway.

EDIT:

It is also displaying a weird amount when one participant is paying for the other only. It is typically what happens when you create a reimbursement bill. But then the info is showing the amount payed, divided by the weight of the person that the amount was payed for.

In this screen, it was divided by 0.85, but it has no sense this the person weighted 0,85 did not pay anything for this billed

<img width="279" alt="Image" src="https://github.com/user-attachments/assets/ca05fcbb-ca96-488e-89d8-229da611a317" />

zorun:
Small nit: no need to display "custom split", just display nothing between parentheses (as we already do when the total weight is exactly 1).

Jojo144:
Hello,
I see several options to solve the issue:
1. we write something like "XX by portion" instead of "XX each" (in french "XX par part" I don't know what is the correct translation) so that the meaning is correct
2. we remove this indication everytime
3. we remove the indication in case of custom split (the "nit" of @zorun)

I'm in favor of 2. which refines the UI...

Jojo144:
With @JocelynDelalande we choose solution 3. Thanks for your contributions.

Jojo144:
Here is a new idea which is a trade-off to reduce the number of requests:
- display the "xx for each" information only for project which does not use weights
- never display it if when there are weights in the project

Jojo144:
> Here is a new idea which is a trade-off to reduce the number of requests:
> 
>     * display the "xx for each" information only for project which does not use weights
> 
[...]

PR #1385, opened 2025-09-03
fix: remove amounts owed if custom split

GauvainThery:
Hello there 👋 

I'm starting to use my spare time to contribute on some open projects, this would be the first one 🙌 

This should address the issue as described in #1380 . I simply removed the amounts owed when the number of participant does not match the weight of the bill.

Translations coming from Weblate seem to be blocked though, so not sure what to do here.

Cheers,

### DIFF
--- a/ihatemoney/messages.pot
+++ b/ihatemoney/messages.pot
@@ -775,6 +775,9 @@ msgstr ""
 msgid "%(amount)s each"
 msgstr ""
 
+msgid "Custom split"
+msgstr ""
+
 msgid "you sure?"
 msgstr ""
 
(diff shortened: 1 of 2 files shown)


Jojo144:
Thanks for your contribution and sorry for the late answer.

We are not ready to merge this because we have to choose the good UI, see the discussion in the issue #1380 ...

Jojo144:
Hello.
Thanks for your contribution.
As said in #1380 we decided to display nothing in case of custom split.
Your proposition does not handle the case where the weights are equal but not to one (e.g. 0.5 / 0.5).
I propose another way here: #1460 let's continue the discussion there.

PR #1387, opened 2025-10-11
Use vermin 1.7.0

netromdk:
After having released version [1.7.0](https://github.com/netromdk/vermin/releases/tag/v1.7.0) of Vermin ([PyPi](https://pypi.org/project/vermin/1.7.0/)) it would make sense to also use it for the ihatemoney CI pipelines.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,7 +58,7 @@ database = [
 dev = [
     "ruff==0.8.4",
     "isort==5.11.5",
-    "vermin==1.6.0",
+    "vermin==1.7.0",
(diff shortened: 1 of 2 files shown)


JocelynDelalande:
Hi @netromdk that was legit but we are too slow at reviewing, since then, codebase already [upgraded to 1.8.0](https://github.com/spiral-project/ihatemoney/blob/main/pyproject.toml#L60). Thanks anyway, all the best !

netromdk:
You are very welcome. And that is great to hear (⌐■_■)

PR #1398, opened 2026-01-18
Prepare for 7.0.0 release

JocelynDelalande:


zorun:
For a major release, I would separate the changelog in several sections (usually "Breaking changes / Add / Fix", but here maybe "Breaking changes / Add / Updates"). It helps to understand that we actually have a single new feature, the rest is mostly tooling and dependency changes.

I'm surprised at the number of translation strings updates. Some of them date back from the currency hotfix in #1240, and there have been a release in-between. I guess I forgot to update strings when I did a release.

zorun:
I am usually grumpy about dropping Python versions, but I think it's fine in this case:

- 6.x supports Python 3.7-3.11
- 7.x will support Python 3.11-3.13

So, there is still an acceptable upgrade path for people running Debian: first upgrade to Debian bookworm with IHM 6.x, then upgrade IHM to 7.x, then upgrade Debian to trixie.

CONTRIBUTORS
almet: Maybe update this heading as well, as the statement "friends are helping since then" isn't really true anymore =D (I don't know the vast majority of contributors here)

CHANGELOG.md
```diff
@@ -2,14 +2,17 @@
 
 This document describes changes between each past release.
 
-## 6.2.0 (unreleased)
+## 7.0.0 (unreleased)
 
+- Update all Python dependencies (#1392)
+- Update db versions used in CI to match Debian Trixie (#1392)
```
almet: Stating what's the version of the database using in the CI might help make things clearer.


```suggestion
- Use postgreSQL 17 and mariadb 11.8 in the CI, to match Debian Trixie (#1392)
```

almet (approved):
Hey, so cool to see this happening! Thanks for the release preparation, it's long overdue and will be useful :-) It's exciting to see new efforts in here.

Here are quick thoughts:

- Splitting the changelog in "Added", "Removed", "Development changes" might help folks better understand that some work has been done to help developers (for instance the CI changes), but will not affect them directly (so I agree with you @zorun)

- Related to this last point, it might be useful to introduce a "Breaking changes" section, so that folks understand directly why a new major version is released rather than a minor one (we drop python support).

- Removing the `CONTRIBUTORS` file might make sense. It's an artifact of another age, and it's easy to access this information in the repository.

Otherwise it all looks good to me!

JocelynDelalande:
I thank I addressed all your comments (thanks !), let's publish :-)

PR #1429, opened 2026-03-20
Fix currency conversion by switching to working exchange rate API

atfg:
## Summary

- Replace defunct `api.exchangerate.host` with `open.er-api.com`, which returns the same JSON structure and is free without an API key
- Remove the blanket workaround in `validate_original_currency` (added in #1232) that blocked all multi-currency bills
- The validator now only raises an error if the rates fetch actually fails or the currency isn't in the returned rates

## Test plan

- [ ] Create a project with EUR as default currency
- [ ] Add a bill with CHF — should convert instead of raising "Failed to convert from CHF currency to EUR"
- [ ] Verify same-currency and no-currency bills still work
- [ ] Simulate API failure and confirm the error message still appears

### DIFF
--- a/ihatemoney/currency_convertor.py
+++ b/ihatemoney/currency_convertor.py
@@ -17,7 +17,7 @@ class Singleton(type):
 class CurrencyConverter(object, metaclass=Singleton):
     # Get exchange rates
     no_currency = "XXX"
-    api_url = "https://api.exchangerate.host/latest?base=USD"
+    api_url = "https://open.er-api.com/v6/latest/USD"
 
     def __init__(self):
         pass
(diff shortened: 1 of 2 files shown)


Jojo144:
Hi,
Please do not spend time on this as we plan to remove multi currencies support... See #1362. Thanks.

atfg:
> Hi, Please do not spend time on this as we plan to remove multi currencies support... See #1362. Thanks.

Multi-currency support is essential for me. I’ll continue this work in my fork if there's no interest here.

JocelynDelalande:
@atfg Feel free, that's what open-source allows you to do :-). 

And sorry that the project current strategy diverge from your needs. Cheers !

PR #1432, opened 2026-03-27
Don't blindly default to the first participant as payer.

firetech:
This has caused some confusion among my friends.

### DIFF
--- a/ihatemoney/forms.py
+++ b/ihatemoney/forms.py
@@ -83,7 +83,8 @@ def get_billform_for(project, set_default=True, **kwargs):
 
     active_members = [(m.id, m.name) for m in project.active_members]
 
-    form.payed_for.choices = form.payer.choices = active_members
+    form.payed_for.choices = active_members
+    form.payer.choices = [("", "")] + active_members
     form.payed_for.default = [m.id for m in project.active_members]
 
     if set_default and request.method == "GET":
@@ -352,7 +353,11 @@ class ResetPasswordForm(FlaskForm):
 class BillForm(FlaskForm):
     date = DateField(_("When?"), validators=[DataRequired()], default=datetime.now)
     what = StringField(_("What?"), validators=[DataRequired()])
-    payer = SelectField(_("Who paid?"), validators=[DataRequired()], coerce=int)
+    payer = SelectField(
+        _("Who paid?"),
+        validators=[DataRequired()],
+        coerce=lambda v: int(v) if v != "" else "",
+    )
     amount = CalculatorStringField(_("How much?"), validators=[DataRequired()])
     currency_helper = CurrencyConverter()
     original_currency = SelectField(_("Currency"), validators=[DataRequired()])


ihatemoney/forms.py
```diff
@@ -83,7 +83,8 @@ def get_billform_for(project, set_default=True, **kwargs):
 
     active_members = [(m.id, m.name) for m in project.active_members]
 
-    form.payed_for.choices = form.payer.choices = active_members
+    form.payed_for.choices = active_members
+    form.payer.choices = [(-1, '')] + active_members
```
JocelynDelalande: If you use empty string as value instead of -1, HTML validation on required field will detect it as no choice and thus prevent form submission.

(I did not test implementing it with Python though).

JocelynDelalande (changes requested):
Hello ! Thanks for your contribution. 

It looks legit to me. I believe it could help IHM use :).

See my remarks for some simplification.

Cheers !

ihatemoney/forms.py
firetech: That was actually my initial thought, but due to the `payer` field having `coerce=int` (at least I think that's the reason), that caused an Exception about invalid int value while loading the page. That said, my experience with these specific libraries is low, so I may have made a mistake somewhere.

firetech: Removing `coerce=int` from the payer field seems to work (including submitting), but I have no idea if this causes issues elsewhere.

firetech: Replaced -1 value and the custom validator with empty string value and a custom coersion.

JocelynDelalande: Great ! Thank you :) 

Just have to test it one of these days and wait for some other maintainer feedback if any.

Also could you please [run linter](https://ihatemoney.readthedocs.io/en/latest/contributing.html#formatting-code) please ?

firetech: Hmm, `make lint` didn't complain. `make format`, however, made changes both to my edits and seven files I didn't touch... I included the changes it made to my edits at least.

JocelynDelalande:
@firetech Thanks for your adjustments :).

I tested your fix localy. Sadly, even if the blank option now appears, it is not selected by default, I tested your branch on demo project : 

image

Does it work for you ?

firetech:
It does work for me, but if you've added a payment in the same session (same session cookie or local storage, not entirely sure), the form (somewhat logically) defaults to the last used payer, same behaviour as before.

My change only affects the state in which you have not added a payment before, in which case the current default is the first payer.

Jojo144:
Thanks for your contribution, it's simple but nice.

I tested locally and read the code. It seems good to me.


> It does work for me, but if you've added a payment in the same session (same session cookie or local storage, not entirely sure), the form (somewhat logically) defaults to the last used payer, same behaviour as before.
> 
> My change only affects the state in which you have not added a payment before, in which case the current default is the first payer.

You are rigth...

JocelynDelalande:
Merged, thank you again :-)

PR #1448, opened 2026-05-09
Remove participant highlighting

Jojo144:
Address #1437


Jojo144:
reviewed with @JocelynDelalande

PR #1452, opened 2026-05-17
Align members_stats to include deactivated members

### LINKED ISSUE #1451: Bad row alignment on Statistic page.
On the Statistic page, deactivated members are visible on the side bar but not on the table, so the data is just wrong.

Image

Image

I think this is a major UX issue.

ArnoldRynux:
Fixes #1451 

Align `members_stats` to include deactivated members with significant balances.

Add test `test_members_stats_includes_deactivated_with_balance` in `budget_test.py` to verify that `members_stats()` only returns activated members or members with `|balance| > 0.01` and that deactivated members with non-zero balances are included.

bills
statics

### DIFF
--- a/ihatemoney/models.py
+++ b/ihatemoney/models.py
@@ -192,7 +192,8 @@ class Project(db.Model):
                 "received": -1.0 * received[member.id],
                 "balance": balance[member.id],
             }
-            for member in self.active_members
+            for member in self.members
+            if member.activated or abs(balance[member.id]) > 0.01
         ]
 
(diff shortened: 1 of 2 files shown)


Jojo144:
Thanks for your contribution, I'll review it as far as I have time! I'm not quick but I should not forget...

almet (approved):
Thanks, that looks good to me (and thanks for adding tests).

Jojo144:
Perfect, thanks.

PR #1458, opened 2026-06-28
Close connection in _pre_alembic_db to fix MySQL 8 migration deadlock

### LINKED ISSUE #1457: First migration deadlocks on MySQL 8: _pre_alembic_db() leaks an unclosed connection
### Summary

On a fresh / empty **MySQL 8** database, the first startup hangs forever in `setup_database()`. `_pre_alembic_db()` opens `db.engine.connect()` and never closes it; under SQLAlchemy 2.0 that connection keeps an open transaction holding metadata locks on `project` / `alembic_version`. Alembic then opens its own connection (the separate engine in `migrations/env.py`) and `CREATE TABLE alembic_version` blocks on that metadata lock. MySQL's `lock_wait_timeout` defaults to ~1 year, so it hangs silently instead of erroring.

### Environment

- ihatemoney 7.1.1 (SQLAlchemy 2.0.x)
- MySQL 8.0
- Fresh, empty database (first migration)

### Symptoms

App logs stop at:

```
INFO  [alembic.runtime.migration] Context impl MySQLImpl.
INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
```

`SHOW FULL PROCESSLIST` shows two connections from the same process:

```
ihm  Sleep  ...                              -- holds the metadata lock
ihm  Query  Waiting for table metadata lock  -- CREATE TABLE alembic_version (...)
```

### Root cause

`ihatemoney/run.py`, `_pre_alembic_db()`:

```python
con = db.engine.connect()
tables_exist = db.engine.dialect.has_table(con, "project")
alembic_setup = db.engine.dialect.has_table(con, "alembic_version")
return tables_exist and not alembic_setup
```

`con` is never closed. Under SQLAlchemy 1.x (ihatemoney ≤ 6.x) the implicit-autocommit connection didn't hold a transaction, so this was harmless; SQLAlchemy 2.0 removed autocommit, so the connection now holds the transaction (and its locks) until it is closed. SQLite (the default) and already-provisioned databases don't surface it, which is likely why it has gone unnoticed.

### Fix

Close the connection (e.g. via a context manager) before `upgrade()` runs. PR incoming.

jfilter:
Fixes #1457.

`_pre_alembic_db()` leaves its `db.engine.connect()` connection open; under SQLAlchemy 2.0 that holds a transaction + metadata locks and deadlocks the first migration on a fresh MySQL 8 database. This closes it via a context manager before `upgrade()` runs — no change on SQLite or already-migrated databases.

Tested against a fresh MySQL 8 (docker-compose + a REST-API smoke test); SQLite unaffected. Happy to add a regression test if wanted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@ This document describes changes between each past release.
 ## 7.1.2 (unreleased)
 ---------------------
 
-- Nothing changed yet.
+### Fixed
+- Fix first-database migration deadlock on MySQL 8 by closing the connection opened in `_pre_alembic_db` (#1457)
 
 
 ## 7.1.1 (2026-05-09)
--- a/ihatemoney/run.py
+++ b/ihatemoney/run.py
@@ -36,10 +36,10 @@ def setup_database(app):
 
     def _pre_alembic_db():
         """Checks if we are migrating from a pre-alembic ihatemoney"""
-        con = db.engine.connect()
-        tables_exist = db.engine.dialect.has_table(con, "project")
-        alembic_setup = db.engine.dialect.has_table(con, "alembic_version")
-        return tables_exist and not alembic_setup
+        with db.engine.connect() as con:
+            tables_exist = db.engine.dialect.has_table(con, "project")
+            alembic_setup = db.engine.dialect.has_table(con, "alembic_version")
+            return tables_exist and not alembic_setup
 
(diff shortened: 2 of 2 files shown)


Jojo144:
Thanks for your contribution, I'll review it at my next coding session.

zorun (approved):
Looks good to me!

Jojo144:
Thanks again.

PR #1460, opened 2026-07-19
Display "... each" only when there are several participants and when the split is even

Jojo144:
Supersede #1385


Jojo144:
Hum. There is a problem: it multiplies the number of requests ... (several additional request for each bill).

PR #1462, opened 2026-07-20
CI: update versions for Ubuntu & Python

zorun:


zorun:
Hoping to catch #1457 (although it seems that Mysql 8 has really diverged from MariaDB)

zorun:
Just found out that @JocelynDelalande already updated stuff a few months ago, so we're good.

Next step for MariaDB is 12.3 ( https://endoflife.date/mariadb ) but let's wait its introduction in Debian, probably for the next stable.

zorun:
Actually, let's at least update Ubuntu

